PR #10, opened 2019-06-14
Conutils

vpetrovykh:
There are 2 commits here:
- setting up connection parameters parser in con_utils and unit tests for that
- integrating the con_utils connection parser into the client code and slight tweaks to how tests are run

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -34,6 +34,11 @@ import {
   LocalTime,
   Duration,
 } from "./datatypes/datetime";
+import {
+  parseConnectArguments,
+  ConnectConfig,
+  NormalizedConnectConfig,
+} from "./con_utils";
 
 type QueryArgPrimitive =
   | number
@@ -48,13 +53,6 @@ type QueryArgPrimitive =
   | UUID;
 type QueryArg = QueryArgPrimitive | QueryArgPrimitive[] | null;
 type QueryArgs = {[_: string]: QueryArg} | QueryArg[] | null;
-
-export interface ConnectConfig {
-  port?: number;
-  host?: string;
-  user?: string;
-  database?: string;
-}
 
 enum AuthenticationStatuses {
   AUTH_OK = 0,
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -71,6 +69,15 @@ enum TransactionStatus {
   TRANS_UNKNOWN = 4, // cannot determine status
 }
 
+class TimeoutError extends Error {
```
1st1: I don't like this -- ideally this should be (a) exported, (b) defined in the `errors/` module.  The latter is a bit complicated, since it's autogenerated.  I suggest to simply use the standard `Error` class for timeout errors (as do 99% of all NodeJS packages).

src/con_utils.ts
```diff
@@ -0,0 +1,402 @@
...
+  host,
+  port,
+  user,
+  password,
+  database,
+  admin,
+  server_settings,
+}: ConnectConfig): NormalizedConnectConfig {
+  if (dsn) {
+    const dsnHostMatch = /\/\/(.+?@)?(.*)\//.exec(dsn);
```
1st1: I think you can use https://nodejs.org/docs/latest/api/url.html here.

src/client.ts
vpetrovykh: OK, I've removed `TimeoutError` in favor of `Error`.

src/con_utils.ts
vpetrovykh: I've tried it already and it had many issues with DSN containing comma-separated list of hosts like this one `edgedb://user@host1,host2/db`. This type of "host" messed up the pathname and port results, hence why I ended up replacing it, parsing the rest of the DSN as if the host was some simple placeholder and then hand-parsing the comma-separated hosts separately.

1st1: A comment in the code explaining this would be great.

vpetrovykh: Yeah, good point. I've updated the comments.

PR #12, opened 2019-07-18
Docs

vpetrovykh:
Update some inconsistencies in the code that came up during documenting the API.

Add Sphinx documentation.

### DIFF
--- a/docs/api/connection.rst
+++ b/docs/api/connection.rst
@@ -1,0 +1,280 @@
+.. _edgedb-js-api-reference:
+
+===
+API
+===
+
+.. _edgedb-js-api-connection:
+
+Connection
+==========
+
+.. js:function:: connect(options)
+                 connect(options, callback)
+
+    Establish a connection to an EdgeDB server.
+
+    :param options: Connection parameters object.
+
+    :param string options.dsn:
+        Connection arguments specified using as a single string in the
+        connection URI format:
+        ``edgedb://user:password@host:port/database?option=value``.
+        The following options are recognized: host, port,
+        user, database, password.
+
+    :param string|string[] options.host:
+        Database host address as one of the following:
+
+        - an IP address or a domain name;
+        - an absolute path to the directory containing the database
+          server Unix-domain socket (not supported on Windows);
+        - an array of any of the above, in which case the addresses
+          will be tried in order, and the first successful connection
+          will be returned.
+
+        If not specified, the following will be tried, in order:
+
+        - host address(es) parsed from the *dsn* argument,
+        - the value of the ``EDGEDB_HOST`` environment variable,
+        - on Unix, common directories used for EdgeDB Unix-domain
+          sockets: ``"/run/edgedb"`` and ``"/var/run/edgedb"``,
+        - ``"localhost"``.
+
+    :param number|number[] options.port:
+        Port number to connect to at the server host
+        (or Unix-domain socket file extension).  If multiple host
+        addresses were specified, this parameter may specify an
+        array of port numbers of the same length as the host array,
+        or it may specify a single port number to be used for all host
+        addresses.
+
+        If not specified, the value parsed from the *dsn* argument is used,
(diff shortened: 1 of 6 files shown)


docs/api/types.rst
```diff
@@ -0,0 +1,423 @@
...
+Objects
+=======
+
+``Object`` represents an object instance returned from a query. The value of an
+object property or a link can be accessed through a corresponding object key:
+
+.. code-block:: js
+
+    // Use the Node.js assert library to test results.
+    const assert = require("assert");
```
1st1: Can we use `import` statement?

src/datatypes/datetime.ts
```diff
@@ -99,7 +99,7 @@ export class LocalDateTime {
   }
 
   toString(): string {
-    return this._date.toUTCString();
+    return this._date.toUTCString().slice(0, -4); // cut off " GMT"
```
1st1: Does it always end with "GMT"? Maybe we want to add an assertion here?

docs/api/connection.rst
```diff
@@ -0,0 +1,278 @@
...
+        const edgedb = require("edgedb");
+        const connect = edgedb.default;
+        var conn;
+
+        async function main() {
+          conn = await connect({
+            dsn: "edgedb://edgedb@localhost/"
+          });
+
+          let data = await conn.fetchOne("SELECT 1 + 1");
```
1st1: Just enclose this block in `try { ... } finally { await conn.close() }`. This will make code more idiomatic async/await

docs/api/connection.rst
```diff
@@ -0,0 +1,278 @@
...
+                    console.log('migration complete');
+                }
+            })
+
+    .. js:method:: fetchAll(query: string, args)
+
+        Run a query and return the results as a
+        :js:class:`Set` instance.
+
+        This method takes :js:class:`optional arguments <Connection>`.
```
1st1: `optional arguments` -- you mean "optional query arguments"?

docs/api/connection.rst
```diff
@@ -0,0 +1,278 @@
...
+          let data = await conn.fetchOne("SELECT 1 + 1");
+
+          // The result is a number 2.
+          assert(typeof data === "number");
+          assert(data === 2);
+        }
+
+        main().finally(() => {
+          // Close the connection.
+          conn.close();
```
1st1: I think that this is also technically incorrect, as you're supposed to return Promise to chain them -- in your case you're starting an async operation (close) but not waiting for it to complete.  Again, better to replace with try..finally in the main function.  Then this snippet becomes `main()` -- short and sweet.

src/index.ts
```diff
@@ -16,7 +16,12 @@
  * limitations under the License.
  */
 
-export {LocalDateTime} from "./datatypes/datetime";
+export {
+  LocalDateTime,
+  LocalDate,
+  LocalTime,
+  Duration,
+} from "./datatypes/datetime";
```
1st1: Good catch.

docs/api/types.rst
vpetrovykh: `import` is not supported in plain nodejs without `--experimental-modules` and using `*.mjs` files. The bottom line is that the import syntax requires the use of some extra things like babel or special node mode.

I have, however, changes the package to expose `connect` as both a default and a named export so that `const {connect} = require("edgedb");` should work (currently it doesn't).

src/datatypes/datetime.ts
vpetrovykh: OK, I've added a check.

docs/api/connection.rst
vpetrovykh: Yes. Updated.

docs/api/connection.rst
vpetrovykh: Updated

docs/api/connection.rst
vpetrovykh: I've replaced this pattern in all code samples.

PR #16, opened 2019-12-06
Fix outdated tests, namely:

tailhook:
1. Unsupported date format
2. cal:: module
3. Remove duration syntax that going to be disallowed
4. The `serializable` thing probably, something of the new postgres


tailhook:
@1st1, I don't have write permissions for this repository.

PR #17, opened 2019-12-10
Fix duration to obey new semantics

tailhook:
The controversial decisions:
1. `new Duration` takes **milliseconds** as it's what JS uses for `Date - Date`
2. Internally we store microseconds (but this can be changed in future)
3. Currently, it crashes with overflow when duration doesn't fit 53 bits (size of a float). But there are few ways to fix it (bigint in node > 12.0, split the number for earlier versions). But fortunately, it's forward-compatible.

### DIFF
--- a/docs/api/types.rst
+++ b/docs/api/types.rst
@@ -394,24 +394,29 @@ Local Date and Time
 Duration
 ========
 
-.. js:class:: Duration(\
-        months: number = 0, \
-        days: number = 0, \
-        microseconds: number = 0)
+.. js:class:: Duration(milliseconds: number = 0)
 
     A JavaScript  representation of an EdgeDB ``duration`` value.
 
-    .. js:method:: getMonths(): number
+    .. js:method:: fromMicroseconds(us: bigint): Duration
+        :staticmethod:
 
-        Get the number of months in the duration.
+        Create duration from bigint number of microseconds.
 
-    .. js:method:: getDays(): number
-
-        Get the number of days in the duration.
+        Note: new Duration() accepts fractional seconds too but can loose
+        precision because it's floating point.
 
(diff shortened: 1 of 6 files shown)


1st1:
Note: bigint is in node >= 10.0 as a BigInt class; bigint literals are only in 12

tailhook:
Fixed duration to use bigint. Things to look at:
1. I don't know how to set private field in custom constructor (i.e. static function). Added an undocumented parameter instead
2. We might consider using `fromMicros` instead of `fromMicroseconds` (borrowed from rust: shorter and still clean enough)
3. We should consider `getWhatever` -> `asWhatever`. The `get` prefix makes sense for dates, because `getHours` actually gets hours (without anything else). Bug `getMilliseconds` returns the whole duration in millisecond equivalent (also borrowed from rust)

1st1:
> We might consider using fromMicros instead of fromMicroseconds (borrowed from rust: shorter and still clean enough)

I was going to suggest this -- let's do it.

> getWhatever -> asWhatever.

Also +1 where it makes sense.

tailhook:
> > We might consider using fromMicros instead of fromMicroseconds (borrowed from rust: shorter and still clean enough)
> 
> I was going to suggest this -- let's do it.

Well, I'm going to back off this a bit. The issue is that in `datetime_trunc` and `datetime_get` and we have `microseconds` and `milliseconds`. Remembering where they are abbreviated is annoying.

Or does it make sense to fix all of them?

tailhook:
I've converted methods to using `to` instead of `get`. The units are left longer, to be consistent to `duration_trunc` and other datetime things.

@edgedb/engineering, do anybody want to make another review?

PR #21, opened 2020-01-10
Start temporary EdgeDB cluster automatically in tests

1st1:


test/globalTeardown.ts
```diff
@@ -0,0 +1,14 @@
+export default async () => {
+  // @ts-ignore
+  const proc = global.edgedbProc;
+  // @ts-ignore
+  const conn = global.edgedbConn;
+  // tslint:disable-next-line
+  console.log("Shutting down EdgeDB test cluster...");
+
+  await conn.close();
+  proc.kill();
```
elprans: Why an explicit `kill` if you rely on `auto-shutdown`?  Maybe `join`?

1st1: >  if you rely on auto-shutdown

Only when you press ^C -- in which case nodejs env just craps out and dies immediately. In this case EdgeDB will shutdown itself.

>  Maybe join?

No ".join()" in nodejs APIs

I'll try removing `.kill` though...

1st1: There's 'close' event I can probably use https://nodejs.org/api/child_process.html#child_process_child_process

Issue #30, opened 2020-01-23
"unexpected message" error when queries run concurrently


jaclarke:
```js
const edgedb = require("edgedb")

async function main() {
  const conn = await edgedb.connect({
    // connection config
  })

  try {

    console.log( await conn.fetchAll(`SELECT sys::get_version()`) )
    // Set [ NamedTuple [ major := 1, minor := 0, stage := 'alpha', stage_no := 2, local := [ 'dev', '535', 'gbd7bd5fd' ] ] ]
    console.log( await conn.fetchAll(`SELECT sys::get_version_as_str()`) )
    // Set [ '1.0-alpha.2+dev.535.gbd7bd5fd' ]

    console.log(
      await Promise.all([
        conn.fetchOne(`SELECT sys::get_version()`),
        conn.fetchOne(`SELECT sys::get_version_as_str()`)
      ])
    )
    // Error: unexpected message type 49 ("1")
    //  at AwaitConnection._fallthrough (#################\node_modules\edgedb\dist\src\client.js:244:23)
    //  at AwaitConnection._execute (#################\node_modules\edgedb\dist\src\client.js:536:26)
    //  at process._tickCallback (internal/process/next_tick.js:68:7)

  } finally {
    await conn.close()
  }
}

main()
```

1st1:
Running queries concurrently on **the same one connection** isn't going to be supported. There's a bug though: the driver should detect such situations and issue a clear error. 

To run multiple queries in parallel use multiple different connections.

1st1:
Opening multiple connections will be simplified once we land https://github.com/edgedb/edgedb-js/issues/5

1st1:
I've published 0.7.0-alpha.6 where the error message is more descriptive. Thank you for reporting this!

PR #31, opened 2020-01-24
Complain when concurrent operations are run on the same connection

1st1:
Addresses #30.


src/client.ts
```diff
@@ -967,24 +969,65 @@ export class AwaitConnection {
     if (error != null) {
       throw error;
     }
+  }
 
-    return;
+  private _enterOp(): void {
+    if (this.opInProgress) {
+      throw new Error(
+        "Another operation is in progress. Use multiple separate " +
```
elprans: Remove one “multiple,”

PR #34, opened 2020-01-29
Fix bug when arg properties are in different order to query args

jaclarke:
Fixes bug in `NamedTupleCodec.encodeArgs` where the properties of an args object are just assigned to the query args in the order that they are returned by `Object.keys`, not by name, leading to correctness bugs (if the types happen to be compatible) or codec encoding errors.
```js
const edgedb = require('edgedb') // 0.7.0-alpha.5

async function main() {
  const conn = await edgedb.connect({
    // connection config
  })

  try {

    console.log(
      await conn.fetchAll(`SELECT <int64>$a - <int64>$b`, {
        a: 10,
        b: 5,
      })
    )
    // Set [ 5 ]

    console.log(
      await conn.fetchAll(`SELECT <int64>$a - <int64>$b`, {
        b: 5,
        a: 10,
      })
    )
    // Set [ -5 ]

    console.log(
      await conn.fetchAll(`SELECT (
        <int64>$number,
        <str>$string
      )`, {
        string: 'test',
        number: 42,
      })
    )
    // Error: a number was expected, got "test"
    // at Int64Codec.encode (##############\node_modules\edgedb\dist\src\codecs\numbers.js:24:19)
    // at NamedTupleCodec.encodeArgs (##############\node_modules\edgedb\dist\src\codecs\namedtuple.js:61:23)
    // at AwaitConnection._encodeArgs (##############\node_modules\edgedb\dist\src\client.js:485:28)
    // at AwaitConnection._execute (##############\node_modules\edgedb\dist\src\client.js:494:31)

  } catch (error) {
    console.error(error)
  }

  await conn.close()
}

main()
```
This commit does also change how extra properties on the args object are handled, as they are now ignored, which allows using an existing object with other properties as the args parameter, for example:
```js
const user = {
  id: 12345,
  name: 'test',
  // ...etc
}

await conn.fetchOne(`
UPDATE User
FILTER .id = <int64>$id
SET {
  status := 'online'
};
`, user)

// instead of ...
await conn.fetchOne(`
UPDATE User
FILTER .id = <int64>$id
SET {
  status := 'online'
};
`, {
  id: user.id
})
```
Though I can add checks for extra properties if you prefer enforcing the more explicit usage.

### DIFF
--- a/src/codecs/namedtuple.ts
+++ b/src/codecs/namedtuple.ts
@@ -49,9 +49,20 @@ export class NamedTupleCodec extends Codec implements ICodec, IArgsCodec {
       );
     }
 
+    const keys = Object.keys(args);
     const names = this.names;
+    const namesSet = this.namesSet;
     const codecs = this.subCodecs;
     const codecsLen = codecs.length;
+
+    if (keys.length > codecsLen) {
+      const extraKeys = keys.filter((key) => !namesSet.has(key));
+      throw new Error(
+        `unexpected named argument${ extraKeys.length === 1 ? '' : 's' }: "${
+          extraKeys.join('", "')
+        }"`
+      );
+    }
 
     if (!codecsLen) {
       return EmptyTupleCodec.BUFFER;
--- a/src/codecs/namedtuple.ts
+++ b/src/codecs/namedtuple.ts
@@ -58,9 +58,9 @@ export class NamedTupleCodec extends Codec implements ICodec, IArgsCodec {
     if (keys.length > codecsLen) {
       const extraKeys = keys.filter((key) => !namesSet.has(key));
       throw new Error(
-        `unexpected named argument${ extraKeys.length === 1 ? '' : 's' }: "${
-          extraKeys.join('", "')
-        }"`
+        `unexpected named argument${
+          extraKeys.length === 1 ? "" : "s"
+        }: "${extraKeys.join('", "')}"`
       );
     }
 
--- a/src/codecs/namedtuple.ts
+++ b/src/codecs/namedtuple.ts
@@ -49,18 +49,9 @@ export class NamedTupleCodec extends Codec implements ICodec, IArgsCodec {
       );
     }
 
-    const keys = Object.keys(args);
-    const names = this.namesSet;
+    const names = this.names;
     const codecs = this.subCodecs;
     const codecsLen = codecs.length;
-
-    if (keys.length !== codecsLen) {
-      throw new Error(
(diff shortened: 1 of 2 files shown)


src/codecs/namedtuple.ts
```diff
@@ -49,31 +49,24 @@ export class NamedTupleCodec extends Codec implements ICodec, IArgsCodec {
       );
     }
 
-    const keys = Object.keys(args);
-    const names = this.namesSet;
+    const names = this.names;
     const codecs = this.subCodecs;
     const codecsLen = codecs.length;
 
-    if (keys.length !== codecsLen) {
```
1st1: Good catch! Although I'd keep this check -- we don't want users to supply arguments that aren't used in the query.

1st1:
> Though I can add checks for extra properties if you prefer enforcing the more explicit usage.

Yeah, I'd keep the current "strict mode" for now. We can relax it later if people request that.

src/codecs/namedtuple.ts
jaclarke: Would it be useful to say what the extra unused properties are in the error, or just that there are too many?

1st1: I'd say it would be useful, yes. Please add a test for that.

1st1:
I'll publish a new bugfix release as soon as this lands.

test/client.test.ts
```diff
@@ -275,11 +275,19 @@ test("fetch: named args", async () => {
...
       b: "abc",
+      a: "123",
     });
     expect(res).toBe("123abc");
 
+    res = await con.fetchOne(`select <str>$a ++ <str>$b`, {
+      b: "abc",
+      a: "123",
+      c: "def",
+    }).catch((e) => {
```
1st1: Add a `then` that should raise an exception -- that way we ensure that if we accidentally stop raising an error we detect that.

1st1:
Looks good, thanks so much! Need to fix the CI though. Could you please run `yarn lint` and fix what it complains about? Also `yarn format`.

jaclarke:
Sorry that took a while, I was getting a 'No matching files error' from yarn format and yarn lint was completing without any errors, and I haven't been able to work out what the problem is (windows environment maybe?), so I've fixed the formatting manually from the CI logs

1st1:
Thanks so much! I'll push a release now.

> Sorry that took a while, I was getting a 'No matching files error' from yarn format and yarn lint was completing without any errors, and I haven't been able to work out what the problem is (windows environment maybe?)

Sorry to hear that. Maybe it's indeed Windows. :(

PR #43, opened 2020-04-07
Fix bug where connection is in broken state after no codec error

jaclarke:
Fixes bug where queries made after a query that fails with a no codec error, also fail as there is still the "T" message from the previous query left over in the read message buffer.
```js
const conn = await connect(...)

async function runQuery(query) {
  try {
    return await conn.fetchOne(query)
  } catch (err) {
    console.log(err)
  }
}

console.log( await runQuery(`SELECT 123.456n`) )
// Error: no JS codec for std::decimal                 <-- Expected error

console.log( await runQuery(`SELECT 123.456`) )
// Error: unexpected message type 84 ("T")
//   at AwaitConnection._fallthrough (.../edgedb-js/dist/src/client.js:247:23)
//   at AwaitConnection._parse (.../edgedb-js/dist/src/client.js:426:26)
//   at AwaitConnection._fetch (.../edgedb-js/dist/src/client.js:637:58)
//   at AwaitConnection.fetchOne (.../edgedb-js/dist/src/client.js:736:31)
```

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -246,27 +246,32 @@ export class AwaitConnection {
   }
 
   private _parseDescribeTypeMessage(): [number, ICodec, ICodec] {
-    this._rejectHeaders();
+    try {
+      this._rejectHeaders();
 
-    const cardinality: char = this.buffer.readChar();
+      const cardinality: char = this.buffer.readChar();
 
-    const inTypeId = this.buffer.readUUID();
-    const inTypeData = this.buffer.readLenPrefixedBuffer();
-    let inCodec = this.codecsRegistry.getCodec(inTypeId);
-    if (inCodec == null) {
-      inCodec = this.codecsRegistry.buildCodec(inTypeData);
+      const inTypeId = this.buffer.readUUID();
+      const inTypeData = this.buffer.readLenPrefixedBuffer();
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -246,27 +246,32 @@ export class AwaitConnection {
...
 
-    const outTypeId = this.buffer.readUUID();
-    const outTypeData = this.buffer.readLenPrefixedBuffer();
-    let outCodec = this.codecsRegistry.getCodec(outTypeId);
-    if (outCodec == null) {
-      outCodec = this.codecsRegistry.buildCodec(outTypeData);
-    }
+      const outTypeId = this.buffer.readUUID();
+      const outTypeData = this.buffer.readLenPrefixedBuffer();
+      let outCodec = this.codecsRegistry.getCodec(outTypeId);
```
1st1, line 263: I'd do it the other way:

* parsing of the protocol message and `this.buffer.finishMessage();` should be outside the `try` block; 
* `this.codecsRegistry.getCodec(outTypeId);` should be after the parsing -- and then you probably don't need `try..catch` at all

1st1:
Thanks so much, @jaclarke!

PR #46, opened 2020-04-24
Implement a (private) API to toggle text decoders for some scalars

1st1:
The new API allows to decode bigint, decimal, and int64 into strings,
allowing edgedb-js to work in browsers without bigint and to decode
decimal values.


1st1:
I'll push another PR with one more change in it; closing this one.

PR #52, opened 2020-06-15
Adds a connection pool

RobertoPrevato:
The code mostly replicates the logic of the Python driver implementation.

### DIFF
--- a/docs/api/connection.rst
+++ b/docs/api/connection.rst
@@ -44,13 +44,13 @@ Connection
     :param number|number[] options.port:
         Port number to connect to at the server host
         (or Unix-domain socket file extension).  If multiple host
-        addresses were specified, this parameter may specify an
+        addresses are specified, this parameter may specify an
         array of port numbers of the same length as the host array,
         or it may specify a single port number to be used for all host
         addresses.
 
         If not specified, the value parsed from the *dsn* argument is used,
-        or the value of the ``EDGEB_PORT`` environment variable, or ``5656``
+        or the value of the ``EDGEDB_PORT`` environment variable, or ``5656``
         if neither is specified.
 
     :param boolean options.admin:
@@ -82,8 +82,8 @@ Connection
         Connection timeout in seconds.
 
     :param callback:
-        A callback function that will be invoked when the conenction is ready.
-        The callback funciton should be of the form ``function(error,
+        A callback function that will be invoked when the connection is ready.
+        The callback function should be of the form ``function(error,
         connection)``. The *connection* is an instance of
         :js:class:`Connection`.
 
@@ -175,7 +175,7 @@ Connection
         <edgedb-js-api-async-optargs>`.
 
         The *query* must return exactly one element.  If the query returns
-        more than one element or an empty set, an ``Error`` is thown.
+        more than one element or an empty set, an ``Error`` is thrown.
 
     .. js:method:: fetchAllJSON(query: string, args)
 
@@ -205,7 +205,7 @@ Connection
         <edgedb-js-api-async-optargs>`.
 
         The *query* must return exactly one element.  If the query returns
-        more than one element or an empty set, an ``Error`` is thown.
+        more than one element or an empty set, an ``Error`` is thrown.
 
         .. note::
 
@@ -286,7 +286,7 @@ Connection
         <edgedb-js-api-sync-optargs>`.
 
         The *query* must return exactly one element.  If the query returns
-        more than one element or an empty set, an ``Error`` is thown.
+        more than one element or an empty set, an ``Error`` is thrown.
 
     .. js:method:: fetchAllJSON(query: string, args, callback)
 
@@ -316,7 +316,7 @@ Connection
         <edgedb-js-api-sync-optargs>`.
 
         The *query* must return exactly one element.  If the query returns
-        more than one element or an empty set, an ``Error`` is thown.
+        more than one element or an empty set, an ``Error`` is thrown.
 
         .. note::
 
@@ -339,10 +339,379 @@ Connection
             The *callback* to be invoked after closing the connection.
 
 
-    .. js:method:: wrap(conn: AwaitConnection): Connection
-        :staticmethod:
-
-        Convert an :js:class:`AwaitConnection` into :js:class:`Connection`.
-
 .. _BigInt:
     https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt
+
+
+.. _edgedb-js-api-pool:
+
+Pool
+==========
+
+.. js:function:: createPool(options)
+                 createPool(options, callback)
+
+    Create a connection pool to an EdgeDB server.
+
+    :param options: Connection pool parameters object.
+
+    :param ConnectConfig options.connectOptions:
+        Connection parameters object, used when establishing new connections.
+        Refer to the documentation at :ref:`edgedb-js-api-connection`.
+
+    :param number options.minSize:
+        The minimum number of connections initialized by the connection pool.
+        If not specified, this value is by default 0: the first connection is
+        created when required.
+
+    :param number options.maxSize:
+        The maximum number of connections created by the connection pool.
+        If not specified, this value is by default 100.
+
+    :param func options.onAcquire:
+        Optional callback, called when a connection is acquired.
+        *(proxy: PoolConnectionProxy) => Promise<void>*
+
+    :param func options.onRelease:
+        Optional callback, called when a connection is released.
+        *(proxy: PoolConnectionProxy) => Promise<void>*
+
+    :param func options.onConnect:
+        Optional callback, called when a new connection is created.
+        *(connection: AwaitConnection) => Promise<void>*
+
+    :param func options.connectionFactory:
+        Optional function, used to obtain a new connection. By default, the
+        function is :js:func:`connect` *(options?: ConnectConfig) =>
+        Promise<AwaitConnection>*
+
+    :param callback:
+        A callback function that will be invoked when the connection pool is
+        ready. The callback function should be of the form ``function(error,
+        pool)``. The *pool* is an instance of :js:class:`CallbackPool`.
+
+    :returns:
+        There are two ways of creating an EdgeDB connection pool: a
+        Promise-based approach and a callback-based. When a *callback*
+        argument is provided, the function does not return anything and instead
+        uses the *callback* with an instance of :js:class:`CallbackPool`.
+        Otherwise, a ``Promise`` of an :js:class:`Pool` is returned.
+
+.. js:class:: Pool
+
+    A connection pool is used to manage a set of connections to a database.
+    Since opening connections is an expensive operation, connection pools are
+    used to maintain and reuse connections, enhancing the performance of
+    database interactions.
+
+    Pools can be created using the method ``createPool``:
+
+    .. code-block:: js
+
+        const edgedb = require("edgedb");
+
+        async function main() {
+            const pool = await edgedb.createPool({
+                connectOptions: {
+                    user: "edgedb",
+                    host: "127.0.0.1",
+                },
+            });
+
+            try {
+                let data = await pool.fetchOne("SELECT [1, 2, 3]");
+
+                console.log(data);
+            } finally {
+                // in this example, the pool is closed after a single
+                // operation; in real scenarios a pool is initialized
+                // at application startup, and closed at application shutdown
+                await pool.close();
+            }
+        }
+
+        main();
+
+    The pool accepts the following parameters:
+
+    .. js:method:: execute(query: string)
+
+        Acquire a connection, then execute an EdgeQL command (or commands).
+        The commands must take no arguments.
+
+        :param query: Query text.
+
+        .. code-block:: js
+
+            await pool.execute(`
+                CREATE TYPE MyType {
+                    CREATE PROPERTY a -> int64
+                };
+                FOR x IN {100, 200, 300}
+                UNION INSERT MyType { a := x };
+            `)
+
+    .. js:method:: fetchAll(query: string, args)
+
+        Acquire a connection, then run a query and return the results as a
+        :js:class:`Set` instance.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-sync-optargs>`.
+
+        .. code-block:: js
+
+            const items = await pool.fetchAll(
+                `SELECT Movie {
+                    title,
+                    year,
+                    director: {
+                        first_name,
+                        last_name
+                    },
+                    actors: {
+                        first_name,
+                        last_name
+                    }
+                }
+                FILTER .id = <uuid>$id;`,
+                {
+                    id: movieId,
+                }
+            );
+
+    .. js:method:: fetchOne(query: string, args)
+
+        Acquire a connection, then run a query that returns a single item
+        and return its result.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-async-optargs>`.
+
+        The *query* must return exactly one element.  If the query returns
+        more than one element or an empty set, an ``Error`` is thrown.
+
+        .. code-block:: js
+
+            await pool.fetchOne("SELECT 1");
+
+    .. js:method:: fetchAllJSON(query: string, args)
+
+        Acquire a connection, then run a query and return the results as JSON.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-async-optargs>`.
+
+    .. js:method:: fetchOneJSON(query: string, args)
+
+        Acquire a connection, then run a singleton-returning query and return
+        its element in JSON.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-async-optargs>`.
+
+        The *query* must return exactly one element.  If the query returns
+        more than one element or an empty set, an ``Error`` is thrown.
+
+    .. js:method:: acquire()
+
+        Acquire a connection proxy, which provides access to an open database
+        connection. The proxy must be released to return the connection to the
+        pool.
+
+        Example:
+
+        .. code-block:: js
+
+            const connection = await pool.acquire();
+            let value: number;
+
+            try {
+                value = await connection.fetchOne("select 1");
+            } finally {
+                await pool.release(connection);
+            }
+
+    .. js:method:: release(connectionProxy: PoolConnectionProxy)
+
+        Release a previously acquired connection proxy, to return it to the
+        pool.
+
+    .. js:method:: run<T>(action: func)
+
+        Acquire a connection and use it to run the given action that accepts
+        a connection, and return *T*, which is any type returned by the user's
+        defined function argument. The connection is automatically returned
+        to the pool.
+
+        Example:
+
+        .. code-block:: js
+
+            const result = await pool.run(async (connection) => {
+                return await connection.fetchOne("SELECT 1");
+            });
+            expect(result).toBe(1);
+
+    .. js:method:: getStats()
+
+        Return information about the current state of the pool. Information
+        include the number of currently open connections, and the number of
+        pending consumers, awaiting for a connection to become available.
+
+        Example:
+
+        .. code-block:: js
+
+            const stats = pool.getStats();
+            const queueLength = stats.queueLength;
+            const openConnections = stats.openConnections;
+
+    .. js:method:: expireConnections()
+
+        Expire all currently open connections.
+        Cause all currently open connections to be replaced when they are
+        acquired by the next *.acquire()* call.
+
+    .. js:method:: close()
+
+        Close the connection pool gracefully. When a connection pool is closed,
+        all its underlying connections are awaited to complete their pending
+        operations, then closed. A warning is produced if the pool takes more
+        than 60 seconds to close.
+
+    .. js:method:: terminate()
+
+        Terminate all connections in the pool, closing all connections non
+        gracefully. If the pool is already closed, return without doing
+        anything.
+
+.. _edgedb-js-api-callbackpool:
+
+CallbackPool
+============
+
+.. js:class:: CallbackPool
+
+    A :js:class:`CallbackPool` exposes a callback-based API, and can be created
+    using the function :js:func:`createPool`, specifying a callback as second
+    argument.
+
+    Every method of this class takes a *callback* of the form
+    ``function(err, data)``.
+
+    .. code-block:: js
+
+      const edgedb = require("edgedb");
+
+      edgedb.createPool({
+          connectOptions: {
+              user: "edgedb",
+              host: "127.0.0.1"
+          },
+      }, (err, pool) => {
+          if (err) {
+              throw err;
+          }
+
+          pool.fetchOne("select <int64>$i + 1", { i: 10 }, (err2, data2) => {
+              if (err2) {
+                  throw err2;
+              }
+
+              console.log(data2);
+
+              pool.close();
+          });
+      });
+
+    .. js:method:: execute(query: string, callback)
+
+        Acquire a connection and execute an EdgeQL command (or commands).
+
+        The commands must take no arguments.
+
+        Example:
+
+        .. code-block:: js
+
+            pool.execute(`
+                CREATE TYPE MyType {
+                    CREATE PROPERTY a -> int64
+                };
+                FOR x IN {100, 200, 300}
+                UNION INSERT MyType { a := x };
+            `, (err, data) => {
+                if (err) {
+                    console.log('migration failed: ', err)
+                } else {
+                    console.log('migration complete');
+                }
+            })
+
+    .. js:method:: fetchAll(query: string, args)
+
+        Acquire a connection and run a query and return the results as a
+        :js:class:`Set` instance.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-sync-optargs>`.
+
+    .. js:method:: fetchOne(query: string, args, callback)
+
+        Acquire a connection and run a singleton-returning query and return
+        its element.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-sync-optargs>`.
+
+        The *query* must return exactly one element.  If the query returns
+        more than one element or an empty set, an ``Error`` is thrown.
+
+    .. js:method:: fetchAllJSON(query: string, args, callback)
+
+        Acquire a connection and run a query and return the results as JSON.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-sync-optargs>`.
+
+    .. js:method:: fetchOneJSON(query: string, args, callback)
+
+        Acquire a connection and run a singleton-returning query and return
+        its element in JSON.
+
+        This method takes :ref:`optional query arguments
+        <edgedb-js-api-sync-optargs>`.
+
+        The *query* must return exactly one element.  If the query returns
+        more than one element or an empty set, an ``Error`` is thrown.
+
+    .. js:method:: close(callback)
+
+        Close the connection pool gracefully and invoke the *callback*.
+
+        :param callback:
+            The *callback* to be invoked after closing the connection.
+
+    .. js:method:: acquire(callback)
+
+        Acquire a connection proxy, which provides access to an open database
+        connection. The proxy must be released to return the connection to the
+        pool. Then invoke the *callback*.
+
+        :param callback:
+            The *callback* to be invoked after closing the connection.
+
+    .. js:method:: release(proxy: CallbackPoolConnectionProxy, callback)
+
+        Release a previously acquired connection proxy, to return it to the
+        pool, and invoke the *callback*.
+
+        :param callback:
+            The *callback* to be invoked after closing the connection.
+
+    .. js:method:: expireConnections()
+
+        Expire all currently open connections.
+        Cause all currently open connections to be replaced when they are
+        acquired by the next *.acquire()* call.
--- a/src/client.ts
+++ b/src/client.ts
@@ -19,6 +19,7 @@
 import * as net from "net";
 
 import char, * as chars from "./chars";
+import * as errors from "./errors";
 import {resolveErrorCode} from "./errors/resolve";
 import {ReadMessageBuffer, WriteMessageBuffer, ReadBuffer} from "./buffer";
 import {CodecsRegistry} from "./codecs/registry";
@@ -57,7 +58,7 @@ type QueryArgPrimitive =
   | Duration
   | UUID;
 type QueryArg = QueryArgPrimitive | QueryArgPrimitive[] | null;
-type QueryArgs = {[_: string]: QueryArg} | QueryArg[] | null;
+export type QueryArgs = {[_: string]: QueryArg} | QueryArg[] | null;
 
 enum AuthenticationStatuses {
   AUTH_OK = 0,
@@ -93,7 +94,7 @@ export default function connect(
   if (callback) {
     AwaitConnection.connect(options)
       .then((conn) => {
-        callback(null, Connection.wrap(conn));
+        callback(null, new Connection(conn));
       })
       .catch((error) => {
         callback(<Error>error, null);
@@ -103,11 +104,30 @@ export default function connect(
   }
 }
 
-export class AwaitConnection {
+export interface ConnectionProxy {
+  onConnectionClose: () => void;
+}
+
+export interface IConnection {
+  execute(query: string): Promise<void>;
+
+  fetchAll(query: string, args?: QueryArgs): Promise<Set>;
+
+  fetchAllJSON(query: string, args?: QueryArgs): Promise<string>;
+
+  fetchOne(query: string, args?: QueryArgs): Promise<any>;
+
+  fetchOneJSON(query: string, args?: QueryArgs): Promise<string>;
+
+  close(): Promise<void>;
+}
+
+export class AwaitConnection implements IConnection {
   private sock: net.Socket;
   private config: NormalizedConnectConfig;
   private paused: boolean;
   private connected: boolean = false;
+  private _proxy: ConnectionProxy | null;
 
   private lastStatus: string | null;
 
@@ -137,6 +157,7 @@ export class AwaitConnection {
     this.queryCodecCache = new LRU({capacity: 1000});
 
     this.lastStatus = null;
+    this._proxy = null;
 
     this.serverSecret = null;
     this.serverSettings = new Map<string, string>();
@@ -158,6 +179,7 @@ export class AwaitConnection {
     this.sock.on("error", this._onError.bind(this));
     this.sock.on("data", this._onData.bind(this));
     this.sock.on("connect", this._onConnect.bind(this));
+    this.sock.on("close", this._onClose.bind(this));
 
     this.config = config;
   }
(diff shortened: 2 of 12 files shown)


src/pool.ts
```diff
@@ -0,0 +1,677 @@
...
+  async close(): Promise<void> {
+    if (this._connection !== null) {
+      // TODO: complete, remove following comment
+      // AsyncIOConnection.aclose() will call _release_on_close() to
+      // finish holder cleanup.
+      await this._connection.close();
+    }
+  }
+
+  _release_on_close(): void {
```
1st1, line 233: Let's camelCase everywhere.

src/pool.ts
```diff
@@ -0,0 +1,677 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import connect, {AwaitConnection, QueryArgs} from "./client";
+import {InvalidValueError} from "./errors";
```
1st1, line 20: Please import as `import * as errors from "./errors"` and refrain from defining new exceptions in this file.

src/pool.ts
```diff
@@ -0,0 +1,677 @@
...
+    if (!this._minSize) {
+      return;
+    }
+
+    // Since we use a LIFO queue, the first items in the queue will be
+    // the last ones in `self._holders`.  We want to pre-connect the
+    // first few connections in the queue, therefore we want to walk
+    // `self._holders` in reverse.
+
+    const reversedHolders = reversed(this._holders);
```
1st1, line 471: I guess there's no need in the `reversed` helper -- just iterate the loop in other direction.

test/testbase.ts
```diff
@@ -21,6 +21,7 @@ import * as process from "process";
 import connect from "../src/index.node";
 import {NodeCallback, AwaitConnection, Connection} from "../src/client";
 import {ConnectConfig} from "../src/con_utils";
+import { Pool } from "../src/pool";
```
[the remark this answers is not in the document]
RobertoPrevato: Yes, I am using `esbenp.prettier-vscode`, but I didn't configure it to automatically format on save. I'm doing that now and I'll soon do a push to correct the formatting errors.

src/pool.ts
RobertoPrevato: An oversight. By the way, how do you feel about private methods starting with an underscore? I have mixed feelings about this, because TypeScript supports `private` anyway.

test/testbase.ts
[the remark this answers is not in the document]
RobertoPrevato: I was missing this setting in VS Code, that is why formatting on save wasn't working: :innocent:

```js
"editor.defaultFormatter": "esbenp.prettier-vscode"
```

src/client.ts
```diff
@@ -1065,6 +1090,13 @@ export class AwaitConnection {
       this._abort();
     } finally {
       this._leaveOp();
+      this._cleanup();
```
1st1, line 1093: Is this the only way how the Connection object can be closed?  The method has to be explicitly called by the user, what if the Connection is closed because of a network error? Can we listen for some relevant event on `this.sock`?

src/client.ts
```diff
@@ -103,11 +104,16 @@ export default function connect(
   }
 }
 
+export interface ConnectionUser {
```
1st1, line 107: This isn't related to a DB user, right? Why did you decide to use this terminology instead of "proxy" in edgedb-python?

src/client.ts
```diff
@@ -1051,6 +1058,24 @@ export class AwaitConnection {
...
 
+  isClosed(): boolean {
+    return !this.connected;
+  }
+
+  public get user(): ConnectionUser | null {
+    return this._user;
+  }
+
+  setUser(user: ConnectionUser | null): void {
```
1st1, line 1069: In addition to the naming in https://github.com/edgedb/edgedb-js/pull/52/files#r441916382, I also think that this shouldn't be a "public" method. We can't make it "private", but I'd at least prefix it with an undersore and mark it `/** @internal */`.

src/queues.ts
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export class LifoQueue<T> {
+  private _promises: Array<Promise<T>>;
```
1st1, line 20: Probably no need to use `_` -- it's already private.

src/queues.ts
```diff
@@ -0,0 +1,75 @@
...
+  private _resolvers: Array<(t: T) => void>;
+
+  constructor() {
+    this._resolvers = [];
+    this._promises = [];
+  }
+
+  private _add(): void {
```
1st1, line 28: Ditto

src/pool.ts
```diff
@@ -0,0 +1,677 @@
...
+
+  public get result(): T | PromiseLike<T> | undefined {
+    if (!this._done) {
+      throw new InterfaceError("The deferred is not resolved");
+    }
+    return this._result;
+  }
+
+  setResult(value?: T | PromiseLike<T> | undefined): void {
+    this.resolve(value);
```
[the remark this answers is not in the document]
1st1: `setTimeout` would be unnecessarily slow for this, use `process.nextTick`:

```
await new Promise((resolve) => process.nextTick(resolve));
```

> Normally, awaiting a timeout of 0 in a while loop would cause high
> CPU usage

Not really, it's not a busy loop, it's only executing once per event loop iteration.

I think the hack is OK for now. I plan to later add a proper implementation of `Deferred` that wouldn't rely on `Promises` at all. There's some code in `client.ts` that would be simplified if we had it.

src/pool.ts
```diff
@@ -0,0 +1,754 @@
...
+        this._done = true;
+        this._result = value;
+        resolve(value);
+      };
+    });
+  }
+}
+
+class PoolConnectionHolder {
+  private _pool: Pool;
```
1st1, line 86: No need for `_` for private fields. At least that's the decision I made for `client.ts` and other files, so let's keep the style consistent.

1st1: Hm, I actually looked at my code and I see that it's not as consistent as I thought. We don't have underscores for `AwaitConnection` attributes, but then we do have them for some methods.

Given that this driver will also be used by vanilla JS code, we might want to keep the underscores here and add them to `client.ts`.

src/client.ts
RobertoPrevato: I can rename to proxy. The only reason why I wrote it different is that I wanted to avoid adding complexity of defining an interface in a third common file, to reference the common object in both client.ts and pool.ts, and seemed reasonable to kept it abstract (since not only a proxy might need to cleanup resources when a connection is closed).

src/pool.ts
RobertoPrevato: I asked somewhere the same question in one of the other comments (whether we should keep underscores for private attributes). In some cases, having underscores make sense if you want to have, or add in the future, `public get` accessor.

I mean for example:
```
private _foo: boolean;

public get foo(): boolean { return this._foo; }
```

src/client.ts
RobertoPrevato: I registered an event listener to socket `onClose` event. I read your code several times in the existing `async close`, and it seems reasonable to call this method in the `onClose` event handler (I mean: not only the method that cleans the proxy), to ensure that all necessary operations are executed. Can you please confirm this?

1st1: > Can you please confirm this?

Yeah, it's totally fine to call it from `close`, just need to make sure it's called consistently in all scenarios, both user-caused and due to an internal error. Listening on `onClose` should do the trick for both cases, no?

src/pool.ts
1st1: Let's have underscores. Our vanilla JS users will appreciate that. And we'll add them to client.ts in a follow up PR.

1st1:
> Improved the api to instantiate a Pool, adding a PoolOptions interface, and also a static async method to instantiate a pool and have it initialized automatically:

I like this. Let's disable `new Pool` / `pool.initialize()` altogether.

1st1:
> The Python driver supports custom connection classes. To support the same in the Node.js driver, we would need to modify the current implementation of Connection and AwaitConnection to use protected constructors, instead of private constructors. How would you like to proceed?

We don't want to support subclassing. The only reason AwaitConnection is even exported is typing, and I think I'll fix that some other way (like exporting only the type via `ReturnOf<>`).

1st1:
> In the code above, the code execution is not resumed after the CrashTestError is thrown while iterating the async generator.

Why should it be resumed? The exception just crashes the `main()` function, I don't see what's there to execute? What does specifically fail in that example?

src/client.ts
RobertoPrevato: I asked your confirmation because I was in doubt whether I should **only** do the proxy cleaning during the `onClose` socket event handler, rather than calling the `close` method that does more cleaning.

RobertoPrevato:
> > In the code above, the code execution is not resumed after the CrashTestError is thrown while iterating the async generator.
> 
> Why should it be resumed? The exception just crashes the `main()` function, I don't see what's there to execute? What does specifically fail in that example?

I looked more into this, and my understanding was wrong. This test wanted to reproduce this one in Python:

```py
    async def test_pool_handles_transaction_exit_in_asyncgen_1(self):
        pool = await self.create_pool(min_size=1, max_size=1)

        async def iterate(con):
            async with con.transaction():
                for record in await con.fetchall("SELECT {1, 2, 3}"):
                    yield record

        class MyException(Exception):
            pass

        with self.assertRaises(MyException):
            async with pool.acquire() as con:
                agen = iterate(con)
                try:
                    async for _ in agen:  # noqa
                        raise MyException()
                finally:
                    await agen.aclose()

        await pool.aclose()
```

My understanding is that when `MyException` is raised, code execution resumes in the async generator at the point that yielded, exiting the transaction with a `ROLLBACK`, thanks to the fact that `con.transaction()` is an async context manager and its `__aexit__` method is called even in this case.

```ts
test(
  "pool handles transaction exit in asyncgen 1",
  async () => {
    const pool = await Pool.create({
      connectOptions: getConnectOptions(),
      minSize: 1,
      maxSize: 1,
    });

    let calledRollback = false;

    async function* iterate(
      connection: AwaitConnection
    ): AsyncGenerator<number, void, unknown> {
      await connection.execute("START TRANSACTION");
      try {
        const values = await connection.fetchAll("SELECT {1, 2, 3}");

        let value: number;
        for (value of values) {
          yield value;
        }
        await connection.execute("COMMIT");
      } catch (err) {
        calledRollback = true;
        await connection.execute("ROLLBACK");

        throw err;
      }
    }

    async function main(): Promise<void> {
      const proxy = await pool.acquire();

      const agen = iterate(proxy.connection);

      try {
        for await (const _ of agen) {
          throw new CrashTestError();
        }
      } finally {
        await pool.release(proxy);
      }
    }

    await expect(main()).rejects.toThrow(new CrashTestError().message);

    expect(calledRollback).toBe(true);
    await pool.close();
  },
  BiggerTimeout
);
```

The same doesn't happen in the TypeScript code I wrote: my mistake was thinking that a `catch` block would caught an exception happened during the async iteration. Instead, only the code inside a `finally` block is executed and it's the same in Python (I just verified in an example).

Anyway, since the JS driver doesn't have something like the Python driver transaction context manager, there is nothing to test here.

1st1:
> My understanding is that when MyException is raised, code execution resumes in the async generator 

The code doesn't magically resume. It's caused by the `await agen.aclose()` call, that would throw in a cancellation exception into the `yield record`. The exception would then be handled by `async with con.transaction():` in the generator.

Here's an equivalent JS code:

```javascript
async function *iter() {
    let i = 0;
    try {
        while(1) {
            await new Promise(resolve => setTimeout(resolve, 1000));
            yield i++
        }
    }
    finally {
        console.log('in finally')
        await new Promise((r) => {
            setTimeout(() => {
                console.log('finally done')
            }, 1000)
        })
    }
}

async function main() {
    const it = iter();
    try {
        for await (let i of it) {
            console.log('>>>', i);
            throw new Error;
        }
    }
    finally {
        await it.return()
    }
}

main()
```

We'll need to think how to implement `async with transaction()` in javascript so that this semantics is handled properly.

RobertoPrevato:
Unless I am wrong, the `transaction` object is not implemented in the JS driver (I'll verify later today).

Earlier today I tried the following piece of code, and the code in the `finally` block and `__aexit__` are executed even without calling the generator `.aclose()`: why is that? (just out of curiosity)

```py
import asyncio


class AsyncContext:

    async def __aenter__(self):
        return self

    async def __aexit__(self, exc_type, exc, tb):
        # And we get here, too
        print("And here, too")


class MyException(Exception):
    pass


async def iterate(times):
    async with AsyncContext():
        try:
            for x in range(times):
                await asyncio.sleep(0.01)
                yield x
        except Exception as ex:
            # We never get here!!
            print("Not here")
        finally:
            # We get here...
            print("We get here...")


async def main():
    async for item in iterate(10):
        print(item)
        if item > 3:
            raise MyException()


asyncio.run(main())
```

Produces the output:
```
0
1
2
3
4
We get here...
And here, too
Traceback (most recent call last):
...
```

1st1:
> Unless I am wrong, the transaction object is not implemented in the JS driver (I'll verify later today).

Yes, it's still a todo. 

> Earlier today I tried the following piece of code, and the code in the finally block and __aexit__ are executed even without calling the generator .aclose(): why is that? (just out of curiosity)

Because asyncio uses https://docs.python.org/3/library/sys.html#sys.set_asyncgen_hooks to call `aclose()` implicitly when an async generator is GCed. I'm not sure if nodejs works the same way, need to test that.

RobertoPrevato:
@1st1 a few days ago I wanted to propose something like this for transactions (you would find it the edit history of the description of this PR), this seems reasonable to me.
I then removed this part of text because it's a different topic then connection pooling, and deserves a dedicated PR.

```ts
async transaction<T>(
  action: (connection: AwaitConnection) => Promise<T>
): Promise<T> {
  let result: T;
  await this.execute("START TRANSACTION");
  try {
    result = await action(this);
    // All done - commit.
    await this.execute("COMMIT");
  } catch (err) {
    await this.execute("ROLLBACK");

    throw err;
  }
  return result;
}

```

1st1:
> For example, if added to the AwaitConnection class:

Yes, that's the gist of it. There are a few arguments that the function should accept (basically allowing tuning of the transaction parameters), but we can nail that in another PR.

This PR is still in draft, I assume it's not yet ready for a final review?

RobertoPrevato:
The only missing is the documentation: I wrote it but I didn't see it yet on the `edgedb-site` locally. Because I only looked at it using `make html` from Sphinx (I didn't know yet how to run `edgedb-site` when I used Sphinx directly).

docs/api/connection.rst
```diff
@@ -346,3 +346,207 @@ Connection
...
+        await pool.initialize();
+
+        try {
+            let data = await pool.fetchOne("SELECT [1, 2, 3]");
+
+            // The result is an Array.
+            assert(data instanceof Array);
+            assert(typeof data[0] === "number");
+            assert(data.length === 3);
+            assert(data[2] === 3);
```
1st1, line 389: I think that for an example in the docs using `assert` is a bit too much. I'd just

```suggestion
            console.log(data)
```

docs/api/connection.rst
```diff
@@ -346,3 +346,207 @@ Connection
...
+        } finally {
+            // in this example, the pool is closed after a single
+            // operation; in real scenarios a pool is initialized
+            // at application startup, and closed at application shutdown
+            await pool.close();
+        }
+        }
```
1st1, line 396: This bracket is weird. I guess it should be removed?

docs/api/connection.rst
```diff
@@ -346,3 +346,207 @@ Connection
...
+        Optional callback, called when a new connection is created.
+        *(connection: AwaitConnection) => Promise<void>*
+
+    :param func options.connectionFactory:
+        Optional function, used to obtain a new connection. By default, the function is `connect`
+        *(options?: ConnectConfig) => Promise<AwaitConnection>*
+
+    .. js:method:: initialize()
+
+        Initializes a connection pool, if not initialized yet. If the `pool.minSize` is a positive
```
1st1, line 435: Line lengths should be capped at 79

package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/edgedb/edgedb-js.git"
   },
-  "version": "0.7.3",
+  "version": "0.7.4",
```
1st1, line 13: Hm, why this change?

src/client.ts
```diff
@@ -1051,6 +1063,21 @@ export class AwaitConnection {
     this.connected = false;
   }
 
+  isClosed(): boolean {
+    return !this.connected;
+  }
+
+  /** @internal */
+  setProxy(user: ConnectionProxy | null): void {
```
1st1, line 1071: Can we also prefix it with an `_`? This is *really* internal (and we'll prefix private attributes with `_`s in all public classes later anyways).

test/queues.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    done();
+  });
+
+  setTimeout(() => {
+    result = new Date().getTime();
+    queue.push(result);
+  }, 10);
+});
+
+test("LifoQueue resolves items in LIFO order", async (done) => {
```
1st1, line 37: Looks like this test doesn't need to be async

test/queues.test.ts
```diff
@@ -0,0 +1,143 @@
...
+import {LifoQueue} from "../src/queues";
+
+test("LifoQueue `get` awaits for items in the queue", async (done) => {
+  let result = -1;
+  const queue = new LifoQueue<number>();
+
+  queue.get().then((value: number) => {
+    expect(value).toBe(result);
+
+    done();
```
1st1, line 28: Hm... can we rewrite this in async/await terms? I don't understand if this `done()` is here by mistake and should be moved to the below `setTimeout` closure, or if it's here on purpose.

test/queues.test.ts
1st1: Better: rewrite it to use async/await.

src/pool.ts
```diff
@@ -0,0 +1,729 @@
+/*!
+ * This source file is part of the EdgeDB open source project.
+ *
+ * Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
```
1st1, line 4:
```suggestion
 * Copyright 2020-present MagicStack Inc. and the EdgeDB authors.
```

src/pool.ts
```diff
@@ -0,0 +1,729 @@
...
+
+import connect, {AwaitConnection, QueryArgs} from "./client";
+import {ConnectConfig} from "./con_utils";
+import * as errors from "./errors";
+import {LifoQueue} from "./queues";
+import {Set} from "./datatypes/set";
+
+export class Deferred<T> {
+  private _promise: Promise<T>;
+  private resolve?: (value?: T | PromiseLike<T> | undefined) => void;
```
1st1, line 27: Let's have some consistency here and prefix all  privates with _

src/pool.ts
```diff
@@ -0,0 +1,729 @@
...
+/**
+ * A connection pool.
+ *
+ * Connection pool can be used to manage a set of connections to the database.
+ * Connections are first acquired from the pool, then used, and then released
+ * back to the pool. Once a connection is released, it's reset to close all
+ * open cursors and other resources *except* prepared statements.
+ * Pools are created by calling :func:`~edgedb.pool.create`.
+ */
+export class Pool {
```
1st1, line 358: Can you also add a `CallbackPool` implementation that would provide a callback based API for non async/await code? See the `client.ts` second `Connection` class.

test/pool.test.ts
```diff
@@ -0,0 +1,701 @@
+/*!
+ * This source file is part of the EdgeDB open source project.
+ *
+ * Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
```
1st1, line 4: 2020

src/queues.ts
```diff
@@ -0,0 +1,75 @@
+/*!
+ * This source file is part of the EdgeDB open source project.
+ *
+ * Copyright 2019-present MagicStack Inc. and the EdgeDB authors.
```
1st1, line 4: 2020

src/pool.ts
```diff
@@ -0,0 +1,729 @@
...
+
+  static async create(options: PoolOptions): Promise<Pool> {
+    const pool = new Pool(options);
+    await pool.initialize();
+    return pool;
+  }
+
+  private validateSizeParameters(minSize: number, maxSize: number): void {
+    if (maxSize <= 0) {
+      throw new errors.InvalidValueError(
```
1st1, line 436: Should be `errors.InterfaceError`.

In general, most of exceptions defined in errors (coming from [`edb/api/errors.txt`](https://github.com/edgedb/edgedb/blob/master/edb/api/errors.txt)) are expected to be raised from the EdgeDB *server*.

We have a few errors reserved for *client side* errors, such as errors caused by improper use of our APIs:

```
#### Suggested errors for EdgeDB clients

0x_FF_00_00_00   ClientError
0x_FF_01_00_00   ClientConnectionError

0x_FF_02_00_00   InterfaceError

0x_FF_02_01_00   QueryArgumentError
0x_FF_02_01_01   MissingArgumentError
0x_FF_02_01_02   UnknownArgumentError
0x_FF_02_01_03   InvalidArgumentError

0x_FF_03_00_00   NoDataError
```

package.json
RobertoPrevato: I thought we would publish a new version of the package to npm, but I made a mistake not asking. Besides, I see there is already 0.7.4 in npmjs.com. I revert to the previous value.

1st1: That's ok! I think `yarn publish` increments the version itself, so basically we never need to change it manually in package.json.

docs/api/connection.rst
RobertoPrevato: No, indentation was wrong, the bracket is right. I am fixing it. :)

src/pool.ts
RobertoPrevato: So maybe for consistency, I will rename the current `Pool` into `AwaitPool`, and have the `Pool` class work using callbacks. To follow: `AwaitConnection` - `Connection`; `AwaitPool` - `Pool`.

1st1: I actually want to rename AwaitConnection to Connection, and Connection to CallbackConnection ;) So I'd keep Pool as is.

test/queues.test.ts
RobertoPrevato: Done here is right, because `get()` gets resolved after an item is pushed to the queue. I am rewriting to use async/await syntax.

src/pool.ts
RobertoPrevato: @1st1 I propose this solution to implement the `CallbackPool`:
instead of having code repetition of a second class that looks almost the same to the existing `Connection` in client.ts, I propose to add a common interface with methods that exist in both `Connection` and `Pool` object, I called it "IConnection"; then a generic class that expects `T extends IConnection`. https://github.com/edgedb/edgedb-js/pull/52/commits/d67f6aff6fb26c3828ea4f4b8c7c83ecab627aa6

This way we don't have code repetition. This approach would be a perfectly legitimate way of working with C#, and TypeScript generics look like ~1:1 replica of C# generics.

Classes defined this way can also be extended if desired, if the constructor is changed to be `protected` instead of `private`. For example:

```ts
export class CallbackPool extends CallbackConnectionBase<Pool> {

  constructor(pool: Pool) {
    super(pool);
  }

  example(callback: NodeCallback<null> | null = null): void {
    this.conn
      .close()
      .then((_value) => (callback ? callback(null, null) : null))
      .catch((error) => (callback ? callback(error, null) : null));
  }
}
```

src/pool.ts
```diff
@@ -0,0 +1,752 @@
...
+          "before using it."
+      );
+    }
+
+    if (this._closed) {
+      throw new errors.InterfaceError("The pool is closed");
+    }
+  }
+
+  async acquire(): Promise<PoolConnectionProxy> {
```
1st1, line 560: This and other pool-specific methods should also be exposed in CallbackPool.

src/pool.ts
1st1: Yeah, that works

src/pool.ts
RobertoPrevato: I added a `CallbackPool` with the public methods from `Pool`, added tests, and documented it. I also added a `createPool` function counterpart of `connect` function. The only thing I needed to modify to work with generics and achieve DRY, was to remove the static method `wrap` and modify `Connection` constructor to be public. So instead of:
```
Connection.wrap(awaitConnection)

new Connection(awaitConnection)
```
In the process I corrected a few typos in the documentation (mystiped connection / throw / function).

docs/api/connection.rst
```diff
@@ -339,10 +339,339 @@ Connection
...
+            } finally {
+                await pool.release(proxy);
+            }
+
+    .. js:method:: release(connectionProxy: PoolConnectionProxy)
+
+        Release a previously acquired connection proxy, to return it to the
+        pool.
+
+    .. js:method:: acquireAndExecute<T>(action: func)
```
1st1, line 542: I'd call this method `pool.run(...)`.

src/pool.ts
```diff
@@ -0,0 +1,789 @@
...
+        throw error;
+      }
+    }
+
+    // Free this connection holder and invalidate the
+    // connection proxy.
+    await this._release();
+  }
+
+  async waitUntilReleased(): Promise<void> {
```
1st1, line 222: I think this one should be `/** @internal */` and potentially prefixed with `_`

src/pool.ts
```diff
@@ -0,0 +1,789 @@
...
+
+  /** @internal */
+  public get generation(): number {
+    return this._generation;
+  }
+
+  /** @internal */
+  public get holders(): PoolConnectionHolder[] {
+    return this._holders;
+  }
```
1st1, line 381: Are these getters used anywhere? Why are they public?

src/pool.ts
```diff
@@ -0,0 +1,789 @@
...
+    // Let the connection do its internal housekeeping when it's released.
+    const connectionHolder = connectionProxy.holder;
+    return await connectionHolder.release();
+  }
+
+  /**
+   * Acquires a connection and executes a given function, returning its return
+   * value. The connection is released once done.
+   */
+  async acquireAndExecute<T>(
```
1st1, line 606: Let's rename to "run"

docs/api/connection.rst
```diff
@@ -339,10 +339,339 @@ Connection
...
+
+.. _edgedb-js-api-pool:
+
+Pool
+==========
+
+.. js:function:: createPool(options)
+                 createPool(options, callback)
+
+    Create a connection pool to an EdgeDB server.
```
1st1, line 354: A quick async/await complete example would be cool: how to create a pool, how to run a `SELECT 1` in it, how to close the pool.

docs/api/connection.rst
```diff
@@ -339,10 +339,339 @@ Connection
...
+
+        async function main() {
+            const pool = await edgedb.createPool({
+                connectOptions: {
+                    user: "edgedb",
+                    host: "127.0.0.1",
+                },
+            });
+
+            await pool.initialize();
```
1st1, line 421: Why is this necessary? Why can't `createPool` take care of this?

1st1: Although it seems that `initialize()` is now protected and this looks like a line we just need to drop, right?

src/pool.ts
RobertoPrevato: You are right, we don't need these. I have the tendency of leaving things public and open, because I think: "You never know what users will need to do" (I like to think that Python doesn't have private modifier for a similar reason). I will remove these public getters :)

RobertoPrevato: On a second thought, having the queue public might have a good use case: consider if somebody implements telemetry and wants to monitor the size of the queue. And `generation` is used by the ConnectionHolder, so it makes sense to leave it as public get only property (public is actually redundant since things are public by default anyway in TS)

docs/api/connection.rst
RobertoPrevato: I added an example to `usage.rst` file. Should I add the same example to `connection.rst`?

docs/api/connection.rst
RobertoPrevato: Yes, it's a mistake.

src/pool.ts
RobertoPrevato: The only way I was using `holders` was to verify that a pool is created with a `minSize` of open connections. This seems a valid use case to me, too.

```ts
describe("pool.initialize: creates minSize count of connections", () => {
  each([0, 1, 5, 10, 20]).it(
    "when minSize is '%s'",
    async (minSize) => {
      const maxSize = minSize + 50;
      const pool = await Pool.create({
        connectOptions: getConnectOptions(),
        minSize,
        maxSize,
      });

      expect(pool.queue).toHaveLength(maxSize);

      const itemsWithOpenConnection = pool.holders.filter(
        (holder) => holder.connection !== null
      );

      expect(itemsWithOpenConnection).toHaveLength(minSize);
```

RobertoPrevato: Now I recall one of the reasons why I left `holders` and `holder` public: the `ConnectionProxy` is exposed by `onAcquire` and `onRelease` callbacks, and the proxy's holder is read inside a pool, to ensure that the proxy belongs to the pool where it's being released.
1. I used public getter because I wanted certain properties to be read-only, and didn't take into consideration something like:

```ts
private __something: T;  // *really* private

/** @internal */
get _something(): T {
  return this.__something;
}
```

2. An alternative would be to replace the get-only property with a method like:

```
/** @internal */
_getHolder() {
  return this._holder;
}
```

docs/api/connection.rst
```diff
@@ -539,7 +537,7 @@ Pool
         Release a previously acquired connection proxy, to return it to the
         pool.
 
-    .. js:method:: acquireAndExecute<T>(action: func)
+    .. js:method:: run<T>(action: func)
```
1st1, line 540: What's `<T>` here?

src/pool.ts
```diff
@@ -278,7 +279,7 @@ export class PoolConnectionProxy {
     connection._setProxy(this);
   }
 
-  public get connection(): AwaitConnection {
+  get connection(): AwaitConnection {
```
1st1, line 282: This is still public (IIRC `public` is implicit, and to make `private` you always add `private` explicitly)

1st1: It shouldn't be possible for users to unwrap real connections here, so I'd say this should be private.

docs/api/connection.rst
RobertoPrevato: You're right, I didn't include it in the documentation. In this context, is the same value returned by the `func` that receives a connection.

1st1: Yeah, I suggest including the full type info here.

src/pool.ts
RobertoPrevato: I handled this a bit differently than how it is done in Python. In the Python driver, the metaclass `PoolConnectionProxyMeta` is used to copy all methods from the connection class into the proxy class. `acquire` returns a proxy, `release` accepts a proxy (this is the same in both drivers). So if the user acquires a proxy, it will behave like a connection object while not exposing the underlying connection.

I know this kind of dynamism is possible in JavaScript and TypeScript, but I wasn't sure whether this is a common approach when working with TS. I kept the proxy class simpler, and exposed the connection.

The difference in practice is:
```python
    async def execute(self, query):
        async with self.acquire() as con: # <-- here "con" is a connection proxy
            return await con.execute(query)
```

```ts
  async run<T>(
    action: (connection: AwaitConnection) => Promise<T>
  ): Promise<T> {
    const proxy = await this.acquire();

    try {
      return await action(proxy.connection);  // <-- here the actual connection is used to execute
    } finally {
      await this.release(proxy);
    }
  }

    return await this.run(async (connection) => {
      return await connection.execute(query);
    });
  }
```

I wanted to speak with you about this detail, then I forgot while asking about other details.
If you prefer to make the connection private, then I propose to add the 5 methods to the proxy class (fetchOne, fetchOneJSON, fetchAll, fetchAllJSON, execute).

1st1: > I kept the proxy class simpler, and exposed its connection.

Yeah, we have to follow Python's design here.

> In the Python driver, the metaclass PoolConnectionProxyMeta is used to copy all methods from the connection class into the proxy class.  [...]  I know this kind of dynamism is possible in JavaScript and TypeScript, but I wasn't sure whether this is a common approach when working with TS. 

We don't need to replicate metaclasses or employ other magic, like introspecting prototypes. The methods can simply be hardcoded. The `ProxyConnection` type should just implement `IConnection` interface, which will give us a heads up when we add new APIs and forget to add them to `ProxyConnection`.

The connection should be private. The whole point is that it's impossible for the user to accidentally use a released pool connection. If the connection is exposed, nothing precludes the user from storing a reference to the wrapped connection object and using it later when it's acquired by some other call site.

test/pool.test.ts
```diff
@@ -298,6 +298,7 @@ test(
     async function user(pool: Pool): Promise<void> {
       const proxy = await pool.acquire();
 
+      // @ts-ignore
```
1st1, line 301: I'd suggest using `@ts-ignore` on as small operations as possible.  E.g. instead of 

```
// @ts-ignore
if (connections.indexOf(proxy.connection) === -1) {
```

do

```
// @ts-ignore
const proxyCon = proxy.connection;

if (connections.indexOf(proxyCon) === -1) {
```


This way you're minimizing the impact of disabling typing.

RobertoPrevato: Thanks for the heads up. By the way, what do you think about using @ts-ignore to keep the `connectionHolder`'s proxy private?

```ts
    // @ts-ignore
    if (connectionProxy.holder.pool !== this) {
```

I like it on one hand, but I feel it's a compromise.

1st1: I had the same thought. Let's do that. Essentially we'd be emulating C++'s friendly classes this way.

src/pool.ts
RobertoPrevato: I modified the code to keep the connection private. :+1: I also made the queue private, and finally included a `PoolStats` returning the list of pending consumers + number of open connections. We will probably do a second pass over the pool stats when we have occasion to focus on this subject.

test/pool.test.ts
1st1: OTOH we can do differently by using symbols and not resorting to `@ts-ignore`:

```
const getConnection = Symbol.for('connection getter');

class Foo {
  [getConnection] = () => {
  }
}
```

Now because `getConnection` is not exported it can be used privately in our edgedb codebase, but everyone else wouldn't be able to access it.

1st1: This gives you the benefit of continuing to use strict typing.

PR #53, opened 2020-07-13
Replace Spectrum badge with GitHub discussions

Fogapod:
As Yury suggested in https://github.com/edgedb/edgedb/pull/1531


1st1:
Thank you!

PR #54, opened 2020-07-14
Rename "fetch*" methods to "query*"

1st1:
See https://github.com/edgedb/edgedb-python/issues/51 for
the relevant discussion.


1st1:
The change is backwards compatible https://github.com/edgedb/edgedb-js/pull/54/files#diff-3cafb86a5b0a9d0916a7fed543a1fb53R1294-R1355

PR #61, opened 2020-08-10
Add a transaction api

RobertoPrevato:


RobertoPrevato:
I have doubts about some details:
- whether we should add a method to the `Connection` interface that returns the `Transaction` object, to expose it to the user. Currently it is managed and not exposed by the `transaction` method, that accepts function as input and has the same return object of the function argument.
- whether we should define a `Transaction` interface in `ifaces` for consistency, and use it instead of the current `Transaction` class. But then I am not sure how could we import `TransactionImpl` in `connection.ts` without making it importable by the end user?

RobertoPrevato:
:boom: there is a complication with the `pool.transaction` method. The PR is not ready yet.

src/pool.ts
```diff
@@ -689,6 +698,24 @@ class PoolImpl implements Pool {
...
+    try {
+      result = await action();
+      await transaction.commit();
+    } catch (err) {
+      await transaction.rollback();
+
+      throw err;
+    }
+    return result;
+  }
```
1st1, line 717: Yeah, this needs to be fixed to explicitly acquire a connection from the pool, call `.transaction` on it, and then release the connection back to the pool.

1st1:
>     I don't understand if the code api for the Connection interface is truly working for connection pools (Pool.transaction method), when we cannot be sure that the same connection is used across different method calls of the Pool. The test I added for the Pool is passing, but I don't understand if it's correct - honestly I don't even understand why it is passing.

I replied in https://github.com/edgedb/edgedb-js/pull/61#pullrequestreview-474931860

> whether we should add a method to the Connection interface that returns the Transaction object, to expose it to the user. 

I don't think we need to do that at this stage. We can expose it in later versions.

src/pool.ts
RobertoPrevato: Will this be sufficient? I suspect not. If we want to support the same code api:

```
      await pool.transaction(async () => {
        await pool.execute(`
          INSERT ${typename} {
            name := 'Test Transaction'
          };
        `);

        // how can we be sure that the same connection is used in this second call to execute?
        await pool.execute(`
          INSERT ${typename} {
            name := 'Test Transaction'
          };
        `);
        await pool.execute("SELECT 1 / 0;");
      });
```

Wouldn't something like this be needed? (Looks a bit ugly and potentially confusing, though..)

```
      await pool.transaction(async (connectionInTransaction) => {
        await connectionInTransaction.execute(`
          INSERT ${typename} {
            name := 'Test Transaction'
          };
        `);

        await connectionInTransaction.execute(`
          INSERT ${typename} {
            name := 'Test Transaction'
          };
        `);
        await connectionInTransaction.execute("SELECT 1 / 0;");
      });
```

1st1: This wouldn't work for sure:

```
      await pool.transaction(async () => {
        await pool.execute(`
```

so this is indeed the only way:

```
      await pool.transaction(async (con) => {
        await con.execute(`
```

We could, perhaps, prohibit using `pool` while the transaction closure runs... but I don't think we do that in our Python binding or in asyncpg and they seem to be OK.

@elprans what do you think?

1st1: Actually, I don't see neither asyncpg nor edgedb-python implementing transaction helper for the Pool object. So we shouldn't do that for the JS binding. Just remove the `Pool.transaction` helper.

elprans: Yes, `transaction` only makes sense on a specific connection.

src/client.ts
```diff
@@ -1022,6 +1024,24 @@ class ConnectionImpl implements Connection {
...
 
+  async transaction<T>(
+    action: () => Promise<T>,
+    options?: TransactionOptions
+  ): Promise<T> {
+    let result: T;
+    const transaction = new Transaction(this, options);
+    await transaction.start();
+    try {
+      result = await action();
```
1st1, line 1035: @elprans @RobertoPrevato We should pass the transaction object into the async closure, like this:


```
await con.transaction(async (tr) {
  ...
  if (something) {
    await conn.rollback();
  }
});
```

1st1: Although we don't currently do that in asyncpg or in edgedb-python...

RobertoPrevato: @1st1; good point, I was also thinking that it would be nice to expose the transaction object (I also wrote this in my comment above a few days ago). I'm checking if TS supports defining overload by function argument's signature as it would be possible in C# or Java. Maybe we could manage the Transaction object only when the user passes a callback that doesn't expect any argument, and instead don't manage the transaction object when the user passes a callback that expects a transaction object as input.

PR #62, opened 2020-08-14
Implement credentials_file support

tailhook:
Also deprecate `admin: true` and `edgedbadmin:` protocol, as
they aren't needed if proper permissions are set up. And was introduced
mainly for command-line tools.


src/con_utils.ts
```diff
@@ -79,6 +81,7 @@ export function parseConnectArguments(
 
 function parseConnectDsnAndArgs({
   dsn,
+  credentials_file,
```
1st1, line 84:
```suggestion
  credentialsFile,
```

tailhook: Fixed

tailhook: BTW, why lint doesn't enforce this?

src/con_utils.ts
```diff
@@ -280,6 +280,9 @@ function parseConnectDsnAndArgs({
       } else {
         host = ["/run/edgedb", "/var/run/edgedb"];
       }
+      if (!admin) {
```
tailhook, line 283: @1st1 this is how respective code in python looks like. Was this omission intentional?

1st1:
I'll take another look once https://github.com/edgedb/edgedb-rust/issues/45 is settled.

PR #63, opened 2020-08-21
Implement named instances support

tailhook:
This adds positional argument to `connect()` and `createPool()` functions which can be either DSN or an instance name.

This is a part of what's discussed in https://github.com/edgedb/edgedb/discussions/1683 other things will follow in the next release.

This also deprecates `admin=True` parameter and `edgedbadmin` schema, as
they aren't needed if proper permissions are set up. And were introduced
mainly for command-line tools.

Supersedes #62

### DIFF
--- a/docs/api/connection.rst
+++ b/docs/api/connection.rst
@@ -111,6 +111,20 @@ Connection
 
         main();
 
+    .. note::
+
+       For compatibility this function also supports passing options as
+       the first argument:
+
+       .. code-block:: js
+
+          await connect({host: 'localhost', port: 5656})
+          // or
+          await connect({dsn: 'edgedb://localhost'})
+
+       But this form is deprecated and will be removed in the future.
+
+
 .. js:class:: Connection
 
     A representation of a database session.
@@ -279,6 +293,19 @@ Pool
     :returns:
         Returns a ``Promise`` of an :js:class:`Pool` is returned.
 
+    .. note::
+
+       For compatibility this function also supports passing options as
+       the first argument:
+
+       .. code-block:: js
+
+          await createPool({
+            maxSize: 10,
+            connectOptions: {dsn: 'edgedb://localhost'},
+          })
+
+       But this form is deprecated and will be removed in the future.
 
 .. js:class:: Pool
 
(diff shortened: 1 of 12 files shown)


docs/api/connection.rst
```diff
@@ -9,19 +9,21 @@ API
 Connection
 ==========
 
-.. js:function:: connect(options)
+.. js:function:: connect(dsn, options)
```
1st1, line 12: Paul, I don't like breaking the API just yet. I'd do this differently: let's make `options` to be of type `ConnectConfig | string | null`.

We can then gradually introduce "instances" - they'll be a separate JS type that we will be able to check with an `instanceof` - and ban plain objects (currently `ConnectConfig`)

tailhook: Well, then you can't set timeout. Or the API would look like:
```
connect(name_or_dsn_or_options: string | ConnectOptions | null, options: ConnectOptions | null)
```
(and then throw error if both are objects?) I.e. jQuery-age libraries had a lot of API looking like this, but in 2020... 

And then what to do with `createPool`? This function without options is next to useless.

1st1: As discussed on the call, let's try:

Support all `connect(instanceName)`, `connect(ConnectConfig)` and `connect(instanceName, ConnectConfig)` use cases. The signature will likely look like this:

`connect(ConnectConfig | string | null, ConnectConfig | null)`

The code will need to validate that only one `ConnectConfig` is actually passed.

We'll start issuing deprecation warnings once we fully agree on the new instance API and start implementing it.

tailhook:
Updated the code. Docs insist on the new signature and mention compatible form as a side note.

docs/api/connection.rst
```diff
@@ -109,6 +111,20 @@ Connection
...
+       For compatibility this function also supports passing options as
+       the first argument:
+
+       .. code-block:: js
+
+          await connect({host: 'localhost', port: 5656})
+          // or
+          await connect({dsn: 'edgedb://localhost'})
+
+       But this form is deprecated and will be removed in the future.
```
1st1, line 125: 👍

PR #65, opened 2020-09-02
Add new "RawConnection" class

1st1:
The class allows to use low-level "parse" and "execute" protocol messages
directly, using binary data.  This is a private API needed to EdgeDB
Studio.


1st1:
@jaclarke feel free to do a post-commit review of this one.

PR #67, opened 2020-09-14
Add retry in case of ECONNRESET

RobertoPrevato:
#### Context:
We are experiencing transient failures in one application connecting to an EdgeDB Server:

```js
at TCP.onStreamRead (internal/stream_base_commons.js:205:27) {
  errno: -104,
  code: 'ECONNRESET',
  syscall: 'read'
}
```
I cannot recreate this error when running the application locally, even connecting to an EdgeDB Server in AWS ECS. But this error is happening frequently, about once every 5 ~ 10 minutes of inactivity, when both server and application are running in AWS ECS.

A way to recreate this error (originating from the exact same place in code) is, for example, to simulate an exception in a socket server:

##### server.js
```js
const net = require("net");

const server = net.createServer(function (socket) {
  throw new Error("Crash!");
})

server.listen(6666, function () {
  console.info("Listening...");
});
```

##### client.js
```js
const net = require("net");

const socket = net.createConnection({ port: 6666 }, () => {
  socket.write("Hi There!\r\n");
})

socket.on("error", (data) => {
  console.info("************", data);
  console.info("-->", data.code);
  console.log(data.toString());
  socket.end();
});
```

```bash
$ node client.js 
************ Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:205:27) {
  errno: -104,
  code: 'ECONNRESET',
  syscall: 'read'
}
--> ECONNRESET
```

---
Aside from investigating and resolving the source that is causing `ECONNRESET`, I think the client might automatically attempt a retry in such cases (maybe adding also a delay between attempts).

Do you think something like the change I am proposing would make sense?

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1111,7 +1111,8 @@ class ConnectionImpl implements Connection {
   /** @internal */
   static async connect(
     dsn?: string | ConnectConfig | null,
-    options?: ConnectConfig | null
+    options?: ConnectConfig | null,
+    attempt: number = 0
   ): Promise<ConnectionImpl> {
     let config: ConnectConfig | null = null;
 
@@ -1156,6 +1157,15 @@ class ConnectionImpl implements Connection {
         await connPromise;
       } catch (e) {
         conn._abort();
+
+        // in case of ECONNRESET, try again
+        if (typeof e.code === "string" && e.code === "ECONNRESET") {
+          if (attempt < 3) {
+            return await ConnectionImpl.connect(dsn, options, attempt + 1);
+          } else {
+            throw e;
+          }
+        }
 
         // on timeout Error proceed to the next address, otherwise re-throw
         if (


1st1:
@elprans thoughts?

elprans:
I suspect this is related to TCP keepalive.  We should make it configurable (and enable by default).  This applies to all clients.  (and we should also make sure we do it on the backend connections to Postgres).

1st1:
Hm, good point.

elprans:
Actually, we don't have to do it from the client side.  Server can configure its sockets instead (and Postgres has `tcp_keepalives_idle`).

src/client.ts
```diff
@@ -1157,6 +1158,15 @@ class ConnectionImpl implements Connection {
       } catch (e) {
         conn._abort();
 
+        // in case of ECONNRESET, try again
+        if (typeof e.code === "string" && e.code === "ECONNRESET") {
```
tailhook, line 1162: I don't think typeof check is needed when you use `===`

tailhook:
This is what will be solved more generally by implementing https://github.com/edgedb/edgedb/discussions/1859

src/client.ts
RobertoPrevato: You are right, that's a mistake. If any check, it should check if `e` is undefined. Anyway I can close this PR, since Elvis recommended to not handle this in the client.

RobertoPrevato:
Closing since it has been decided to handle this on the server side.

PR #68, opened 2020-09-16
Add EdgeDBDateTimeCodec

jaclarke:
Related issue: https://github.com/edgedb/edgedb/issues/1769#issuecomment-692379938

### DIFF
--- a/src/compat.ts
+++ b/src/compat.ts
@@ -16,6 +16,9 @@
  * limitations under the License.
  */
 
+import * as bi from "./bigint";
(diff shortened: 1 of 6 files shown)


src/compat.ts
```diff
@@ -106,3 +109,55 @@ export function decodeInt64ToString(buf: Buffer): string {
 
   return negative ? `-${result}` : result;
 }
+
+export interface CustomDateTime {
```
1st1, line 113: I'd call it `EdgeDateTime`; it's not so much custom as it is a direct representation of what's in the DB.

1st1 (approved):

PR #76, opened 2020-10-26
Move request methods into Executor interface

tailhook:
Part of #1896

### DIFF
--- a/src/ifaces.ts
+++ b/src/ifaces.ts
@@ -55,12 +55,15 @@ export interface TransactionOptions {
   readonly?: boolean;
 }
 
-export interface Connection {
+interface Executor {
   execute(query: string): Promise<void>;
   query(query: string, args?: QueryArgs): Promise<Set>;
   queryJSON(query: string, args?: QueryArgs): Promise<string>;
   queryOne(query: string, args?: QueryArgs): Promise<any>;
   queryOneJSON(query: string, args?: QueryArgs): Promise<string>;
+}
+
+export interface Connection extends Executor {
   transaction<T>(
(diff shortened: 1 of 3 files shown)


src/ifaces.ts
```diff
@@ -55,12 +55,26 @@ export interface TransactionOptions {
...
-export interface Connection {
+export interface ReadOnlyExecutor {
   execute(query: string): Promise<void>;
   query(query: string, args?: QueryArgs): Promise<Set>;
   queryJSON(query: string, args?: QueryArgs): Promise<string>;
   queryOne(query: string, args?: QueryArgs): Promise<any>;
   queryOneJSON(query: string, args?: QueryArgs): Promise<string>;
+}
+
+export const allow_modifications = Symbol();
```
1st1, line 66:
```suggestion
export const ALLOW_MODIFICATIONS = Symbol();
```

src/ifaces.ts
```diff
@@ -55,12 +55,26 @@ export interface TransactionOptions {
...
+
+interface Modifiable {
```
1st1, line 68: MutatingExecutor?

1st1 (approved):
Please address the comments before merging

src/ifaces.ts
tailhook: We use "modifications" rather than "mutations" in RFC1004. Should we change that?

Also it's not executor, it becomes executor when combining `ReadOnlyExecutor & Modifiable`. So this is kinda "aspect" kind of type. But it's internal anyway, so I'm not going to fight for the name too much.

PR #77, opened 2020-11-03
Add github action to deploy site on changes to documentation

jaclarke:


colinhacks:
Any reason we haven't merged this @jaclarke?

jaclarke:
Originally versioning concerns (https://github.com/edgedb/edgedb-python/pull/126), but I think the new plan to always build the latest docs from the master branch and use some kind of version directives resolves that?

PR #79, opened 2020-12-03
Fixes forward compability with the new edgedb

tailhook:
With how it's written before the change server returns the largest
protocol it supports and as long as server is updated JavaScript
bindings crash with unsupported protocol error.

This, together with edgedb/edgedb#2019 fixes tests with nightly edgedb.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,12 +39,12 @@ jobs:
       - name: Install EdgeDB
         env:
           OS_NAME: ${{ matrix.os }}
-          SLOT: 1-alpha4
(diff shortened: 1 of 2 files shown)


1st1 (approved):

.github/workflows/tests.yml
```diff
@@ -39,12 +39,12 @@ jobs:
       - name: Install EdgeDB
         env:
           OS_NAME: ${{ matrix.os }}
-          SLOT: 1-alpha4
+          SLOT: 1-alpha6-dev5104
```
elprans, line 42: The latest nightly is actually `1-alpha7-dev5259`

PR #81, opened 2020-12-11
Use underscore sys::_sleep function

tailhook:


1st1 (approved):
Do we already run JS tests in tsetmode?

PR #82, opened 2020-12-14
Implement reconnect and wait-until-available

tailhook:
I'm not sure how to test this, so I've only included tests for proper error conversion and tested manually.

We probably should test this more thoughtfully when we made actual end-to-end application tests which includes network split and server restart conditions.


tailhook:
Test are crashing because of #83

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+      );
+    }
+    config = {...dsn, ...options};
+  }
+  return StandaloneConnection.connect(parseConnectArguments(config));
+}
+
+function sleep(durationMicros: number): Promise<void> {
+  return new Promise((accept, reject) => {
+    setTimeout(() => accept(), durationMicros);
```
1st1, line 100: setTimeout's argument is in milliseconds.

1st1: But I suggest to not include timeout type in the name -- in JS ecosystem all timeouts by default are milliseconds, so suffixes like Milli or Micro just add noise.

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+    if (proxy != null) {
+      proxy[onConnectionClose]();
+      proxyMap.delete(this);
+    }
+  }
+  isClosed(): boolean {
+    return this._isClosed;
+  }
+  async execute(query: string): Promise<void> {
+    return await (await this.connection()).execute(query);
```
1st1, line 143: I'd like to avoid having one extra await here (for perf reasons) and replace this with something like this, in pseudocode:

```
let connection: Connection | null = this._tryGetConnection();
if (connection == null) {
  connection = await this._getConnection();
}
```

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+  async queryJSON(query: string, args?: QueryArgs): Promise<string> {
+    return await (await this.connection()).queryJSON(query, args);
+  }
+  async queryOne(query: string, args?: QueryArgs): Promise<any> {
+    return await (await this.connection()).queryOne(query, args);
+  }
+  async queryOneJSON(query: string, args?: QueryArgs): Promise<string> {
+    return await (await this.connection()).queryOneJSON(query, args);
+  }
+  async connection(): Promise<ConnectionImpl> {
```
1st1, line 157: This should be a private function (with `private` modifier), and I'd also rename it to `_getConnection`

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+    if (this._isClosed) {
+      throw new errors.InterfaceError("Connection is closed");
+    }
+    if (this._connection && !this._connection.isClosed()) {
+      return this._connection;
+    }
+    const max_time = process.hrtime.bigint()
+      + BigInt(Math.ceil((this.config.waitUntilAvailable || 0) * 1000_000));
```
1st1, line 165: Nit, imo this reads better

```suggestion
      + BigInt(Math.ceil((this.config.waitUntilAvailable || 0) * 1_000_000));
```

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+    return await (await this.connection()).queryOneJSON(query, args);
+  }
+  async connection(): Promise<ConnectionImpl> {
+    if (this._isClosed) {
+      throw new errors.InterfaceError("Connection is closed");
+    }
+    if (this._connection && !this._connection.isClosed()) {
+      return this._connection;
+    }
+    const max_time = process.hrtime.bigint()
```
1st1, line 164:
```suggestion
    const maxTime = process.hrtime.bigint()
```

We don't use snake_case in JS code except in constants.

src/client.ts
1st1: But why do you want to use high-res bigint timing here? IMO the precision of milliseconds is more than enough here, no?

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+      + BigInt(Math.ceil((this.config.waitUntilAvailable || 0) * 1000_000));
+    let iteration = 1;
+    while (true) {
+      for (const addr of this.config.addrs) {
+        try {
+          this._connection = await ConnectionImpl.connect_with_timeout(
```
1st1, line 170:
```suggestion
          this._connection = await ConnectionImpl.connectWithTimeout(
```

src/client.ts
```diff
@@ -76,7 +78,128 @@ export default function connect(
...
+              throw e;
+            }
+          } else {
+            throw e; // this shouldn't happen
+          }
+        }
+      }
+
+      iteration += 1;
+      await sleep(Math.trunc(10 + Math.random() * 200));
```
1st1, line 192: Worst case you'll have 10 milliseconds delay here which feels like way too low for any attempt. I'd suggest changing to 

```suggestion
      await sleep(Math.max(50, Math.random() * 200));
```

But then why isn't `iteration` involved in this function? Ideally attempts should get slightly slower and slower with time.

src/client.ts
```diff
@@ -360,6 +483,60 @@ class ConnectionImpl implements Connection {
...
+
+    try {
+      await connPromise;
+    } catch (e) {
+      // TODO(tailhook) wrap IO errors into EdgeDBError type
+      conn._abort();
+      if (e instanceof errors.EdgeDBError) {
+        throw e;
+      } else {
+        let err;
```
1st1, line 516:
```suggestion
        let err: Error;
```

src/errors/tags.ts
```diff
@@ -0,0 +1,18 @@
...
+  }
+  isMatching(err: EdgeDBError): boolean {
+    return this.classes.some((type) => err instanceof type);
+  }
+}
+
+export const SHOULD_RECONNECT = new ListTag([
+  errors.ConnectionFailedTemporarilyError,
+  errors.ConnectionTimeoutError,
+]);
```
1st1, line 18: IMO we need to implement this differently.

I'd modify the current `edb/api/errors.txt` to include optional string tags, like

```
0x_05_02_00_00   IntegrityError                   repeatable, other_tag
0x_05_02_00_01   ConstraintViolationError
0x_05_02_00_02   CardinalityViolationError
0x_05_02_00_03   MissingRequiredError

0x_05_03_00_00   TransactionError                 repeatable
0x_05_03_00_01   TransactionSerializationError
0x_05_03_00_02   TransactionDeadlockError
```

and modify our JS errors code gen to include them in error classes, and also create a union string type of all available tags.

And then the client API will be

```
catch (e) {
   if (e instanceof EdgeDBError && e.hasTag('repeatable')) { ... }
}
```

This would align better with https://github.com/edgedb/edgedb/discussions/1849 and the idea that "tags" will be implemented via type annotations.

src/client.ts
tailhook: It's the only monotonic clock available as far as I know.

src/client.ts
tailhook: Well, `Micros` instead of `Millis` is for some reason my common typo :( For the suffix itself, see slack discussion.

src/client.ts
tailhook: There are two cases where delay is applied:
1. It's "connection refused", which happens via single roundtrip, which on a fast link (and I expect most app -> database links be fast) is the order of `100-1000` micros. Which is orders of magnitude smaller than minimum reconnect of 10 ms, not even saying that it is 100 ms on average. 
2. Connection is dropped by timeout on authentication. That latter is usually in a matter of seconds (we have 60 seconds in current python implementation, should be down to 10 seconds I think, but the order is the same). In this case it doesn't matter how much reconnect delay is as the interval between attempts is dominated by connection timeout anyway.

So my point here is:
1. Minimum interval needed to prevent process from taking 100% CPU and huge network utilization (10 ms is enough)
2. But it should be small enough so that when database is already started we don't wait significantly more before processing requests. I.e. if database starts in 10 seconds, it would be ugly to wait +5 seconds before serving requests, that exponential back-off could add.
3. Randomization is needed just to prevent lots of connections reconnecting simultaneously. And `Math.max(50` is bad for that, because that means 25% of connections will reconnect simultaneously.
4. Our attempts are extremely cheap. This is why we exponential backoff for potentially much expensive things in  `retry` (per RFC1004 spec), but not here.

src/client.ts
tailhook: Yes, struggling to switch between languages. I still wonder why we don't have a lint for that.

src/errors/tags.ts
tailhook: I think tags have to be constants somewhere, not arbitrary strings. At least for easy discovery and to prevent typos. And I think we should postpone custom tags to later RFC when we implement custom errors.

Codegen for them is fine. But I'd rather make them opaque things for now, instead of hardcoding specific strings. I'm not sure I've achieved the goal in current implementation, though, as privacy rules of TypeScript are a bit unclear for me now. Also we can probably make them Symbols in JS?

src/client.ts
tailhook: Well, can't V8 do it for us? If we're going to inline functions manually, why don't we inline the `execute` function too, as it's kinda about the same number of lines.
```
    this._enterOp();
    try {
      await this._execute(query);
    } finally {
      this._leaveOp();
    }
  }

```
Do we have any estimation of how costly awaits are? Especially since `await` you're refering to is not nested comparing to `await` in my example which contributes to the call stack.

1st1: > Well, can't V8 do it for us? If we're going to inline functions manually, why don't we inline the execute function too, as it's kinda about the same number of lines.

I don't think so. async/await isn't yet popular enough to deserve a lot of attention from V8 engineers. And JS isn't Rust; optimizations like this are somewhat more complex when implemented in a JIT. But we're talking about just 5-6 functions here, inlining by hand is easy.

src/errors/tags.ts
1st1: > I think tags have to be constants somewhere, not arbitrary strings. At least for easy discovery and to prevent typos. 

That's why I mentioned union string type. Basically you'll reflect all tags into

```typescript
type ErrorTag = 'repeatable' | 'retriable';

...

class Error {
  hasTag(tag: ErrorTag): bool;
}
```

We can do the same in Python.

> And I think we should postpone custom tags to later RFC when we implement custom errors.

Yes, but I'd like the API to be forward compatible. Annotations are the most straightforward and already supported way of implementing tags.

We can also reflect tags to sentinels:

```
const Repeatable = Symbol();
const Retriable = Symbol();
type ErrorTag = Repeatable | Retriable;

...

class Error {
  hasTag(tag: ErrorTag): bool;
}
```

I'd be OK with that; the benefit is that this would be safer to use without typing. In this case you wouldn't need to use `instanceof` and `map` -- just a `Set().has(tag)` would work.

src/client.ts
1st1: > For the suffix itself, see slack discussion.

Yeah, seen that. Need to think a bit before replying.

src/client.ts
1st1: > It's "connection refused", which happens via single roundtrip, which on a fast link (and I expect most app -> database links be fast) is the order of 100-1000 micros. Which is orders of magnitude smaller than minimum reconnect of 10 ms, not even saying that it is 100 ms on average.

Well, 10ms is in the realm of latency between two different clouds or datacenters that's why it caught my attention as a tad too low delay.

I agree with the rest of your comments.

src/errors/tags.ts
1st1: Let's do the sentinel approach (`const Repeatable = Symbol();`)

tailhook: Fixed.

src/client.ts
tailhook: Fixed.

src/client.ts
tailhook: Done.

src/client.ts
```diff
@@ -76,7 +78,151 @@ export default function connect(
...
+    return await connection.transaction(action, options);
+  }
+  async close(): Promise<void> {
+    try {
+      if (this._connection) {
+        await this._connection.close();
+      }
+      this._connection = undefined;
+      // TODO(tailhook) it makes little sense to close the reconnecting
+      // connection so maybe deprecate this method
```
1st1, line 131: Not sure I follow this TODO comment. Can you elaborate?

src/client.ts
1st1: Ah, makes sense. Could you please add a comment explaining that and mentioning that hrtime is monotonic?

src/client.ts
elprans: Yeah, makes sense to me too.  We can always adjust this later if it turns out to be a poor choice.

1st1 (approved):
IMO this is ready. @elprans please take a quick look.

src/client.ts
tailhook: Currently:
```
await conn.close()
await conn.query('SELECT 1`)
```
Works fine, as we reconnect if connection was closed (i.e. when it is closed because of network error).

As it still closes underlying connection it may be helpful, though. So yeah, perhaps this comment is useless.

1st1: > Works fine, as we reconnect if connection was closed (i.e. when it is closed because of network error).

But it shouldn't. If the connection was explicitly closed by user it should not reconnect, it should be marked as closed and all operations on it should be rejected. It's a basic contract that has existed since forever and I don't think we are ought to break it.

elprans (approved):

PR #83, opened 2020-12-14
Use `edgedb-setup` action and test on stable and nightly

tailhook:


elprans (approved):

tailhook:
@elprans, need to update required checks now. I don't have permissions.

PR #84, opened 2021-01-14
RFC1004: Add `try_transaction` and `retry`

tailhook:
Like in python PR edgedb/edgedb-python#146 this lacks documentations and also more robustness can be added later. But this provides correct future-proof API.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -76,7 +76,7 @@ enum TransactionStatus {
 
 const DEFAULT_MAX_ITERATIONS = 3;
 
-function default_backoff(attempt: number) {
+function default_backoff(attempt: number): number {
   return 2 ** attempt * 100 + Math.random() * 100;
 }
 
@@ -268,8 +268,8 @@ class StandaloneConnection implements Connection {
     }
     return await connection.queryOneJSON(query, args);
   }
-  _check_borrow() {
-    let borrow = this[BORROW];
+  _check_borrow(): void {
+    const borrow = this[BORROW];
     if (borrow) {
       let text;
       switch (borrow) {
--- a/src/client.ts
+++ b/src/client.ts
@@ -185,7 +185,7 @@ class StandaloneConnection implements Connection {
     let result: T;
     for (let iteration = 0; iteration < DEFAULT_MAX_ITERATIONS; ++iteration) {
       const transaction = new Transaction(this);
-      await transaction[START_TRANSACTION_IMPL](iteration != 0);
+      await transaction[START_TRANSACTION_IMPL](iteration !== 0);
       try {
         result = await action(transaction);
       } catch (err) {
@@ -196,7 +196,7 @@ class StandaloneConnection implements Connection {
             // We ignore EdgeDBError errors on rollback, retrying
             // if possible. All other errors are propagated.
             throw rollback_err;
-          }          
+          }
         }
         if (
           err instanceof errors.EdgeDBError &&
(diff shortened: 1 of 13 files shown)


src/pool.ts
```diff
@@ -273,12 +274,12 @@ class PoolConnectionHolder {
   }
 }
 
-const holderAttr = Symbol.for("holder");
-const detach = Symbol.for("detach");
-export const getHolder = Symbol.for("getHolder");
-const connectionAttr = Symbol.for("connection");
-export const unwrapConnection = Symbol.for("unwrap");
-const isDetached = Symbol.for("isDetached");
+const holderAttr = Symbol("holder");
```
1st1, line 277: Good catch!

src/ifaces.ts
```diff
@@ -75,10 +83,16 @@ interface Modifiable {
 export type Executor = ReadOnlyExecutor & Modifiable;
 
 export interface Connection extends Executor {
+  [BORROW]?: BorrowReason;
```
1st1, line 86:
```suggestion
  [BORROWED_FOR]?: BorrowReason;
```

src/client.ts
```diff
@@ -145,40 +229,59 @@ class StandaloneConnection implements Connection {
...
   async query(query: string, args?: QueryArgs): Promise<Set> {
+    this._check_borrow();
     let connection = this._connection;
     if (!connection || connection.isClosed()) {
       connection = await this._reconnect();
     }
     return await connection.query(query, args);
   }
   async queryJSON(query: string, args?: QueryArgs): Promise<string> {
+    this._check_borrow();
```
1st1, line 248: JS JITs don't inline functions that raise errors. So I'd guard with `if (this[BORROW]) {`.

src/client.ts
```diff
@@ -145,40 +229,59 @@ class StandaloneConnection implements Connection {
     return this._isClosed;
   }
   async execute(query: string): Promise<void> {
```
1st1, line 231: Please add newlines between methods

1st1 (commented):
Looks good, let's continue fixing few of the remaining nits

src/client.ts
```diff
@@ -111,16 +123,88 @@ class StandaloneConnection implements Connection {
...
+          continue;
+        }
+        throw err;
+      }
+      // TODO(tailhook) sort out errors on commit, early network errors
+      // and some other errors could be retried
+      // NOTE: we can't retry on all the same errors as we don't know if
+      // commit is succeeded before the database have received it or after
+      // it have been done but network is dropped before we were able
+      // to receive a response
```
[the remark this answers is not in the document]
tailhook: "before we were were to receive a response". I.e. before client received the CommandComplete for the `COMMIT`.

src/client.ts
```diff
@@ -69,6 +74,12 @@ enum TransactionStatus {
   TRANS_UNKNOWN = 4, // cannot determine status
 }
 
+const DEFAULT_MAX_ITERATIONS = 3;
```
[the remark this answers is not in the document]
tailhook: I think we agreed no to do it this round.

src/client.ts
tailhook: Done

src/client.ts
tailhook: I've added whenever found. But honestly I don't understand why we have a prettifier if it doesn't handle such a basic thing for us.

src/ifaces.ts
tailhook: Done

1st1 (approved):

PR #85, opened 2021-01-20
add Set to Datatypes conversion table

fmoor:


docs/api/types.rst
```diff
@@ -12,6 +12,8 @@ The table below shows the correspondence between EdgeDB and JavaScript types.
 +----------------------+-----------------------------------------------------+
 | EdgeDB Type          |  JavaScript Type                                    |
 +======================+=====================================================+
+| ``Set``              | ``Set``                                             |
```
1st1, line 15: This isn't correct; the corresponding JS type is Array

fmoor: changed

Issue #73, opened 2020-09-25
starting index at 0 generates inconsistencies in dates


fimbault:
To reproduce : 
- alpha5
- using edgedb-js

I started from https://github.com/edgedb/edgedb-js/blob/master/docs/usage.rst

And changed the date of birth to `dob: new edgedb.LocalDate(1982, 10, 28)`

The console displays
Set(1) [
  Object [ name := 'Bob, dob := LocalDate [ 1982-11-28 ] ]
]

The documentation does say that the index starts at 0, but it's still quite confusing (october or november?). Starting at 1 would probably be best to avoid mistakes.

1st1:
Yes, we need to implement the stable part of the https://tc39.es/proposal-temporal/docs/datetime.html spec, and that includes dates and months starting from 1.

1st1:
This issue should have been resolved now that we implement the Temporal spec

PR #87, opened 2021-02-01
Update `Local*` datetime types to be compatible with Temporal spec

### LINKED ISSUE #70: Implement https://tc39.es/proposal-temporal

jaclarke:
Partly closes/fixes #70 and #73

Updates the `LocalDate`, `LocalTime` and `LocalDateTime` types to be minimal implementations of `PlainDate`, `PlainTime` and `PlainDateTime` (https://tc39.es/proposal-temporal/docs), with only the basic properties/parameters (ignoring `calendar` related things) and a few easy to implement properties in `LocalDate` and `LocalDateTime` (`.dayOfWeek`, etc.), no static methods (`.from()`, etc.), and only the `.toString()`, `.toJSON()` and `.valueOf()` methods.

A minimal `Temporal.Duration` type is also implemented, however it doesn't seem possible to easily make it compatible with the EdgeDB `duration` type without also implementing the `Temporal.Duration.total()` method, since `Temporal.Duration` has units like weeks, months and years, which are ambiguous to convert to microseconds without being relative to a specific date. This looks like the reason that those units were dropped from the EdgeDB duration type: (https://github.com/edgedb/edgedb/pull/947, https://github.com/edgedb/edgedb/issues/903).
I've left the current (microseconds only) `Duration` type as it is for now, and it seems the best thing to do would be to keep that as a custom edgedb type and add methods to convert to/from `Temporal.Duration` when that's officially finalised and implemented. @1st1 What do you think?

Todo:
- [x] Update docs
- [x] Update tests


1st1:
> I've left the current (microseconds only) Duration type as it is for now, and it seems the best thing to do would be to keep that as a custom edgedb type and add methods to convert to/from Temporal.Duration when that's officially finalised and implemented. @1st1 What do you think?

Hm, maybe we can actually implement all attributes from temporal.Duration but require that all ambiguous fields must stay 0. I.e. EdgeDB can only send Durations with microseconds in them that are always unambiguous. So we only need to make sure that the client bans them: prohibit passing Duration objects with years/months/weeks/days attributes set. That way we can stay largely compatible with the Temporal spec yet have our types and client behavior fully unambiguous.

cc @elprans @tailhook 

Temporal.Duration docs: https://tc39.es/proposal-temporal/docs/duration.html

elprans:
We have two duration types: `std::duration` and `cal::relativedelta`.  The latter can accept days as input.  It seems like `Temporal.Duration` designers intended for it to be used both as an absolute and a relative time delta, so I think @1st1 is right in that we can simply restrict which values of `Temporal.Duration` are acceptable as `std::duration` input.

tailhook:
I'm not sure why we have both `edgedb.Duration` and `TemporalDuration` in edgedb. The idea of these specs is that there will be a common type `Temporal` namespace in browser, no that it's interface a lot of types would support.

So we need conversion of `edgedb.Duration` to `Temporal.Duration` but don't need the actual type for latter (basically because `TemporalDuration` is `cal::relativedelta` rather than a duration). That conversion should assert that fields like `years` or `days` are zero.

And then it's clear that we don't want to return `Temporal.Duration` for `std::duration` from edgedb. But we can discuss whether we want implicit conversion when `Temporal.Duration` is used as an argument. Which given the general type juggling nature of JavaScript I think we should.

jaclarke:
> I'm not sure why we have both edgedb.Duration and TemporalDuration in edgedb. The idea of these specs is that there will be a common type Temporal namespace in browser, no that it's interface a lot of types would support.

As I understand it the idea is that these types will eventually be replaced by their `Temporal` equivalents when browsers natively support them. So we won't have both, but we want the current types to be compatible with the `Temporal` spec now, so it's not a breaking change when they are replaced.

haikyuu:
> when browsers natively support them.
In the JS ecosystem, we almost never write directly in Javascript. Instead we use babel plugins or pollyfills  to start using future proposals. e.g: many companies are using decorator proposal in production even though it's still experimental --> through angular for example.

1st1:
> I'm not sure why we have both edgedb.Duration and TemporalDuration in edgedb.

Yeah, we should have just one to avoid confusion.

> As I understand it the idea is that these types will eventually be replaced by their Temporal equivalents when browsers natively support them. So we won't have both, but we want the current types to be compatible with the Temporal spec now, so it's not a breaking change when they are replaced.

Yeah, all true. But why have two duration types? We should have just one (we can call it just `Duration`) which would be a subtype of the proposed Duration type in the temporal spec. Once the spec is approved and is implemented in nodeJS we just drop our own Duration type implementation and use the temporal one.

tailhook:
> > As I understand it the idea is that these types will eventually be replaced by their Temporal equivalents when browsers natively support them. So we won't have both, but we want the current types to be compatible with the Temporal spec now, so it's not a breaking change when they are replaced.
> 
> Yeah, all true. But why have two duration types? We should have just one (we can call it just `Duration`) which would be a subtype of the proposed Duration type in the temporal spec. Once the spec is approved and is implemented in nodeJS we just drop our own Duration type implementation and use the temporal one.

`Temporal.Duration` is a `cal::relativedelta` type not `std::duration`. We already commited to have different types for that in all languages and database itself. So I don't understand your point.

1st1:
> We already commited to have different types for that in all languages and database itself. So I don't understand your point.

Temporal.Duration, as designed in their spec, can represent both relative and absolute deltas (at least from what I see). Why don't we just use it then?

jaclarke:
I've removed the `LocalDate.weekOfYear` property for now, since it doesn't match the value the temporal polyfill gives but there's an issue in the polyfill's calculation of `weekOfYear`.

1st1 (approved):
Good stuff, James.

PR #91, opened 2021-02-19
Execute statement disabling transactions

tailhook:
This is based on #90  and #84 so look at latest commit only.


1st1:
I've rebased & force pushed this branch

1st1 (approved):

PR #93, opened 2021-02-22
Add Deno support

jaclarke:
Moves import of any node api's to `adaptor.node.ts`, and adds a script to generate a deno compatible version of the library.

The command `yarn compileForDeno` copies everything from `./src` to `./deno` and applies the following transformations:
- `tsc` only allows import paths without a file extension, since it doesn't rewrite the paths and relies on node's module resolution, deno however requires `.ts` file extensions (and no implicit index paths), so import paths are rewritten with `.ts` extensions (and implicit index paths are resolved)
- Imports of `adaptor.node.ts` are replaced with `adaptor.deno.ts`, which exports deno equivalents of the node api's we use
- Deno doesn't have the globals `process` and `Buffer` that exist in node, import of these globals from `globals.deno.ts` is injected where needed

Also deno doesn't have an equivalent of node's `url` module and this api was deprecated in node v11 (https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost), so it has been replaced with the WHATWG url api (https://developer.mozilla.org/en-US/docs/Web/API/URL), which introduces a few small changes in dsn parsing.

Todo:
- [x] Fix dsn parsing
- [x] Documentation
- [ ] Publishing
- [x] Tests?


haikyuu:
@jaclarke esbuild can be used instead of tsc:
- limitation 1 is gone
- possible to intersect import of node builtins and provide alternatives using [resolve callbacks](https://esbuild.github.io/plugins/#resolve-callbacks)
- 3 can be done as well with ease

Benefits:
- No need to write and maintain the script.
- Super fast (esbuild is amazing)
- Esbuild is used under the hood in deno itself

1st1:
@jaclarke 

> @jaclarke esbuild can be used instead of tsc:

This is actually pretty interesting. Can you take a look?

jaclarke:
Unfortunately I don't think we can replace tsc with esbuild, since esbuild doesn't do typechecking.

Though it looks like it could be good for the 'deno-ification' build (and probably less brittle than rewriting the imports with regexes).

1st1:
> Unfortunately I don't think we can replace tsc with esbuild, since esbuild doesn't do typechecking.

Yes, `tsc` must stay for sure.

> Though it looks like it could be good for the 'deno-ification' build (and probably less brittle than rewriting the imports with regexes).

Yep, this is exactly what I had in mind when I read @haikyuu's suggestion.  Specifically:

> possible to intersect import of node builtins and provide alternatives using resolve callbacks

sounds interesting.

jaclarke:
On a deeper look into esbuild, there doesn't appear to be a way to do what we want. It only allows generation of js, but we would need to output back to ts.

We aren't really 'building' the ts source code, more re-writing it (mostly the imports) to make it compatible with deno. The typescript package itself actually looks like it exposes functionality we could use here. It exports a compiler api, which allows to parse ts source files to an ast, and print that ast back to ts source code. (https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#creating-and-printing-a-typescript-ast)

1st1:
> The typescript package itself actually looks like it exposes functionality we could use here. 

I'm not against that and if you like using typescript compiler APIs more than regexps feel free to do it that way.

jaclarke:
@1st1 Is it still the plan to publish the deno version to a separate repository?

Also what testing do you think we need? None of the actual functionality is changed from the node version, so the jest tests should still cover that right? The bit that may need testing is the adaptors?

1st1:
> @1st1 Is it still the plan to publish the deno version to a separate repository?

I think so. Otherwise we'd have to litter this repo with deno builds. edgedb/edgedb-deno should work. We should port our CI release infra from edgedb-python to this repo and extend it with building the demo version, testing it, and if everything is green, pushing an update to edgedb/edgedb-deno automatically.

> Also what testing do you think we need? None of the actual functionality is changed from the node version, so the jest tests should still cover that right? The bit that may need testing is the adaptors?

I'd just run the jest tests. IMO there's little need to test the adapters (they should be tested just enough by the main test suite)

src/adaptor.node.ts
```diff
@@ -0,0 +1,44 @@
+import * as crypto from "crypto";
```
1st1, line 1: nit: Can we rename these files to `adapter`?

src/con_utils.ts
```diff
@@ -197,47 +198,35 @@ function parseConnectDsnAndArgs({
...
-        user = puser;
-      }
-      if (ppass && password == null) {
-        password = ppass;
-      }
+    if (parsed.username && user == null) {
+      user = parsed.username;
+    }
+    if (parsed.password && password == null) {
+      password = parsed.password;
```
1st1, line 205: Hm, this seems to be unrelated to this PR. I fear we might be accidentally reverting new-ish changes.

1st1: Ah, I see, nvm

test/deno/client.test.ts
```diff
@@ -0,0 +1,1404 @@
+/*!
```
1st1, line 1: Can we somehow avoid duplicating the tests? Perhaps generate test files with some yarn command and gitignore them? Maintaining two test sets will be extremely painful.

test/client.test.ts
```diff
@@ -861,10 +861,13 @@ test("fetch: tuple", async () => {
     expect(t0.length).toBe(2);
     expect(JSON.stringify(t0)).toBe('[1,"abc"]');
 
-    const insp = util.inspect(t0);
-    expect(
-      insp === "Tuple [ 1, 'abc' ]" || insp === "Tuple(2) [ 1, 'abc' ]"
-    ).toBeTruthy();
+    if (!isDeno) {
```
1st1, line 864: This should be factored into a function like `expectInspectToBe()` which would do nothing on Deno

1st1 (approved):

1st1:
Yay! Thanks so much, James! 🎉

PR #94, opened 2021-02-24
Decode scalar type names into codec

jaclarke:
Decodes the scalar type name annotation (https://www.edgedb.com/docs/internals/protocol/typedesc#scalar-type-name-annotation) into KNOWN_TYPES, so `ScalarCodec.getKnownTypeName()` returns the name of enum types.

https://github.com/edgedb/edgedb-js/blob/319857f06b651ea2dc047842ad968b2705990bdd/src/codecs/ifaces.ts#L88-L94

@1st1 `getKnownTypeName` returns the name for the `derivedFromTid` if it exists first. I'm not sure when derived types occur, so I'm not sure if this behaviour is correct? Shouldn't it check the typename of this type first, then for the derived type?

### DIFF
--- a/src/codecs/ifaces.ts
+++ b/src/codecs/ifaces.ts
@@ -74,7 +74,7 @@ export abstract class ScalarCodec extends Codec {
   }
 
   /** @internal */
-  setTypeName(typeName: string) {
+  setTypeName(typeName: string): void {
     this.typeName = typeName;
   }
 
--- a/src/codecs/ifaces.ts
+++ b/src/codecs/ifaces.ts
@@ -66,10 +66,16 @@ export abstract class Codec {
 
 export abstract class ScalarCodec extends Codec {
   private derivedFromTid: uuid | null = null;
+  private typeName: string | null = null;
 
   constructor(tid: uuid, derivedFromTid: uuid | null = null) {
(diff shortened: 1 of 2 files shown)


1st1:
> @1st1 getKnownTypeName returns the name for the derivedFromTid if it exists first. I'm not sure when derived types occur, so I'm not sure if this behaviour is correct? Shouldn't it check the typename of this type first, then for the derived type?

Derived types happen when you declare custom scalar types, like `scalar type MyInt extending int64`. When we receive a type like that, we say that it's essentially `int64`, so we derive our scalar codec from the int64 codec. Now that we sometimes also send scalar type names, we should store that name in the derived codec and return it from `getKnownTypeName`.

So basically

```
  getKnownTypeName(): string {
    if (this.typeName) { return this.typeName; }  // ++

    if (this.derivedFromTid) {
      return KNOWN_TYPES.get(this.derivedFromTid)!;
    }

    return KNOWN_TYPES.get(this.tid) || "anytype";
  }
```

and in `CodecsRegistry.buildCodec` we should set `codec.typeName` when/if we receive the scalar type name annotation.

1st1 (approved):

PR #95, opened 2021-02-25
Add option to implicitly fetch typenames

jaclarke:
Implicit typenames were removed by default in protocol version 0.9 (https://github.com/edgedb/edgedb/pull/2019). We need an api option to enable them (and implicit limits would be useful) for studio.

Todo:
- [x] Add `writeHeaders` method (https://github.com/edgedb/edgedb-python/blob/a7cc6c3811d41aeac0163de6580ab46d5e0b6b46/edgedb/protocol/protocol.pyx#L169)
- [x] Add api to set headers on `Prepare` and `OptimisticExecute` (https://www.edgedb.com/docs/internals/protocol/messages#prepare)

### DIFF
--- a/src/buffer.ts
+++ b/src/buffer.ts
@@ -20,6 +20,7 @@ import char, * as chars from "./chars";
 import {RingBuffer} from "./ring";
 import * as bi from "./bigint";
 import * as compat from "./compat";
+import {MessageHeaders, HeaderCodes} from "./ifaces";
 
 /* WriteBuffer over-allocation */
 const BUFFER_INC_SIZE: number = 4096;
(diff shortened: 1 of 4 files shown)


src/buffer.ts
```diff
@@ -206,6 +207,26 @@ export class WriteMessageBuffer {
...
+      ([_, value]) => value !== undefined
+    ) as Array<[keyof typeof HeaderCodes, string | Buffer]>;
+    this.buffer.writeUInt16(entries.length);
+    for (const [code, value] of entries) {
+      this.buffer.writeUInt16(HeaderCodes[code]);
+      if (Buffer.isBuffer(value)) {
+        this.buffer.writeUInt32(value.byteLength);
+        this.buffer.writeBuffer(value);
+      } else {
+        this.buffer.writeString(value);
```
1st1, line 224: I'd add an assertion `typeof value === 'string'` and raise an error if it's not a string and not a buffer.

src/client.ts
```diff
@@ -924,12 +926,13 @@ export class ConnectionImpl implements Executor {
...
     expectOne: boolean,
-    alwaysDescribe: boolean
+    alwaysDescribe: boolean,
+    options?: ParseOptions
   ): Promise<[number, ICodec, ICodec, Buffer | null, Buffer | null]> {
     const wb = new WriteMessageBuffer();
 
     wb.beginMessage(chars.$P)
-      .writeInt16(0) // no headers
+      .writeHeaders(options?.headers ?? {})
```
1st1, line 935:
```suggestion
      .writeHeaders(options?.headers ?? null)
```

and bail out on null in `writeHeaders` -- that will be a bit faster.

1st1 (approved):

PR #96, opened 2021-02-25
Increase jest timeout on 'connect: timeout' test

jaclarke:


1st1:
I think this test indicates that there's a race somewhere in connect(). I'm debugging.

jaclarke:
> I think this test indicates that there's a race somewhere in connect(). I'm debugging.

I haven't been able to reproduce it locally, so I was thinking maybe it's just the CI being slow, but it still fails even when just that test is run, so it does look like a race somewhere.

1st1:
> I haven't been able to reproduce it locally, so I was thinking maybe it's just the CI being slow, but it still fails even when just that test is run, so it does look like a race somewhere.

Yeah, see the last commit here: https://github.com/edgedb/edgedb-js/pull/97

PR #99, opened 2021-03-19
Transaction options part of RFC1004

tailhook:
This is the JavaScript counterpart of edgedb/edgedb-python#174

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -43,12 +43,12 @@ import {
   BorrowReason,
   IConnectionProxied,
   onConnectionClose,
-  TransactionOptions,
   ParseOptions,
   PrepareMessageHeaders,
 } from "./ifaces";
 import * as scram from "./scram";
-import {Options} from "./options";
+import {Options, RetryOptions, TransactionOptions} from "./options";
+import {PartialRetryRule} from "./options";
 
 import {
   parseConnectArguments,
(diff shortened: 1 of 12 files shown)


src/options.ts
```diff
@@ -0,0 +1,153 @@
+import * as errors from "./errors";
+
+export type BackoffFunction = (n: number) => number;
+
+export function default_backoff(attempt: number): number {
```
jaclarke, line 5:
```suggestion
export function defaultBackoff(attempt: number): number {
```

src/client.ts
```diff
@@ -229,94 +262,160 @@ class StandaloneConnection implements Connection {
...
+    config: NormalizedConnectConfig
+  ): Promise<S> {
+    const conn = new this(config);
+    await conn[INNER].reconnect();
+    return conn;
+  }
+}
+
+export class InnerConnection {
+  borrowed_for?: BorrowReason;
```
jaclarke, line 396:
```suggestion
  borrowedFor?: BorrowReason;
```

src/client.ts
```diff
@@ -88,11 +89,6 @@ function default_backoff(attempt: number): number {
   return 2 ** attempt * 100 + Math.random() * 100;
```
jaclarke, line 89: I think this is redundant now? Looks like it's been moved to options.ts

src/options.ts
tailhook: Fixed, thanks!

src/client.ts
tailhook: Fixed. Thanks! (always mess up with camel-case convention :( )

src/client.ts
tailhook: Yes. Removed.

tailhook:
Updated:
1. No partial options objects
2. `Conflict` condition instead of `Deadlock` / `SerializationError`
3. More tests

1st1 (approved):

tailhook:
Waits for #100

1st1:
Paul, can you rebase this PR and merge it?

PR #102, opened 2021-05-05
Implement reading project config

tailhook:
Details:
1. Uses current directory for project dir
2. Does not search of the ancestor directories (unlike CLI tool)
3. Now crashes if `edgedb.toml` not found and no parameters specified

Also includes implementation of `EDGEDB_INSTANCE` env var support

### DIFF
--- a/src/con_utils.ts
+++ b/src/con_utils.ts
@@ -16,7 +16,9 @@
  * limitations under the License.
  */
 
-import {path, homeDir} from "./adapter.node";
+import {path, homeDir, createHash} from "./adapter.node";
+import {readFileUtf8Sync, existsSync, realpathSync} from "./adapter.node";
+import * as errors from "./errors";
 import {readCredentialsFile} from "./credentials";
 
 const EDGEDB_PORT = 5656;
@@ -92,6 +94,19 @@ export function parseConnectArguments(
   };
 }
 
+function stashPath(projectDir: string): string {
+  let projectPath = realpathSync(projectDir);
+  if (process.platform === "win32" && !projectPath.startsWith("\\\\")) {
+    projectPath = "\\\\?\\" + projectPath;
+  }
+  const hasher = createHash("sha1");
+  hasher.update(projectPath);
+  const hash = hasher.digest("hex");
+  const baseName = path.basename(projectPath);
+  const dirName = baseName + "-" + hash;
+  return path.join(homeDir(), ".edgedb", "projects", dirName);
+}
+
 function parseConnectDsnAndArgs({
   dsn,
   host,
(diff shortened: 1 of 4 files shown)


src/con_utils.ts
```diff
@@ -124,6 +137,41 @@ function parseConnectDsnAndArgs({
...
+      process.env.EDGEDB_PORT
+    )
+  ) {
+    if (process.env.EDGEDB_INSTANCE) {
+      dsn = process.env.EDGEDB_INSTANCE;
+    } else {
+      const dir = process.cwd();
+      if (!fs.existsSync(path.join(dir, "edgedb.toml"))) {
+        throw new errors.ClientConnectionError(
+          "no `edgedb.toml` found and no connection options specified"
```
1st1, line 156:
```suggestion
          "no `edgedb.toml` found and no connection options specified" +
            " neither via arguments to connect API nor via environment" +
            " variables EDGEDB_HOST/EDGEDB_PORT or EDGEDB_INSTANCE"
```

src/con_utils.ts
```diff
@@ -93,6 +96,16 @@ export function parseConnectArguments(
   };
 }
 
+function stashPath(projectDir: string): string {
+  const projectPath = fs.realpathSync(projectDir);
```
jaclarke, line 100:
```suggestion
  let projectPath = fs.realpathSync(projectDir);
  if (process.platform === "win32" && !projectPath.startsWith("\\\\")) {
    projectPath = "\\\\?\\" + projectPath;
  }
```

1st1 (approved):
Would be nice if we could come up with a test for this. Other than that looks good!

src/con_utils.ts
1st1: @jaclarke Should this be implemented in https://github.com/edgedb/edgedb-python/pull/180 too?

jaclarke: Yes, it looks like `os.path.realpath` in python behaves the same as `fs.realpath` in node.js

1st1: Please suggest the change in that PR too, so that Paul does not miss this

src/con_utils.ts
fmoor: I think `neither` should be `either` and `nor` changed to `or`

1st1 (approved):

1st1:
@tailhook Could you please rebase this PR? And let's merge

1st1:
Seems like we need to add a `sha1` implementation to `node.adapter` and `deno.adapter`. You need to import from `node.adapter` in your code (`deno.adapter` is substituted automatically during the build process)

src/con_utils.ts
```diff
@@ -16,7 +16,10 @@
  * limitations under the License.
  */
 
-import {path, homeDir} from "./adapter.node";
+import {path, homeDir, createHash} from "./adapter.node";
+import * as url from "url";
```
1st1, line 20: This import isn't needed?

tailhook: Yes, merge artifact.

src/con_utils.ts
```diff
@@ -92,6 +94,19 @@ export function parseConnectArguments(
   };
 }
 
+function stashPath(projectDir: string): string {
+  let projectPath = fs.realpathSync(projectDir);
```
1st1, line 98: Need to create a shim for this one... Deno has `Deno.realpathSync`

src/con_utils.ts
```diff
@@ -123,6 +138,43 @@ function parseConnectDsnAndArgs({
...
+      host ||
+      port ||
+      process.env.EDGEDB_HOST ||
+      process.env.EDGEDB_PORT
+    )
+  ) {
+    if (process.env.EDGEDB_INSTANCE) {
+      dsn = process.env.EDGEDB_INSTANCE;
+    } else {
+      const dir = process.cwd();
```
1st1, line 154: And we probably need a shim for this one, deno has `Deno.cwd()`

jaclarke: This one's ok, the `process` and `Buffer` globals are injected automatically

1st1: Oh, nice

PR #105, opened 2021-05-06
Fix Deno to run without `--unstable`. Add a test that that's possible.

1st1:
While there, fix Deno (and our test suite) to connect to a TCP socket,
as Deno doesn't work well with Unix sockets on macOS.

### DIFF
--- a/test/deno.test.ts
+++ b/test/deno.test.ts
@@ -17,11 +17,17 @@
  */
 
 import {execFile} from "child_process";
+import * as fs from "fs";
 
 test("run deno test", async () => {
   jest.setTimeout(60_000);
 
-  return new Promise<void>((resolve, reject) => {
+  if (!fs.existsSync("test/deno")) {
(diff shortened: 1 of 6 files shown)


test/deno.test.ts
```diff
@@ -47,3 +53,30 @@ test("run deno test", async () => {
...
+      ["eval", 'import * as edgedb from "./edgedb-deno/mod.ts"'],
+      {
+        env: process.env,
+      },
+      (error, stdout, stderr) => {
+        if (error) {
+          console.error(stderr);
+          reject(error);
+        }
+        console.log(stdout);
```
jaclarke, line 77: Don't really need this `console.log`, nothing should get logged

jaclarke (approved):

test/deno.test.ts
1st1: fixed

PR #108, opened 2021-05-13
add windows to github CI

fmoor:
Run tests on windows as well as linux.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,10 +15,18 @@ jobs:
     strategy:
       matrix:
         node-version: ["12", "13", "14", "15"]
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-latest]
         edgedb-version: ["stable", "nightly"]
 
     steps:
+      - name: Setup WSL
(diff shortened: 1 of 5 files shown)


1st1:
This is amazing, Frederick

1st1 (approved):

1st1:
@jaclarke please also take a look

jaclarke:
For some reason the deno tests are being skipped on windows

.github/workflows/tests.yml
```diff
@@ -15,10 +15,18 @@ jobs:
     strategy:
       matrix:
         node-version: ["12", "13", "14", "15"]
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-latest]
```
elprans, line 18: Any reason to not also include macOS here?

fmoor: No idea.

fmoor: @1st1 or @jaclarke is there a reason that macOS should not be added?

1st1: I think it should be

fmoor: added

PR #115, opened 2021-07-02
Switch test server to --auto-shutdown-after

fmoor:


fmoor:
Looks like `--auto-shutdown-after` has not made it to a stable release yet so those CI runs fail. Should I try to work around that somehow or wait until we release to merge this? @elprans

elprans:
I think we can wait until beta 3 pops up.

fmoor:
This functionality was introduced in https://github.com/edgedb/edgedb-js/pull/117

PR #117, opened 2021-07-09
Add TLS support

fmoor:
Deno ALPN support is tracked at denoland/deno#11479

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -45,6 +45,7 @@ import {
   onConnectionClose,
   ParseOptions,
   PrepareMessageHeaders,
+  ProtocolVersion,
 } from "./ifaces";
 import * as scram from "./scram";
 import {Options, RetryOptions, TransactionOptions} from "./options";
@@ -59,38 +60,8 @@ import {
 import {Transaction as LegacyTransaction} from "./legacy_transaction";
 import {Transaction, START_TRANSACTION_IMPL} from "./transaction";
 
-class ProtocolVersion {
-  readonly major: number;
-  readonly minor: number;
-
-  constructor(major: number, minor: number) {
-    this.major = major;
-    this.minor = minor;
-  }
-
-  greaterThan(other: ProtocolVersion): boolean {
-    if (this.major > other.major) {
-      return true;
-    }
-
-    if (this.major < other.major) {
-      return false;
-    }
-
-    return this.minor > other.minor;
-  }
-
-  greaterThanOrEqual(other: ProtocolVersion): boolean {
-    if (this.major === other.major && this.minor === other.minor) {
-      return true;
-    }
-
-    return this.greaterThan(other);
-  }
-}
-
-const PROTO_VER = new ProtocolVersion(0, 11);
-const PROTO_VER_MIN = new ProtocolVersion(0, 9);
+const PROTO_VER: [number, number] = [0, 11];
+const PROTO_VER_MIN: [number, number] = [0, 9];
 
 enum AuthenticationStatuses {
   AUTH_OK = 0,
@@ -118,6 +89,32 @@ const OLD_ERROR_CODES = new Map([
 ]);
 
 const DEFAULT_MAX_ITERATIONS = 3;
+
+export function versionGreaterThan(
+  left: [number, number],
+  right: [number, number]
+): boolean {
+  if (left[0] > right[0]) {
+    return true;
+  }
+
+  if (left[0] < right[0]) {
+    return false;
+  }
+
+  return left[1] > right[1];
+}
+
+export function versionGreaterThanOrEqual(
+  left: [number, number],
+  right: [number, number]
+): boolean {
+  if (left[0] === right[0] && left[1] === right[1]) {
+    return true;
+  }
+
+  return versionGreaterThan(left, right);
+}
 
 export default function connect(
   dsn?: string | ConnectConfig | null,
@@ -510,7 +507,7 @@ export class ConnectionImpl {
 
   private opInProgress: boolean = false;
 
-  protected protocolVersion: ProtocolVersion = PROTO_VER;
+  protected protocolVersion: [number, number] = PROTO_VER;
 
   /** @internal */
   protected constructor(sock: net.Socket, config: NormalizedConnectConfig) {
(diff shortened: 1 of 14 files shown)


src/client.ts
```diff
@@ -17,6 +17,7 @@
  */
 
 import {net, hrTime} from "./adapter.node";
+import * as tls from "tls";
```
jaclarke, line 20: An export for `tls` should be added to `adapter.node` and then imported from there. Currently it looks like Deno doesn't support ALPN, so I think the only thing we can do is export some mocked tls api from `adapter.deno` that falls back to the plaintext protocol?

fmoor: I'm under the impression that plane text is not an option going forward... @1st1 what do you think about this?

fmoor: changed

src/con_utils.ts
```diff
@@ -304,6 +338,20 @@ function parseConnectDsnAndArgs({
         delete parsedQ.password;
       }
 
+      if ("tls_cert_file" in parsedQ) {
```
1st1, line 341: Hm, using `in` is an antipattern in JS. @fmoor while here, can you add a separate commit making `parsedQ` a `Map` and using `.has()` instead of `in`?

fmoor: PR https://github.com/edgedb/edgedb-js/pull/119

fmoor: changed

src/client.ts
```diff
@@ -903,6 +943,19 @@ export class ConnectionImpl {
 
         case chars.$Z: {
           this._parseSyncMessage();
+
+          if (
+            !(this.sock instanceof tls.TLSSocket) &&
+            // @ts-ignore
+            typeof Deno === "undefined" &&
```
fmoor, line 950: I'm not sure this is the best way to special case deno :thinking:

src/con_utils.ts
```diff
@@ -422,12 +478,52 @@ function parseConnectDsnAndArgs({
...
+    }
+
+    tlsOptions = {ALPNProtocols: ["edgedb-binary"]};
+
+    if (tlsCAData.length !== 0) {
+      if (tlsVerifyHostname == null) {
+        tlsVerifyHostname = false;
+      }
+
+      // this option only needs to be set for self signed certificates
```
fantix, line 494: I think `tlsCAData` can be either self-signed certificates or future cloud CA certificate?

src/credentials.ts
```diff
@@ -65,5 +67,21 @@ export function validateCredentials(data: any): Credentials {
...
+    if (typeof certData !== "string") {
+      throw Error("`tls_cert_data` must be string");
+    }
+    result.tlsCAData = certData;
+  }
+
+  const verifyHostname = data.tls_verifyhostname;
+  if (verifyHostname != null) {
+    if (typeof verifyHostname !== "boolean") {
+      throw Error("`tls_verifyhostname` must be boolean");
```
fantix, line 81: I didn't test but looks like this should be `tls_verify_hostname`?

fantix (approved):

src/client.ts
fantix: This looks reasonable to me - any comments @jaclarke or @1st1 pls?

jaclarke: I think it's ok, we already use that check in other places: https://github.com/edgedb/edgedb-js/blob/master/src/compat.ts#L44

src/client.ts
```diff
@@ -60,9 +59,38 @@ import {
 import {Transaction as LegacyTransaction} from "./legacy_transaction";
 import {Transaction, START_TRANSACTION_IMPL} from "./transaction";
 
-const PROTO_VER_MAJOR = 0;
-const PROTO_VER_MINOR_MIN = 9;
-const PROTO_VER_MINOR = 11;
+class ProtocolVersion {
```
jaclarke, line 62: The protocol version is also used in 'src/codecs/registry', I think that may also need updating to use this class.

fmoor: changed

src/client.ts
```diff
@@ -60,9 +59,38 @@ import {
 import {Transaction as LegacyTransaction} from "./legacy_transaction";
 import {Transaction, START_TRANSACTION_IMPL} from "./transaction";
 
-const PROTO_VER_MAJOR = 0;
-const PROTO_VER_MINOR_MIN = 9;
-const PROTO_VER_MINOR = 11;
+export class ProtocolVersion {
```
1st1, line 62: IMO this is an overkill and should be done with two functions (`versionGreaterThan` and `versionGreaterThanOrEqual`) typed to accept a tuple of two numbers.

1st1 (commented):
Overall looks great, but the version comparison code needs to be simplified.

src/client.ts
fmoor: what about using arrays of numbers? `[0, 11] >= [0, 10]` this _seems_ to work the same way comparing tuples in python works.

fmoor: changed

jaclarke: > what about using arrays of numbers? `[0, 11] >= [0, 10]` this _seems_ to work the same way comparing tuples in python works.

I think this just converts them to strings, then compares them. So `[0, 11] >= [0, 9] === false`

src/client.ts
```diff
@@ -903,6 +910,19 @@ export class ConnectionImpl {
 
         case chars.$Z: {
           this._parseSyncMessage();
+
+          if (
+            !(this.sock instanceof tls.TLSSocket) &&
+            // @ts-ignore
+            typeof Deno === "undefined" &&
+            this.protocolVersion >= [0, 11]
```
1st1, line 918: This is, unfortunately, an incorrect way -- JS would cast the arrays to strings and compare strings. Most of the times that will work until it doesn't one day. Instead, do what you did before -- have two functions to do the version comparisons, but on two-element tuples, not objects.

src/client.ts
fmoor: ok :+1:

src/client.ts
fmoor: changed

1st1 (approved):
lgtm, please squash before merging

src/client.ts
fmoor: removed array comparison

PR #118, opened 2021-07-17
Use generics to able specify return types of queries

kafein:


1st1 (approved):

1st1:
@jaclarke I think this is a great idea. Thoughts?

haikyuu:
Great idea. It's going to be a first step in the direction of automatic extraction of types from queries. And that's super useful to have and it's one of the main selling points of prisma.

In the future, I hope we could use something like this:

```ts
import getUsersQuery from "./getUsers.edgeql`

const users = await query(getUsersQuery);
// user has User[] type
```
That would require:

- Fixing query introspection https://github.com/edgedb/edgedb/issues/2098
- Translating an introspection result to typescript types
- Making a babel plugin like [this graphql one](https://github.com/detrohutt/babel-plugin-import-graphql). Should be straightforward as Graphql and EdgeQL are both query languages, so that code should be a good starting point

1st1:
> Great idea. It's going to be a first step in the direction of automatic extraction of types from queries.

Yeah, we're also working on a full-featured query builder for TS.

jaclarke:
Maybe there should be a generic for the query parameters type as well?

1st1:
Closing this in favor of https://github.com/edgedb/edgedb-js/pull/135. Thanks a lot @kafein for bringing this up!

PR #130, opened 2021-08-06
Infer cardinality in `.filter`

colinhacks:
Basic type inference for `.filter`. Works for queries with this structure:

`select X filter X.property = <value>` 

where `X` is any non-terminal path and `property` has a unique constraint.

TODO: support queries like: 

`select X.property filter X.property = <value>`


src/syntax/select.ts
```diff
@@ -91,9 +97,32 @@ export type $expr_Select<
...
-} & SelectMethods<Set>;
+} & SelectMethods<Set, Expr>;
 
-interface SelectMethods<Self extends TypeSet> {
+// Base is PathNode &
+// Filter is equality &
+// Filter.args[0] is PathLeaf &
+// Filter.args[0] is unique &
+// Filter.args[0].parent.__element__ === Base.__element__
+export type inferCardinality<
```
jaclarke, line 107: I think this needs to also check that the cardinality of the second arg is `Cardinality.AtMostOne`, since for example `select User filter .username = {'userA', 'userB'}` can return multiple users.

colinhacks: Good call, fixed. 

I also loosened the restriction that `Base` be a `PathNode`. It now works for any Object Expression.

PR #132, opened 2021-08-09
Deprecate access to single connections

fmoor:
The client APIs are moving toward a pool first approach. This is the first step in that direction.

### DIFF
--- a/docs/api/connection.rst
+++ b/docs/api/connection.rst
@@ -527,6 +527,8 @@ Pool
 
     .. js:method:: acquire()
 
+        **Deprecated**: use the query methods on ``Pool`` instead.
+
         Acquire a connection proxy, which provides access to an open database
         connection. The proxy must be released to return the connection to the
         pool.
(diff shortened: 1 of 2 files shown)


docs/api/types.rst
```diff
@@ -97,20 +97,20 @@ EdgeDB ``array``  maps onto the JavaScript ``Array``.
     const edgedb = require("edgedb");
 
     async function main() {
-      const conn = await edgedb.connect({
+      const pool = await edgedb.createPool({
```
elprans, line 100: I think we should continue to call this `connect()`.  Technically, this is a backwards-compatible change as all the methods behave the same. @1st1?

1st1: +1

src/client.ts
```diff
@@ -120,6 +120,11 @@ export default function connect(
   dsn?: string | ConnectConfig | null,
   options?: ConnectConfig | null
 ): Promise<Connection> {
+  // tslint:disable-next-line: no-console
+  console.warn(
+    "The `connect()` function is deprecated and is scheduled to be " +
+      "removed. Use `createPool()` instead."
+  );
```
1st1, line 127: Yeah, to be clear we'll actually deprecate the `createPool` function in the future, leaving only `connect()`. @elprans

1st1: 1. should not
2. should not

this PR *should only* deprecate `Pool.acquire` and `Pool.release`. Same in edgedb-python. Everything else is a bit too complex to do right now.

docs/api/types.rst
fmoor: https://github.com/edgedb/edgedb-js/pull/132#discussion_r686373221

src/client.ts
fmoor: changed. 

I also deprecated `Pool.run()` because it seemed very similar to `acquire()`.

1st1 (approved):

PR #135, opened 2021-08-10
Make query and querySingle generic

colinhacks:
This is a more complete implementation of #118 

`query` and `querySingle` now have the following type signature for across connections, pools, and transactions:

```ts
query<T extends unknown[] = unknown[]>(
    query: string,
    args?: QueryArgs
  ): Promise<T>;
querySingle<T extends unknown>(query: string, args?: QueryArgs): Promise<T>;
```

This also means we aren't reflecting `Set` in the static return type of `query`, as discussed on the team call. `Set` is still used under the hood.

As implemented, this is a breaking change, as the return types now default to `unknown[]` and `unknown`, respectively. This means the returned value cannot be used unless 1) the type signature is provided to the method or 2) the returned data is explicitly cast to an appropriate type first. This is a best practice for TS libraries.


1st1:
The CI is red, please play with `yarn lint` and `yarn format`

docs/usage.rst
```diff
@@ -42,15 +42,12 @@ run queries.
         );
 
         // Select User objects.
-        let userSet = await conn.query(
+        const userSet = await conn.query(
```
1st1, line 45: Perhaps define an interface

```ts
interface User {
  name: string;
  dob: ??
}```

and use `await conn.query<User>()`?

File: docs/api/types.rst [hunk 2]
```diff
@@ -77,13 +77,21 @@ Sets
...
+    ``edgedb.Set``. This class represents a set of values returned by a query.
+    If the query contained an explicit ``ORDER BY`` clause, the values will be
+    ordered, otherwise no specific ordering is guaranteed.
 
     This type also allows to differentiate between a set of values and an
     explicit array.
 
+    .. code-block:: js
+
+        const result = await conn.query(`SELECT {0, 1, 2};`);
```
1st1, line 90:
```suggestion
        const result = await conn.query<number>(`SELECT {0, 1, 2};`);
```

1st1 (commented):
I'd go through the `*.rst` files and fix examples to actually include result types.

1st1 (approved):

colinhacks:
> I'd go through the `*.rst` files and fix examples to actually include result types.

I was leaving the other docs as plain JavaScript (so all the code works as-is in a Node.js environment). I restricted the TypeScript section to a dedicated page. If we we're switching all docs to be TypeScript-first, there are other changes to be made, e.g. changing from `require` to `import` syntax. But it's more conventional to not assume TypeScript.

PR #138, opened 2021-08-12
Retry retryableTransaction on network errors

fmoor:
- network errors during transaction blocks are retried
- the connection is reconnected on network errors
- start transaction errors are considered for retrying

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -63,6 +63,8 @@ import {
 import {Transaction as LegacyTransaction} from "./legacy_transaction";
 import {Transaction, START_TRANSACTION_IMPL} from "./transaction";
 
+const IS_NETWORK_ERROR = Symbol("network error");
+
 const PROTO_VER: ProtocolVersion = [0, 12];
 const PROTO_VER_MIN: ProtocolVersion = [0, 9];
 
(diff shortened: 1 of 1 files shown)


src/client.ts
```diff
@@ -237,8 +237,9 @@ export class StandaloneConnection implements Connection {
           }
         }
         if (
-          err instanceof errors.EdgeDBError &&
-          err.hasTag(errors.SHOULD_RETRY)
+          (err instanceof errors.EdgeDBError &&
+            err.hasTag(errors.SHOULD_RETRY)) ||
+          !(err instanceof errors.EdgeDBError)
```
fmoor, line 242: I'm not sure how to differentiate network errors from other kinds of errors. JS is not my forte. This implementation is rather simplistic. Reading https://nodejs.org/api/errors.html it doesn't seem to offer a solution beyond keeping a curated list of `err.code`s that correspond to network errors. Are there better ways to do this?

1st1: Network errors will come from this callback: https://github.com/edgedb/edgedb-js/blob/b6c705d1ec8c4274f89d838411506083dce0d938/src/client.ts#L531

You can tag the exception object with a custom attribute (like `Symbol("network-error")`). and check that attribute when you're deciding to retry or not.

1st1:
I'm going to wait a bit before merging this. I'd like to move some code around and maybe I'll figure out a better way of solving this problem.

fmoor:
> maybe I'll figure out a better way of solving this problem

Sounds good to me :+1:

PR #147, opened 2021-08-19
Show a clear warning when we're waiting for the DB to become available

1st1:
The `waitUntilAvailable` set to 30 seconds by default makes the
`connect()` call to retry connecting multiple times, until either the DB
is available ot the timeout is hit. This can create an impression that
"edgedb.connect()" is hung indefinitely. Fix this by prninting a warning
to stderr explaining what is happening.

The warning also suggests to try `connect()` without arguments if the
program is run within a directory created with `edgedb project init` yet
`connect()` was called with explicit arguments.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -435,17 +435,39 @@ export class InnerConnection {
     return this._isClosed;
   }
 
+  logConnectionError(...args: any): void {
+    if (!this.config.logging) {
+      return;
+    }
+    if (
+      this.config.inProject &&
+      !this.config.fromProject &&
+      !this.config.fromEnv
+    ) {
+      args.push(
+        `\n\n\n` +
+          `Hint: it looks like the program is running from a ` +
+          `directory initialized with "edgedb project init". ` +
+          `Consider calling "edgedb.connect()" without arguments.` +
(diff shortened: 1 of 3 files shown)


src/client.ts
```diff
@@ -433,17 +433,39 @@ export class InnerConnection {
...
   }
 
+  logConnectionError(...args: any): void {
+    if (!this.config.logging) {
+      return;
+    }
+    if (
+      this.config.inProject &&
+      !this.config.fromProject &&
+      !this.config.fromEnv
```
tailhook, line 443: This looks like doesn't ensure that there actually were any parameters to `.connect()`. And I think it may be confusing to see `Consider calling "edgedb.connect()" without arguments.` when there aren't any. I would desperately try to find actual arguments created in some layer of abstractions, while in fact it maybe instance just isn't running.

tailhook (approved):

src/client.ts
1st1: Let's fix this in later iterations. I agree this can be confusing, but I'd like @jaclarke to finalize his connection testing/rework before I add code there.

PR #148, opened 2021-08-19
Querybuilder: Implement `withParams`

jaclarke:


qb/test/params.test.ts
```diff
@@ -0,0 +1,36 @@
...
+        nums: e.array_unpack(params.numArr),
+        x: e.if_else(e.str("true"), params.optBool, e.str("false")),
+      })
+  );
+
+  expect(query.toEdgeQL()).toEqual(`SELECT {
+  str := (<std::str>$str),
+  nums := (std::array_unpack((<array<std::int64>>$numArr))),
+  x := (("true" IF <OPTIONAL std::bool>$optBool ELSE "false"))
+}`);
```
1st1, line 23: I think that instead of testing the output of `toEdgeQL()` we should execute the queries against the server and check the result. The rendering of queries might change, but what the compile to semantically must be the same across all versions.

1st1: IOW it's more cumbersome but better in the long term, I believe.

PR #149, opened 2021-08-19
Functional tests

colinhacks:
Implement function tests for SELECT
Remove casting on strings, float64, and int64 when unnecessary
Fix toEdgeQL bugs relating to nested shapes

### DIFF
--- a/qb/test/select.test.ts
+++ b/qb/test/select.test.ts
@@ -361,20 +361,21 @@ test("limit 2", async () => {
 });
 
 test("shapes", async () => {
-  const query = e
-    .select(e.Hero, {
+  const query = e.select(
+    e
+      .select(e.Hero)
+      .filter(e.eq(e.Hero.name, e.str("Iron Man")))
+      .$assertSingle(),
+    {
       id: true,
       name: true,
       secret_identity: true,
       villains: {id: true},
-    })
-    .filter(e.eq(e.Hero.name, e.str("Iron Man")));
+    }
+  );
 
-  const result = await query.query(pool);
-
-  expect(result?.id).toEqual(data.iron_man.id);
-  expect(result?.name).toEqual(data.iron_man.name);
-  expect(result?.secret_identity).toEqual(data.iron_man.secret_identity);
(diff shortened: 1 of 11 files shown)


qb/test/select.test.ts
```diff
@@ -317,3 +332,101 @@ test("nonchainable offset/limit", () => {
...
+
+test("shapes", async () => {
+  const query = e
+    .select(e.Hero, {
+      id: true,
+      name: true,
+      secret_identity: true,
+      villains: {id: true},
+    })
+    .filter(e.eq(e.Hero.name, e.str("Iron Man")));
```
1st1, line 371: Do we have assertSingle implemented?

I'm asking because if we do, then that would make it a better place, but also you could then

`expect(result).toMatchObject(data.iron_man)`

1st1 (commented):
Looks amazing :heart_eyes:

qb/test/with.test.ts
```diff
@@ -5,15 +5,15 @@ test("simple repeated expression", () => {
...
+  __withVar_0 := ({ 1, <std::int32>2, <std::int16>3 })
 SELECT ((__withVar_0 + __withVar_0))`);
 });
 
 test("simple expression with alias", () => {
   const numbers = e.set(e.int64(1), e.int32(2), e.int16(3));
 
   expect(e.select(e.plus(numbers, e.alias(numbers))).toEdgeQL()).toEqual(`WITH
-  __withVar_0 := ({ <std::int64>1, <std::int32>2, <std::int16>3 }),
+  __withVar_0 := ({ 1, <std::int32>2, <std::int16>3 }),
```
1st1, line 16: Yeah, this is the reason why I think we should mainly have functional tests

qb/test/select.test.ts
colinhacks: Good call, updated.

qb/test/with.test.ts
colinhacks: Definitely. More fun to write too. 🤙🤙

PR #164, opened 2021-09-07
Support protocol version 0.12

fmoor:
Addjust codec types to support protocol version 0.12.


fmoor:
looks like node 16.9.0 has a bug https://github.com/nodejs/node/issues/40030 :disappointed:

1st1 (approved):

PR #168, opened 2021-09-09
Change `connect()` to return a `Pool`, and deprecate `createPool()`

jaclarke:


colinhacks:
Do we want some fallback function for creating a StandaloneConnection? Maybe `createStandaloneConnection`? Or at least export the `StandaloneConnection` so users can manually call `StandaloneConnection.connect`? @1st1 

Or is the point to disallow standalone connections?

1st1:
A stand-alone connection would be just a pool of min/max size set to 1. I think there's a way to create a raw (unmanaged) single connection object, need to check that.

PR #178, opened 2021-10-14
Update code gen to output .js+.d.ts and add edgedb-generate CLI

jaclarke:


colinhacks:
I made some small changes:

1. Made the CLI flow a little smoother and tweaked some of the copy. 
2. Get rid of `.hgignore` support. We should optimize the DX of the most common workflows as much as possible. Mercurial has 2% market share currently, and it's mostly big companies. 
3. Fixed the `copySyntaxToDist.sh` file for macOS. For some reason macOS doesn't support `cp -u`: https://superuser.com/questions/515373/cp-u-is-illegal-on-mac-what-are-the-alternatives
4. Update the `getTypes` query — it was breaking on a fresh nightly due to the link multiplicity changes in RC1

PR #199, opened 2021-11-16
Implement ALPN support for the Deno adapter.

1st1:
Depends on https://github.com/edgedb/deno/commit/a95fef98b8c73894b985428e12f439c2730a5098

### DIFF
--- a/.github/workflows/deno-release.yml
+++ b/.github/workflows/deno-release.yml
@@ -26,7 +26,7 @@ jobs:
 
       - uses: denoland/setup-deno@v1
         with:
-          deno-version: v1.16.x
+          deno-version: v1.17.x
 
       - name: Install dev deps
         run: |
--- a/.github/workflows/deno-release.yml
+++ b/.github/workflows/deno-release.yml
@@ -26,7 +26,7 @@ jobs:
 
       - uses: denoland/setup-deno@v1
         with:
-          deno-version: v1.14.x
+          deno-version: v1.16.x
 
       - name: Install dev deps
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,7 +65,7 @@ jobs:
         uses: denoland/setup-deno@v1
         if: steps.release.outputs.version == 0
         with:
-          deno-version: v1.14.x
+          deno-version: v1.16.x
 
       - name: Install dev deps
         if: steps.release.outputs.version == 0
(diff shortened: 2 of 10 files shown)


1st1:
My test file (`t.ts`):

```typescript
import {createClient} from "./edgedb-deno/mod.ts";

async function main() {
  let conn = createClient({
    dsn: "_localdev", tlsSecurity: "insecure",
    //tlsCAFile: '/home/yury/.local/share/edgedb/_localdev/edbtlscert.pem'
  });
  console.log('=1');
  await conn.ensureConnected();
  console.log('=2');

  console.log(await conn.querySingle('select 32323'));
  console.log('=3');
}

main()
```

I was running it with 

```shell
yarn compileForDeno && \
   ../../deno/target/debug/deno run --allow-all --unstable \
       --unsafely-ignore-certificate-errors t.ts
```

Assuming that Deno was cloned into the `../../deno` directory from the https://github.com/edgedb/deno/tree/alpn branch.

To build Deno locally I followed https://deno.land/manual@v1.16.1/contributing/building_from_source. It's basically just cloning the repo with all submodules and running `cargo build`.

1st1:
@jaclarke James, please take a look at this. Specifically check if we need any more APIs for a complete edgedb-deno client. The `adapter.deno.ts / tls.connect()` function needs some work to pass all connection options. Then try testing without `--unsafely-ignore-certificate-errors`.

Lastly, by default Deno tries to connect to `::1` at least on my machine, but `edgedb-server` only listens on `127.0.0.1` (and doesn't listen on IPv6). Passing `127.0.0.1` to deno explicitly doesn't work for some reason (a limitation of `rustls`? I didn't spend much time debugging that). Assuming that I'll merge https://github.com/edgedb/edgedb/pull/3187 tonight, run `edb server -I "::1"` to allow Deno to be able to connect using TLS.

jaclarke:
> Specifically check if we need any more APIs for a complete edgedb-deno client.

The client will mostly work without these, but to fully match the behaviour of edgedb-js it looks like we will need a few more:
- To support `tlsSecurity` values `insecure` and `no_host_verification` we need equivalents of the `rejectUnauthorized` and `checkServerIdentity` options
- `ref`/`unref` for allowing deno to exit while idle connections are open

jaclarke:
> by default Deno tries to connect to ::1 at least on my machine

For me I'm assuming `localhost` still resolves to `127.0.0.1`, since it works without `edb server -I "::1"`, so I guess this depends on the user's system.

1st1:
> we need equivalents of the rejectUnauthorized 

Looks like that won't happen: https://github.com/denoland/deno/issues/1371

1st1:
I've filed a PR to add ALPN support: https://github.com/denoland/deno/pull/12786

jaclarke (approved):

PR #204, opened 2021-11-23
Clean up querybuilder test flow + fix import path generation

jaclarke:
@colinhacks I've dropped the separate pretest setup and tests, and just moved everything into a single flow in `qb/test/testRunner.ts` (run with command `yarn test:ci`), so it doesn't need to start up a test server and do the migrations twice. For local testing with a project instance it's still possible to just run the tests separately with `yarn test`.


colinhacks (approved):
Devastatingly superior to my approach 🙌

PR #225, opened 2021-12-13
Query builder 

colinhacks:
To-do:

- [x] Implement `e.update` with closure syntax and a `set` key
```ts
const query = e.update(e.Hero, (hero) => ({
  filter: e.bool(true),
  // offset: 10,
  // limit: 2
  set: {
    name: e.op('The Amazing ', '++', hero.name),
  },
}));
```
- [x] Support raw data in `e.insert` and `e.update` 
- [x] Add support for raw data in other expressions/funcops
- [x] Expand `e.withParams` to support collection/nested types: https://edgedb.slack.com/archives/C0256T82EQZ/p1638995639059500
- [x] Eliminate `namedTuple`: implement as overload of `e.tuple`
- [x] Rename `withParams` to `e.params`: https://edgedb.slack.com/archives/C0256T82EQZ/p1639425872098900
- [x] Re-implement `pathify` with Proxy (if performant)
- [x] Implement direct indexing and dot-notation using Proxy. Using `e.index` and `e.destructure` is a nightmare. https://edgedb.slack.com/archives/C0256T82EQZ/p1639426405101100
   - [x] `e.array(e.str('hi'))[0]`
   - [x] `e.str_split(e.str('qwerty'), e.str(''))['2:4']`
   - [x] `e.tuple(e.str('hi'))[0]`
   - [x] `e.json({hello:"world"}).hello`
   - [x] `e.tuple({hello: e.str('hi')}).hello`
- [x] Expand `e.json` literals to accept any JS data structure. https://edgedb.slack.com/archives/C0256T82EQZ/p1639427201107300
- [x] Add a `.run` method to expressions (runtime only, not TS) that instructs you to wrap the expression in `e.select`. Better error experience for JS users who try to execute expressions with `.run` 
- [x] Move `qb` into `tests`
- [x] Re-export reflection types (e.g. TypeSet, etc) from generated code. Alias `TypeSet` to `set` so it can be used like `e.set`: `const filters: e.set<e.bool, e.Cardinality.Many>[] = [];` . Or without an explicit cardinality: `const filters: e.set<e.bool>[] = [];` — Cardinality should defaults to Many.
- [x] Implement `e.op`, remove functional operators
- [x] Implement a single `number` type instead of separate int16/int32/int64/float32/float64 types. Remove implicit casting overloads.
- [x] Implement clause methods on `expr_Select`
   - [x] `.filter`: make sure re-assignment is possible when adding filters that don't change result cardinality
   - [x] `.order`
   - [x] `.limit/.offset`
- [x] Add support for string literals in number & decimal literal constructors: e.decimal('12348712349871234')
- [x] Add some looser overloads to `e.set`. 
```ts
const qbFilters: edgedb.$.TypeSet<typeof e['bool']>[] = [];
e.set(...qbFilters);
// A spread argument must either have a tuple type or be passed to a rest parameter.ts

e.set(...([] as any[]));
// A spread argument must either have a tuple type or be passed to a rest parameter.ts
```
- [x] Add `e.delete`
- [x] Remove modifier methods
- [x] Make expressions runnable
- [x] Hide `$type` expressions
- [x] Make e.number internal
- [x] `.$is` -> `.is` and `$assertSingle` -> `assert_single`
- [x] Support raw JS data in `e.select`: `e.select(5)`
- [x] Rename `edgedb-generate` to `edgeql-js`
- [x] Try to remove need for `downlevelIteration`: requires moving away from `for/of` syntax for sets/iterators. Low priority.

Keep literal constructors like `e.str('asdf')`, `e.int16(14)` for now.


docs/driver.rst
```diff
@@ -0,0 +1,311 @@
...
+  * RFC1004_
+  * :js:meth:`Client.transaction\<T\>`
+
+  .. _RFC1004: https://github.com/edgedb/rfcs/blob/master/text/1004-transactions-api.rst
+
+
+Next up
+-------
+
+If you're using plain JavaScript, the driver API will likely meet your needs. If you're a TypeScript user and want autocompletion and type inference, head over to the :ref:`Query Builder docs <edgedb-js-qb>`.
```
haikyuu, line 311: I think the query builder is useful even though typescript is not used. VSCode can give autocomplete suggestions even with js and using functions is better than edgeql query strings for decent usage

docs/expressions.rst
```diff
@@ -0,0 +1,44 @@
...
+default import called ``e``.
+
+**#2** The result of ``e.select("Hello world!")`` is an *expression*. In fact, everything you produce with the query builder is an expression. (Don't worry if the API doesn't look familiar, it'll be covered later.)
+
+.. code-block:: typescript
+
+  e.str("Hello world!");
+  e.set(1, 2, 3);
+  e.count(e.Movie);
+  e.select(e.Movie, hero => ({ id: true, title: true }));
```
haikyuu, line 34:
```suggestion
  e.select(e.Movie, movie => ({ id: true, title: true }));
```
right?

docs/index.rst
```diff
@@ -1,42 +1,149 @@
...
+    uppercase_title := str_upper(.title)
+  }
+  filter .title = "Iron Man"
+
+.. code-block:: typescript
+
+  e.select(e.Movie, movie => ({
+    id: true,
+    title: true,
+    uppercase_title := e.str_upper(movie.title),
```
haikyuu, line 138:
```suggestion
    uppercase_title : e.str_upper(movie.title),
```

haikyuu (commented):
Really excited about this. Looks amazing already. 1 week left 🤞

docs/driver.rst
colinhacks: Updated to reflect this

docs/index.rst
colinhacks: 🤙

1st1 (approved):

Issue #242, opened 2022-02-13
Typing inference wrong on inserts


busheezy:
It seems that an insert is showing as a type of many. This makes it impossible to chain with a conflict else.

The error:
``the query has cardinality MANY which does not match the expected cardinality ONE``

https://github.com/busheezy/edgedb-upsert-type-inference-error/blob/main/src/index.ts#L9
I have made an example repo. You can run ``npm run rebuild:db`` to create everything.

https://share.busheezy.dev/wrlepR.png

Thanks to @elprans for the quick reply on discord.

Thanks for the cool new toys!

colinhacks:
Good catch, we'll push some fixes by the end of today!

colinhacks:
Fixed in 0.19.2 👍

PR #245, opened 2022-02-14
Various querybuilder fixes

jaclarke:
- Fix bug @colinhacks found during launch livestream where expressions get wrongly double counted when `select`ing a non basic path object and `select` contains nested shapes, causing expressions to be hoisted out of their parent scope.
- Fix bug in `toEdgeQL` causing this issue: #242 + fix cardinality/element type in other `insert unless conflict` cases
- Fix broken code gen when `default` module is empty (reported here: https://discord.com/channels/841451783728529451/849374705210490900/942414500567588954)

### DIFF
--- a/qb/dbschema/default.esdl
+++ b/qb/dbschema/default.esdl
@@ -24,7 +24,9 @@ module default {
   type Movie {
     property genre -> Genre;
     property rating -> float64;
-    required property title -> str;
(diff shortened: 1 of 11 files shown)


colinhacks (approved):

colinhacks:
This has landed in 0.19.2 👍

PR #246, opened 2022-02-15
Fix getTsType generation on CJS

Sikarii:
Currently there is a broken declaration for `getTsType` when generating for CJS via `--target cjs`, this PR should fix that.
```ts
declaretype getTsType<T extends $.BaseType> = T extends $.ScalarType
  ? T extends _std.$decimal | _std.$json | _std.$uuid
    ? never
    : T["__tstype__"]
  : never;
```


colinhacks (approved):

colinhacks:
This has landed in 0.19.2 👍

PR #247, opened 2022-02-15
Fix getSharedParentScalar declaration

Sikarii:
Currently the declaration for `getSharedParentScalar` has a function body included, this PR should get rid of it in the declaration file.
```ts
declare function getSharedParentScalar<A extends $.ScalarType, B extends $.ScalarType>(a: A, b: B): A | B
 as any
 as any
```


colinhacks (approved):

colinhacks:
This has landed in 0.19.2 👍

PR #261, opened 2022-02-16
Keep shape in nested selects

colinhacks:
Shapes now bubble up `select`s. The `__shape__` type is inherited whereas the runtime `__shape__` is set to null. In `toEdgeQL` null shapes are not rendered.


jaclarke:
The generated edgeql still has `{id}` appended to it due to this line: https://github.com/edgedb/edgedb-js/pull/261/files#diff-74a639e138172ff7ab214ce01b38ac27b24345aca0cbabb377bba4ebf6b0854eL643.

In the general case though I don't think this approach will work, since `toEdgeQL` can pull out any expression into a `with` block, and when that happens the shape isn't always propagated to the final output. So I don't think we can throw away `__shape__` at runtime, and instead will need to make sure the shape gets rendered in the final `select` to be correct.

Example where this breaks:
```ts
const heroNames = e.select(
  e.select(e.Hero, () => ({
    name: true,
  }))
);
const query = e.select({
  heroNames,
  heroNames2: heroNames,
});

console.log(query.toEdgeQL());
// WITH
//   __withVar_0 := (
//     SELECT (WITH
//       __scope_1_Hero := (DETACHED default::Hero)
//     SELECT (__scope_1_Hero) {
//       name
//     }) 
//   )
// SELECT {
//   multi heroNames := (__withVar_0 ),
//   multi heroNames2 := (__withVar_0 )
// }

console.log(await query.run(client));
// Object [ heroNames := Set(3) [
//   Object [ id := '1c143f768f3311ec910e5b737c31d944' ],
//   Object [ id := '1c14e70a8f3311ec910e933674cfe0db' ],
//   Object [ id := '1c156c7a8f3311ec910e433cf65239ae' ]
// ], heroNames2 := Set(3) [
//   Object [ id := '1c143f768f3311ec910e5b737c31d944' ],
//   Object [ id := '1c14e70a8f3311ec910e933674cfe0db' ],
//   Object [ id := '1c156c7a8f3311ec910e433cf65239ae' ]
// ] ]
```

Issue #259, opened 2022-02-16
Enum Not Assignable


bavellone:
Using the latest v0.19.5 with the following minimal schema:
```
module default {
  scalar type Env extending enum<prod, staging>;

  type App {
    required property env -> Env;
  }
}
```

I'm getting this typing error on the `env` property when trying to insert:
```
const app = e.insert(e.App, {
  env: e.Env('prod')
})

(property) env: {
    __element__: never;
    __cardinality__: Cardinality.One;
}
Type '$expr_Literal<$Env>' is not assignable to type '{ __element__: never; __cardinality__: Cardinality.One; }'.
  Types of property '__element__' are incompatible.
    Type '$Env' is not assignable to type 'never'.
```

colinhacks:
Oops, good catch. Fix pending: https://github.com/edgedb/edgedb-js/pull/262

colinhacks:
@bavellone the fix has landed in 0.19.6.

bavellone:
Thanks for the super quick turnaround @colinhacks! Updated to 0.19.6 but I still seem to be getting the same error after regenerating types.

PR #262, opened 2022-02-16
Support raw data for multi properties. Fix enum assignment expressions.

### LINKED ISSUE #260: Array type errors
Using the latest v0.19.5 with the following minimal schema:
```
module default {
  type App {
    required property nicknames -> array<str>;
  }
}
```

According to the [docs](https://www.edgedb.com/docs/clients/01_js/literals#arrays), arrays can accept plain JS data but I'm getting type errors when passing a variable:
```js

// Works
const app = e.insert(e.App, {
  nicknames: e.array(['one', 'two'])
})

// Throwing errors on `nicknames` and `e.array(names)`
const names = ['one', 'two']
const app = e.insert(e.App, {
  nicknames: e.array(names)
})
```
First error
```
(property) nicknames: {
    __element__: ArrayType<$str, "array<std::str>">;
    __cardinality__: Cardinality.One;
}
Property '__cardinality__' is missing in type 'ArrayType<CastableNonArrayType, `array<${string}>`>' but required in type '{ __element__: ArrayType<$str, "array<std::str>">; __cardinality__: Cardinality.One; }'
```
Second error
```
No overload matches this call.
  Overload 1 of 2, '(element: CastableNonArrayType): ArrayType<CastableNonArrayType, `array<${string}>`>', gave the following error.
    Argument of type 'string[]' is not assignable to parameter of type 'CastableNonArrayType'.
      Type 'string[]' is missing the following properties from type 'NamedTupleType<NamedTupleShape>': __name__, __kind__, __shape__
  Overload 2 of 2, '(arg: [string, ...string[]]): $expr_Array<ArrayType<ScalarType<"std::str", string, true, string>, "array<std::str>">, never>', gave the following error.
    Argument of type 'string[]' is not assignable to parameter of type '[string, ...string[]]'.
      Source provides no match for required element at position 0 in target.
```

colinhacks:
- Improve assignment expressions for `multi` properties
- Fix enum assignment expressions

Closes #260 and  #259 

Remaining issue: you can't pass `string[]` into `e.array()`. This is minor for now, we can merge without it since using `e.array` for literals is rarely necessary.

### DIFF
--- a/qb/test/insert.test.ts
+++ b/qb/test/insert.test.ts
@@ -186,7 +186,28 @@ test("insert type enforcement", async () => {
 
 test("optional sequence fields", async () => {
   const query = e.insert(e.Bag, {
-    stringsMulti: "asdf",
+    stringsMulti: ["asdf"],
   });
   await query.run(client);
+});
+
+test("complex raw data in inserts", async () => {
+  const strings = ["aaa", "bbb"];
+  const query = e.insert(e.Bag, {
+    stringsArr: strings,
+    stringsMulti: strings as ["aaa", "bbb"],
+    stringMultiArr: [strings],
+  });
+  const final = e.select(query, () => ({
+    id: true,
+    stringsMulti: true,
+    stringsArr: true,
+    stringMultiArr: true,
+  }));
+  const result = await final.run(client);
+  expect(result).toMatchObject({
+    stringsMulti: strings,
+    stringsArr: strings,
+    stringMultiArr: [strings],
+  });
 });
--- a/qb/test/insert.test.ts
+++ b/qb/test/insert.test.ts
@@ -186,7 +186,7 @@ test("insert type enforcement", async () => {
 
 test("optional sequence fields", async () => {
   const query = e.insert(e.Bag, {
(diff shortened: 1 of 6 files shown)


qb/test/update.test.ts
```diff
@@ -35,7 +35,7 @@ test("update", async () => {
   e.update(e.Bag, () => ({
     set: {
       stringsMulti: {
-        "+=": "new string",
+        "+=": ["new string"] as ["new string"],
```
jaclarke, line 38: I think we should allow non array `string`s here as well since it's modifying the existing multi set.

qb/test/insert.test.ts
```diff
@@ -186,7 +186,28 @@ test("insert type enforcement", async () => {
 
 test("optional sequence fields", async () => {
   const query = e.insert(e.Bag, {
-    stringsMulti: "asdf",
+    stringsMulti: ["asdf"],
```
jaclarke, line 189: And we should probably allow setting non array values for multi properties anyway. The `multi` just makes the upper cardinality `Many` instead of `One`, so singletons are still ok.

PR #273, opened 2022-02-24
Fix implicit any on toEdgeQL.ts

Sikarii:
Fixes an issue where the generated client would not play nicely with `noImplicitAny: false`.
This probably is also a Typescript compiler bug, I would expect `noImplicitAny: false` to do quite the opposite (not care about the implicit anys).
However I saw that the same variables were typed in other places of the code, so this change is probably justified.

This can be reproduced with the following repository: https://gitlab.com/tdolsen/edgedb-testing, thanks to tdolsen on the Discord.


jaclarke (approved):

jaclarke:
Thanks for the PR @Sikarii

Issue #270, opened 2022-02-23
Question: "Cannot find namespace 'e' when using Typescript types from edgeql-js


johanbrook:
Hi! Sorry for a "question issue", but I just can't get past this myself.

- I've setup edgeql-js according to the [docs](https://www.edgedb.com/docs/clients/01_js/generation).
- I've run `npx edgeql-js` to generate all code.
- I can query for results with `e.select().run(client)` and so on.

But when I try to do use a generated TS type from the package, I run into this TS error:

```
error TS2503: Cannot find namespace 'e'
```

Code:

```ts
import e from './edgeql-js'; // auto-generated

// can't use my own types from the schema:
//   error TS2503: Cannot find namespace 'e'
let c: e.Contact = {};

// not even built-types work:
//   error TS2503: Cannot find namespace 'e'
let s: e.str = e.str('hey');
```

I'm following the docs in ["Objects and paths"](https://www.edgedb.com/docs/clients/01_js/objects#object-types), which says:

> All object types in your schema are reflected into the query builder, properly namespaced by module.
> ...
> For convenience, the contents of the `default` module are also available at the top-level of `e`.

Are the docs talking about the *Typescript type* here? Or is it a reference to the EdgeQL schema type?

**Thoughts**

- Am I somehow misunderstanding the use of the types from edgeql-js? Am I not supposed to be using them like this? I was of the impression of that I could generate TS types and use in my app code. I understand the types are inferred when returning from a query function, but there are places where I'd like to get hold of the type too, for use in React component props, for instance.
- Am I importing `e` incorrectly?

**Versions**

```
Typescript: 4.5.5
EdgeDB (npm): 0.19.8
```

***

Many thanks!

johanbrook:
Thanks for the reply!

Yea, I've toyed around with various kinds of `import type` and `typeof e.User` but to no avail 😄 I've read the source of the generated code, and `$ContactλShape` seemed very ... non-public and scary to use in app code.

My angle is from GraphQL, where you'd generate TS types for two things:

1. The schema itself (objects, enums, scalars, ...)
2. Queries and mutations — their result types and input variables.

Is it so that 1) and 2) are inferred by TS with thanks to edgeql-js? Meaning, no "concrete" types are generated, but everything is meant to be inferred when taking care of query results.

Example:

```ts
export const findContacts = async () ⇒
   e.select(e.Contact, () ⇒ {
      name: true,
      // email: true,
   }).run(client);
```

[...]

PR #281, opened 2022-03-02
Generate plain types

jaclarke:
@colinhacks A few questions:
- Should we export all the interfaces? So you can just import the types want directly from `"./dbschema/edgeql-js/types"`, without having to create your own aliases like: `type Movie = types['default']['Movie']`. The potential downside is if the type name isn't ascii we generate a name with the id in it, which won't be stable between different database instances.
- Do we make the types partial types or do we leave that to the user? (It's easy enough to switch between them with `Partial<>` or `Required<>`, so it's a question of which is the more useful default)


src/reflection/util/genutil.ts
```diff
@@ -77,29 +77,38 @@ for (const [scalarType, {type, literalKind}] of Object.entries(
...
-  types: introspect.Types
+  types: introspect.Types,
+  opts = {unionEnums: false, edgedbDatatypePrefix: "_."}
 ): CodeFragment[] {
   switch (type.kind) {
     case "scalar": {
       if (type.enum_values && type.enum_values.length) {
+        if (opts.unionEnums) {
+          return [`(${type.enum_values.map(quote).join(" | ")})`];
+        }
```
johanbrook, line 91: Why union and not the native Typescript `enum` type?

johanbrook:
Whoa, what a quick turnaround 😍😍😍 Super great work!

*** 

If I may input (I did the original [issue](https://github.com/edgedb/edgedb-js/issues/270)):

## Import/export

This feels _much_ better:

```ts
import { Movie } from './dbschema/edgeql-js/types';
```
than this
```ts
import { types } from './dbschema/edgeql-js';

type Movie = types["default"]["Movie"];
```

The latter would make a local `types.ts` file almost required, just to re-export types for ergonomic reasons. Not a huge biggie though, but it'd be one of these occasions where I'd think "Oh, I wonder why they did that".

## Partial/Required

Coming from GraphQL types generation, all libs respect the nullable property on each schema field. So the generated EdgeDB schema types would be optional/null depending on their `required` property in the schema. Unless I'm reading your question incorrectly.

```edgedb
# Schema
type Movie {
   required property title → str;
   property subtitle → str;
}
```

generated:

```ts
// types.ts
export interface Movie {
   title: string;
   subtitle?: string;
}
```

With regards to using optional (`?`) vs. `null`, I've seen the former being nicer to work with in TS codebases. Example: `prop?: X` instead of `prop: X | null`. Former is good, latter less good.

src/reflection/util/genutil.ts
jaclarke: To mirror the type that the querybuilder would infer if you had `select`ed the property. I think there's also a concern with having a TS enum type, that because it exists as both a static type and a runtime object, it would be possible to confuse it with the other similar object that's generated for the enum (which represents the edgeql expression of the enum for use in building the queries).

jaclarke:
> So the generated EdgeDB schema types would be optional/null depending on their required property in the schema. Unless I'm reading your question incorrectly.

That's not quite what I meant. The types mirror what would be inferred if you queried them, so if it's `multi` it will be `T[]` and if it's `optional` it will be `T | null`.

The question I was wondering about was: should the default be as though you `select`ed all the fields in a type or should all the fields have the TS `?` (ie. `prop?: T`), so the type represents the fields that might be returned from querying a type. Though thinking about it more, I think the first option makes more sense.

> With regards to using optional (?) vs. null

The querybuilder has to use `null` here, since the underlying edgedb-js client itself returns empty sets as `null`.

haikyuu:
> should all the fields have the TS ?
From my experience with typescript, it's not a pleasant experience. That's because you end up "skipping" the safety with `movie.title!` because you don't want to always have checks `if(movie.title) ...`

Then it makes more sense to make it as though all the properties were selected and let the user use `Pick` to choose the ones he wants to keep at a query basis or in a helper file.

These are type definitions and it makes sense to make them as close as possible to the schema

johanbrook:
@haikyuu How is it "not pleasant" if it reflects the truth (the schema)? If something is nullable/optional in the schema, the code should check that before operating on a field. The chain operator `?` for `obj.field?.anotherField` is handy.

src/reflection/util/genutil.ts
colinhacks: I think this should probably make these plain `enum`s - I don't think we have to worry deeply about the potential confusion surrounding multiple runtime exports, since the import will be a totally different file. Also, TS treats enums as assignable to equivalent string unions: 

```ts
enum Genre {
  Action = "Action",
  Horror = "Horror",
  Romcom = "Romcom"
}

const gen: "Action" | "Horror" | "Romcom" = Genre.Horror;
```

colinhacks:
This is great, good idea generating into a separate file. 

I think we should add `?` to all optional properties. This can easily be removed with `Required`:

```ts
type Movie = {
  title?: string | null
};
type ReqMovie = Required<Movie>;
// { title: string | null; }
```

jaclarke:
What's the benefit of adding `?` to just optional props? They already have `| null`.

colinhacks:
People will be using (modified versions of) these types for function arguments, etc, where optionality is best represented with `?`. Plus it's tricky to add the `?` to nullable fields, but it's easy to remove it (with `Requiured<T>`)

colinhacks:
> Should we export all the interfaces? 

Yeah we should do this, good call.

colinhacks (approved):

quinnvaughn:
I don't actually love that this generates all the links if they're required, because it causes a ton of problems. I use Pothos for my graphql server and you can pass in types to its schema, but take this example:

I have a user that has a role, which is a required link. Now with the types this generates, it expects a role with EVERY single User that is being returned, which obviously I don't actually need to do, the graphql resolvers will handle that. I would prefer at the very least an option that just has the primitives for a type without any of the links. 

Yes, I can just pass in the Omitted types, but I'd rather not have to do that every time, that's extremely brittle.

haikyuu:
@quinnvaughn There are ways to get types from a graqphQL query. You can also omit a field in a graphql query, so using these generated types isn't the right thing for your use case IMO.

quinnvaughn:
It's not a query. You can pass in types to Pothos to let the autocomplete know what 'User' refers to instead of passing in e.users.User every time. 

```
type Objects = {
  DateIdea: DateIdea
  User: CurrentUser
  DateStop: DateStop
  Location: Location
  Address: Address
  City: City
  State: State
  Country: Country
  PlannedDate: PlannedDate
  Role: users.Role
  Waitlist: Waitlist
  Whitelist: Whitelist
}

export type TypesWithDefaults = PothosSchemaTypes.ExtendDefaultTypes<{
  Context: Context
  Objects: Objects
  Scalars: {
    Date: {
      Input: Date
      Output: Date
    }
    DateTime: {
      Input: Date
      Output: Date
    }
  }
}>
```

This sort of thing.

quinnvaughn:
so when I return to the return type of a resolver as 'User' instead of pass in CurrentUser (which I made to omit role), it knows what I'm referring to. It saves me an import.

quinnvaughn:
Turns out the `Props` helper does this. Nevermind!

PR #316, opened 2022-03-25
Add `.runJSON` to query builder expressions

colinhacks:


src/syntax/query.ts
```diff
@@ -32,3 +32,24 @@ export async function $queryFunc(this: any, cxn: edgedb.Executor, args: any) {
...
+export async function $queryFuncJSON(
+  this: any,
+  cxn: edgedb.Executor,
+  args: any
+) {
+  const expr = runnableExpressionKinds.has(this.__kind__)
+    ? this
+    : wrappedExprCache.get(this) ??
+      wrappedExprCache.set(this, select(this)).get(this);
+  const _args = jsonifyComplexParams(expr, args);
```
1st1, line 45: Hm, what's this btw? Why are we "jsonifying" arguments?

colinhacks: We allow more complex parameter types and implicitly convert them to JSON on both ends: https://www.edgedb.com/docs/clients/01_js/parameters#complex-types This makes it possible to accept more complex data structures in `query.run(client, parameters)` to support an end-to-end typesafe workflow where you take a JSON payload, validate it with a schema library, and pass it in the `params` object, and type safety is enforced at every step along the way

PR #319, opened 2022-03-28
Binary protocol channeling over HTTP

1st1:
Requires the server to run with https://github.com/edgedb/edgedb/pull/3682.

------

@jaclarke James, I hope this would unblock you to implement querying from the admin panel.

Some notes:

1. I've only tested this running from the nodejs environment (not browser). The key limitation here is that node (pre 17.5) doesn't have native `fetch()`, so I'm using `node-fetch` to emulate it for testing purposes. When you run this in the browser, you might need to update `FetchSocket.write()` to properly cast emulated `Buffer` to something that browser's `fetch()` can understand (most likely `ArrayBuffer`).

2. Every HTTP request essentially creates a new (temporary) binary connection on the server. Which means that the typical `[parse] [execute]` message sequence doesn't work! 

   * When you execute `[parse]` it creates an anonymous statement.
   * The subsequent `[execute]` message doesn't see that anonymous statement, because the connection is already different.
   * The way to fix this is to use the `[parse] [optimisticExecute]` sequence, essentially not relying on anonymous statements at all.
   * I've fixed (in a hacky way) the `.query()` method to do just that and it works. However, iirc the studio uses the `rawExecute` method that would have to be updated to also use optimistic execute.

3. Aside from the above it should work in the browser. I'll be completing some ToDos over the next couple of days, but you should be able to start playing with this.

------

ToDo

- [x] Proper state management in `FetchSocket`
- [ ] Think about the API -- do we want `createClient` to automatically detect the browser environment and try to use `fetch()`? (this decision can wait at least until we think about enabling this feature beyond admin panel)

------

@jaclarke to test things from nodejs:

```javascript
edgedb = require('./dist/index.node.js');

async function main() {
  con = edgedb.createFetchClient({host: 'localhost', port: 5656});
  console.log(await con.query('select 1'));
  console.log(await con.query('select {1, 2, 3}'));
}

main()
```

(don't forget to run `yarn` to install `node-fetch@2`)

-------

James, lastly, please commit and push directly to this branch when you make this work in the browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "@types/node": "14",
     "get-stdin": "^7.0.0",
     "jest": "^26.6.3",
+    "node-fetch": "2.6.7",
     "nodemon": "^2.0.13",
     "prettier": "^2.3.2",
     "proposal-temporal": "^0.7.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "edgedb",
-  "version": "0.20.3",
+  "version": "0.20.4",
   "description": "The official Node.js client library for EdgeDB",
   "homepage": "https://edgedb.com/docs",
   "author": "EdgeDB <info@edgedb.com>",
--- a/src/baseConn.ts
+++ b/src/baseConn.ts
@@ -1,0 +1,1012 @@
+/*!
+ * This source file is part of the EdgeDB open source project.
+ *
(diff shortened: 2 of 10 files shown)


1st1:
@jaclarke this should be ready for merge. Please review, push any changes you see necessary, and merge/release.

jaclarke (approved):

PR #332, opened 2022-04-20
Select star

colinhacks:
Usage

```ts
const query = e.select(e.Movie, () => ({
  ...e.Movie["*"],
}));
```


1st1 (approved):
@colinhacks Please add docs before you release!

PR #334, opened 2022-04-25
When selecting a required multi link, wrap expr in `assert_exists` guard

### LINKED ISSUE #333: Bug: QueryError: possibly an empty set returned by an expression for a computed link 'rects' declared as 'required'
I'm using  the `edgedb/edgedb` image (docker compose).

Steps to Reproduce: Try to select a required multi link.

Schema:

Relevant part of schema (I think):

```
	abstract type Post {
		multi link votes := .<post[is Vote];
		multi link answers := .<post[is Answer];
		required property shortId -> str {
			constraint exclusive;
		}
		required link user -> User;
		required property createdAt -> datetime;
		required property title -> str;
		required property content -> str;
		required property score -> int16;
	}

	type PDFPost extending Post {
		required link document -> PDF;

		required property excerpt -> str;
		required link excerptRect -> PDFRect;
		required property page -> int16;

		required property anchorIdx -> int16;
		required property focusIdx -> int16;
		required property anchorOffset -> int16;
		required property focusOffset -> int16;

		# maybe I should inline this as JSON?
		# (reasoning ... less space, more performant?)
		required multi link rects -> PDFRect;
	}

	type PDFRect {
		required property x -> int16;
		required property y -> int16;
		required property width -> int16;
		required property height -> int16;
	}
```

I have 1 post in the database. When I run `select PDFPost { rects };`, I do see a bunch of rectangles being returned, as expected.

image

However, when I run the following query:

```
const Selectors = {
    RECT: {
        x: true,
        y: true,
        width: true,
        height: true,
    },
} as const;

    const query = e.select(e.PDFPost, post => ({
        title: true,
        content: true,
        score: true,
        createdAt: true,
        rects: Selectors.RECT,
        excerptRect: Selectors.RECT,
        answers: {
            content: true,
            createdAt: true,
            user: {
                shortId: true,
                displayName: true,
                image: true,
            },
        },
        user: {
            shortId: true,
            displayName: true,
            image: true,
        },
        filter: e.op(post.shortId, "=", id),
        document: {
            shortId: true,
            club: {
                name: true,
                public: true,
                shortId: true,
            },
        },
    }));
```

I get the error mentioned in the title of this post. Which doesn't make much sense, since I know from the edgedb CLI that there are rectangles.

jaclarke:
Fixes #333

### DIFF
--- a/qb/dbschema/default.esdl
+++ b/qb/dbschema/default.esdl
@@ -47,7 +47,7 @@ module default {
 
   type User {
     required property username -> str;
-    required link favourite_movie -> Movie;
+    required multi link favourite_movies -> Movie;
   }
 
   type MovieShape {
--- a/qb/dbschema/migrations/00007.edgeql
+++ b/qb/dbschema/migrations/00007.edgeql
@@ -1,0 +1,14 @@
+CREATE MIGRATION m163zjk7tidrkupp7i75krcnyyi26aboiyk6zq7mdaoijdkydijyha
+    ONTO m13m5xkrznccqnvocth4njrqz2rnipe5fhlrermlo56nvyp6kgrtuq
+{
+  ALTER TYPE default::User {
+      ALTER LINK favourite_movie {
+          RENAME TO favourite_movies;
+      };
+  };
+  ALTER TYPE default::User {
+      ALTER LINK favourite_movies {
+          SET MULTI;
+      };
+  };
+};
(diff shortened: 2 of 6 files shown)


vedantroy:
Will this be pushed, i'm working around this by dropping the `required` but that is bad for my data integrity

colinhacks:
Landed in `v0.20.9` @vedantroy

PR #335, opened 2022-04-25
Correctly quote backlink keys in shapes

### LINKED ISSUE #329: Selecting backlinks results in an invalid syntax (EdgeQLSyntaxError: Unexpected '<')
- EdgeDB Version: `0.20.7`
- EdgeDB CLI Version: `1.1.1+932d5c6`

When trying to select backlinks on a model, the generated query looks something like this:
```
SELECT __some_scope {
  id,
  name,
  multi <some_model[is some_other_model] := (
...
```

EdgeDB complains that `<` there is unexpected, and I see no way to alias it in the query builder to something else. Replacing  `multi <some_model[is some_other_model] :=` with something like `some_other_models :=` makes the query work.

jaclarke:
Fixes #329

### DIFF
--- a/qb/test/select.test.ts
+++ b/qb/test/select.test.ts
@@ -603,6 +603,30 @@ test("backlinks", async () => {
   expect(
     [data.the_avengers.title, data.civil_war.title].includes(result1[0].title)
   ).toEqual(true);
+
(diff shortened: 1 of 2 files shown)


colinhacks:
Landed in `v0.20.9`

PR #353, opened 2022-05-18
Custom codecs update

jaclarke:
- Update to latest temporal polyfill for tests
- Add `from` method to `Duration` type according to temporal spec: https://tc39.es/proposal-temporal/#sec-temporal.duration.from
- Re-organise custom codecs:
  - drop `bigint` -> `string` codec: not needed since bigint's are supported pretty much everywhere and we can polyfill when they aren't
  - change `int64` -> `string` to `int64` -> `bigint`: so we can add encode codec
  - drop `local_datetime` codec and change `datetime` codec to decode into a `LocalDatetime`: datetime types now fit into `Date` type (https://github.com/edgedb/edgedb/pull/2252), except `datetime` needs `local_datetime` codec because `Date` doesn't have microsecond precision
- Add encode codec for decimal strings

### DIFF
--- a/src/datatypes/datetime.ts
+++ b/src/datatypes/datetime.ts
@@ -27,8 +27,6 @@ import {
   isLeapYear,
 } from "./dateutil";
 
-export type {EdgeDBDateTime} from "../compat";
-
 export const DATE_PRIVATE = Symbol.for("edgedb.datetime");
 
 export class LocalTime {
--- a/src/datatypes/datetime.ts
+++ b/src/datatypes/datetime.ts
@@ -29,6 +29,14 @@ import {
 
 export const DATE_PRIVATE = Symbol.for("edgedb.datetime");
 
+function toNumber(val: any): number {
+  const n = Number(val);
+  if (Number.isNaN(n)) {
+    return 0;
+  }
+  return n;
+}
+
 export class LocalTime {
   private readonly _hour: number;
   private readonly _minute: number;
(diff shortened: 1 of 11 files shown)


colinhacks (approved):

src/datatypes/datetime.ts
```diff
@@ -29,6 +29,14 @@ import {
 
 export const DATE_PRIVATE = Symbol.for("edgedb.datetime");
 
+function toNumber(val: any): number {
+  const n = Number(val);
+  if (Number.isNaN(n)) {
```
1st1, line 34: Can you add a comment here? I'm not sure this is correct just by looking at it.

test/client.test.ts
```diff
@@ -729,7 +729,7 @@ if (!isDeno) {
       new Duration(0, 0, 0, 0, 0, 0, 0, -1),
       new Duration(0, 0, 0, 0, 0, 0, 0, 1),
       new Duration(0, 0, 0, 0, 0, 0, 0, -1),
-      new Duration(0, 0, 0, 0, 0, 0, 0, -752043.296),
+      new Duration(0, 0, 0, 0, 0, 0, 0, -752043),
```
1st1, line 732: Duration no longer supports micros?

jaclarke: It does, it just no longer supports non-integer args to the Duration constructor. https://tc39.es/proposal-temporal/#sec-temporal-duration-objects

src/datatypes/datetime.ts
jaclarke: Done

1st1 (approved):

PR #362, opened 2022-06-03
Implement `e.group`

colinhacks:


jaclarke:
@colinhacks Should we only generate these group parts of the querybuilder if the introspected version of edgedb supports it? Or are we going going to handle it in documentation and just tell users not to use it if they're using edgedb <2.0?

src/syntax/group.ts
```diff
@@ -0,0 +1,184 @@
...
+        throw new Error(
+          `Cannot override pre-existing expression with key "${k}"`
+        );
+      }
+
+      return false;
+    });
+
+    return {
+      [`${Math.round(1000000 * Math.random())}___`]: {
```
jaclarke, line 66: Maybe use `Symbol()` here to ensure the key can't conflict

src/syntax/toEdgeQL.ts
```diff
@@ -388,45 +307,287 @@ export function $toEdgeQL(this: any) {
...
+      childExprs.push(...walkExprTree(expr.__expr__, expr, ctx));
+      break;
+    }
+    case ExpressionKind.Group: {
+      const groupingSet = expr.__grouping__ as any as GroupingSet;
+      for (const [_k, groupExpr] of groupingSet.__exprs__) {
+        // this prevents recurring grouping elements from being walked twice
+        // this way, these won't get pulled into with blocks,
+        // which is good because they need to be rendered in `using`
+        const seen: Set<any> = new Set();
```
jaclarke, line 536: I think we can bypass the whole repeated variable extraction here, since (if I'm understanding it correctly) the `using` block essentially works the same as `with` but actually within the scope of the `group` expression.

colinhacks:
> Should we only generate these group parts of the querybuilder if the introspected version of edgedb supports it? Or are we going going to handle it in documentation and just tell users not to use it if they're using edgedb <2.0

I don't think this is worth the trouble tbh. They'll get a very clear "GROUP isn't implemented yet" runtime error if they try this.

PR #368, opened 2022-06-14
Protocol v1

jaclarke:


src/baseConn.ts
```diff
@@ -774,42 +819,331 @@ export class BaseRawConnection {
...
+    );
+    wb.writeString(query);
+
+    if (state === Session.defaults()) {
+      wb.writeBuffer(NULL_CODEC.tidBuffer);
+      wb.writeInt32(0);
+    } else {
+      wb.writeBuffer(this.stateCodec.tidBuffer);
+      if (this.stateCodec === INVALID_CODEC) {
+        wb.writeInt32(0);
```
fantix, line 872: I'm not sure if this is correct - the server will likely fail to decode this zero-lengthed state bytestring.

src/baseConn.ts
```diff
@@ -104,6 +111,9 @@ export class BaseRawConnection {
   protocolVersion: ProtocolVersion = PROTO_VER;
   isLegacyProtocol = false;
 
+  protected stateCodec: ICodec = INVALID_CODEC;
```
fantix, line 114: Looks like `INVALID_CODEC` is now only used as an initial value, and the `INVALID_CODEC_ID` is never used. Maybe we can avoid calling it "invalid" as we don't have (or - at least - don't recommend) to use the `FFFF...` type id hack.

src/baseConn.ts
```diff
@@ -348,6 +390,11 @@ export class BaseRawConnection {
         break;
       }
 
+      case chars.$s: {
```
fantix, line 393: Perhaps it's better now to call `_parseDescribeStateMessage()` in the handler of parse/execute/connection, instead of `_fallthrough()` because we won't send state descriptors arbitrarily.

src/baseConn.ts
fantix: Ah right. I remember @elprans mentioned in one of the meetings that he's got some idea to handle the state desc for binary-over-http connections, I suppose this is it?

PR #390, opened 2022-07-18
fix: set code block highlight language

lu-zen:
Sets a missing code block language highlight.
Closes  edgedb/edgedb#3940


colinhacks:
Thanks!

PR #391, opened 2022-07-18
Use byteToHex to decode uuid

colinhacks:
Did some simple benchmarking on UUID parsing, see `qb/playground.ts` to see the various implementation.

I landed on an implementation from https://github.com/uuidjs/uuid/blob/main/src/stringify.js that's actually 22% faster than what we have currently and adds hyphens properly.

```
without-dashes: 821.39ms // current
with-dashes-bytetohex: 634.192ms // new
with-dashes-readbuffer: 1.714s
with-dashes-string-manip: 841.354ms
with-dashes-string-addition: 875.214ms
```

### DIFF
--- a/qb/playground.ts
+++ b/qb/playground.ts
@@ -46,9 +46,8 @@ for (let i = 0; i < 256; ++i) {
   byteToHex.push((i + 0x100).toString(16).slice(1));
 }
 
-export function parseWithByteToHex(buffer: Buffer) {
-  const arr = new Uint8Array(buffer);
-
+export function parseWithByteToHex(buf: Buffer) {
+  const arr = new Uint8Array(buf);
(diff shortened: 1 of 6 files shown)


qb/playground.ts
```diff
@@ -1,43 +1,125 @@
...
+  byteToHex.push((i + 0x100).toString(16).slice(1));
 }
 
-const arg: App = {
-  test: "asdf",
-};
-function test(arg: App) {
-  arg.test;
+export function parseWithByteToHex(buffer: Buffer) {
+  const arr = new Uint8Array(buffer);
```
jaclarke, line 50:
```suggestion
export function parseWithByteToHex(arr: Buffer)
```

PR #430, opened 2022-08-24
Fix empty array literal issue

colinhacks:


qb/test/select.test.ts
```diff
@@ -1192,6 +1192,18 @@ test("filter on link prop in nested path", async () => {
   }));
   await query.run(client);
 });
+
+// EdgeQL limitation
+// test("link prop on backlink", async()=>{
```
1st1, line 1197: Just curious - was commenting this out intentional?

colinhacks: Yes - this is something that doesn't work currently but I wanted to leave a record for later.

PR #451, opened 2022-09-26
Restructure and codegen

colinhacks:
- [x] Add `--file` mode
- [x] Finalize command for QB generation (currently `npx @edgedb/generate edgeql-js`)

For the moment, I've disabled CI based releasing to avoid any accidental publishing in advanced of the `edgedb-js` 1.0 release. We can release 1.0 manually and re-introduce down the line.

I also removed `--watch` mode for the moment. It could be implemented later, but third-party tools like `nodemon` are good enough for the moment. There's no decent file watcher package that exists on both NPM and deno.land/x which complicated the implementation.


packages/generate/src/queries.ts
```diff
@@ -0,0 +1,201 @@
...
+  // }
+
+  // find all *.edgeql files
+  // for query in queries:
+  //   generate output file
+}
+
+async function getMatches(root: string) {
+  return adapter.walk(root, {
+    match: [/\.edgeql$/],
```
Sikarii, line 98: Could this match a file without a name (`.edgeql`) and cause issues later down the line?

packages/generate/test/queries/tsconfig.build.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {"allowJs": false},
+  "include": [],
+  "files": [
+    "test.ts", "getMoviesStarring.edgeql.ts"
```
Sikarii, line 6: Is this missing freeShape.edgeql.ts?

Sikarii (commented):
Was slightly painful to review with all the dangling comma changes, but got there regardless.
I spotted a few things that seemed out of place to me, hope this helps.

Additionally, is it worth making the JS/TS client docs to be consistent with the dangling comma changes?

Sikarii:
Should the generators have their own directory in the default dir (`dbschema`), say `dbschema/generated/<generator>`?

This would prevent bloating the dbschema directory if we're gonna have numerous generators.
And in addition, this would make ignoring (for VCS and such) all the generated code very easy.

Lastly, what's the stance on supporting pluggable generators?
From what I understood, all the generators live in the EdgeDB-owned package, is there any room to have a plugin system of sorts?

docs/generation.rst
```diff
@@ -1,266 +1,139 @@
...
-  Checking the generated query builder into version control
-  is NOT RECOMMENDED. Would you like to update .gitignore to ignore
-  the query builder directory? The following line will be added:
+**Node.js**
 
-  dbschema/edgeql-js
+.. code-block:: bash
 
-  [y/n] (leave blank for "y")
+  $ npx edgedb generate <generator> [options]
```
Sikarii, line 76: Shouldn't this command be `npx @edgedb/generate <generator> [options]`?

docs/queries.rst
```diff
@@ -0,0 +1,207 @@
...
+      └── getUser.edgeql
+
+
+The following command will run the ``queries`` generator.
+
+.. code-block:: bash
+
+  $ npx @edgedb/generate queries
+
+The generator will detect the project root by looking for an ``edgedb.toml``, then scan the directory for ``*.edgeql`` files. In this case, there's onlny one: ``queries/getUser.edgeql``.
```
Sikarii, line 55:
```diff
- onlny
+ only
```

docs/queries.rst
```diff
@@ -0,0 +1,207 @@
...
+  ├── edgedb.toml
+  ├── index.ts
+  ├── dbschema
+  │   └── queries.ts  <-- generated file
+  └── queries
+      └── getUser.edgeql
+      └── getMovies.edgeql
+
+
+We can now use this function in our code.
```
Sikarii, line 161:
```diff
- use this function
+ use these functions
```

docs/querybuilder.rst
```diff
@@ -145,75 +106,82 @@ Then create ``import_map.json`` with the following contents. Both lines must be
     }
   }
 
-Optionally, you can specify a version:
+Version control
+^^^^^^^^^^^^^^^
 
-.. code-block:: json
+The first time you run generator, you'll be prompted to add the generated
```
Sikarii, line 112:
```diff
- The first time you run generator
+ The first time you run the generator
```

docs/querybuilder.rst
```diff
@@ -306,6 +295,60 @@ extracted with the ``$infer`` helper.
...
+  }
+
+  interface Movie {
+    id: string;
+    title: string;
+    genre?: Genre | null;
+    actors: Person[];
+  }
+
+Any types declared in a non-``default`` module  will be generated into an
```
Sikarii, line 349:
```diff
- module  will
+ module will

**Review by Sikarii — COMMENTED** [review]
Some nitpicks about the docs.
There were also some places where `.` was used in place of `:` (when listing things).

File: packages/generate/src/codecToType.ts [hunk 8]
```diff
@@ -0,0 +1,152 @@
...
+function generateSetType(type: string, cardinality: Cardinality): string {
+  switch (cardinality) {
+    case Cardinality.MANY:
+      return `${type}[]`;
+    case Cardinality.ONE:
+      return type;
+    case Cardinality.AT_MOST_ONE:
+      return `${type} | null`;
+    case Cardinality.AT_LEAST_ONE:
+      return `AtLeastOne<${type}>`;
```
jaclarke, line 85: We need to import this `AtLeastOne` type when it's used.

colinhacks: I inlined this instead

soilSpoon:
May I know when those changes will be released as npm packages?

PR #463, opened 2022-10-19
Add `is_required` to types

mc-0bit:


colinhacks:
The reason this doesn't exist currently is that this information is already contained in the `card` field — if the cardinality is `One` or `AtLeastOne`, it's `required`. I'm not opposed to adding this for convenience though.

PR #465, opened 2022-10-22
typo: unintended code sample

NurbekGithub:
I believe this code sample doesn't belong here and might cause confusion for newcomers


NurbekGithub:
the page https://www.edgedb.com/docs/clients/js/queries

Sikarii (approved):

PR #479, opened 2022-11-01
Fix Date handling for interface helper types

rileytomasek:
The propertyKeys<T> and linkKeys<T> were incorrectly treating Date properties as links because Dates are objects.

```ts
// Current
helper.Props<{ date: Date; someLink: SomeLink }> === {}
helper.Links<{ date: Date; someLink: SomeLink }> === { date: Date; someLink: SomeLink }

// Updated
helper.Props<{ date: Date; someLink: SomeLink }> === { date: Date; }
helper.Links<{ date: Date; someLink: SomeLink }> === { someLink: SomeLink }
```

### DIFF
--- a/packages/generate/src/edgeql-js/generateInterfaces.ts
+++ b/packages/generate/src/edgeql-js/generateInterfaces.ts
@@ -187,11 +187,11 @@ export const generateInterfaces = (params: GenerateInterfacesParams) => {
 
 export namespace helper {
   export type propertyKeys<T> = {
-    [k in keyof T]: NonNullable<T[k]> extends object ? never : k;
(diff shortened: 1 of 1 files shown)


jaclarke:
Good catch, and thanks for the PR. However `Date`s aren't the only property type that are a type of object (there's also the other local_date/time types, `Uint8Array`, and even arrays of properties for multi-props), so I think a better fix is approach from the other direction and to check for objects extending the `std::BaseObject` interface instead of plain `object` for links.
It looks like Colin has already done that in this PR: #480, so I'm going to close this PR in favour of that one.

rileytomasek:
For sure. I knew `Date` wouldn't fix everything, but didn't know which object to extend. @colinhacks quickly offered to do it for me :)

PR #482, opened 2022-11-07
Add an example for using the query builder in a transaction

raddevon:
A user created a repo comparing a task model implementation in Knex and EdgeDB. User was under the impression it was not possible to use the querybuilder with transactions. This is part of an effort to better communicate that it is possible.

See https://discord.com/channels/841451783728529451/1038780233458909324/1039094288354652180

Companion to https://github.com/edgedb/edgedb/pull/4640

### DIFF
--- a/docs/querybuilder.rst
+++ b/docs/querybuilder.rst
@@ -195,6 +195,7 @@ We can also run the same query as above, built with the query builder, in a
 transaction.
 
 .. code-block:: typescript
+
     const query = e.select(e.datetime_current());
     client.transaction(async tx => {
       const result = await query.run(tx);
--- a/docs/querybuilder.rst
+++ b/docs/querybuilder.rst
@@ -174,22 +174,6 @@ Now we have everything we need to write and execute our first query!
     }
     run();
 
-We can also run queries built with the querybuilder in a transaction.
-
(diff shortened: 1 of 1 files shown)


docs/querybuilder.rst
```diff
@@ -174,11 +174,27 @@ Now we have everything we need to write and execute our first query!
...
+    import e from "./dbschema/edgeql-js";
+
+    const client = createClient();
+
+    const query = e.select(e.datetime_current());
+    client.transaction(async tx => {
+      const result = await query.run(tx);
+      console.log(result);
+    });
+
```
jaclarke, line 192: Maybe this example should be in a note block at the end of the section, so it doesn't break up the flow of the 'run a first query' mini tutorial, and make it stand out a bit. Also, probably can skip the imports and `createClient` lines.

raddevon: By "end of the section," do you mean at the end of the quickstart section? Would you maybe put it in the cheatsheet section as another example, or is that too buried?

jaclarke: Yes, end of the quickstart section.

raddevon: Thank you. Got it moved.

PR #508, opened 2022-12-12
Implement better `__type__.name` type inference + infer result of polymorphic queries as discriminated union

### LINKED ISSUE #454: Better return types for `.__type__.name` and polymorphic queries if possible
A clear and concise description of the syntax, what you want to happen, and why it's needed.

### Part 1. `__type__.name`
Currently the return type of `.__type__.name` is just `string`. We know the name of the object type that's being selected, so it should be possible to infer this as the correct string literal (or union of string literals if it's a parent type).

### Part 2. Polymorphic queries
Currently a query like this:
```
e.select(e.Content, () => ({
  title: true,
  ...e.is(e.Movie, { release_year: true }),
  ...e.is(e.TVShow, { num_seasons: true }),
}));
```
gets a return type like:
```
{
  title: string;
  release_year: number | null;
  num_seasons: number | null;
}[]
```
While this is technically correct, it would be more useful if we could return a discriminated union. I think it might be possible for the querybuilder to automatically insert the typename in the shape, and return a type like so:
```
({
  title: string;
} & ({
  __typename: 'default::Movie';
  release_year: number;
} | {
  __typename: 'default::Show';
  num_seasons: number;
}))[]
```

jaclarke:
Closes #454


jaclarke:
Suggestion from Elvis: Now the v2 type descriptors contain type ids/names, we should query the typeid, and conditionally convert that to the typename client side to avoid performance penalty of joining types on the server.

PR #517, opened 2022-12-29
Fix generator to compile with TypeScript strict configs

NexZhu:
This fix allow generated code to compile in user projects with the following `tsconifg.json` options on:
```
{
  "compilerOptions": {
    /* Strict Type-Checking Options */
    "strict": true,                              /* Enable all strict type-checking options. */
    "noImplicitAny": true,                       /* Raise error on expressions and declarations with an implied 'any' type. */
    "strictNullChecks": true,                    /* Enable strict null checks. */
    "strictFunctionTypes": true,                 /* Enable strict checking of function types. */
    "strictBindCallApply": true,                 /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
    "strictPropertyInitialization": true,        /* Enable strict checking of property initialization in classes. */
    "noImplicitThis": true,                      /* Raise error on 'this' expressions with an implied 'any' type. */
    "alwaysStrict": true,                        /* Parse in strict mode and emit "use strict" for each source file. */

    /* Additional Checks */
    "noImplicitReturns": true,                   /* Report error when not all code paths in function return a value. */
    "noFallthroughCasesInSwitch": true,          /* Report errors for fallthrough cases in switch statement. */
    "noUncheckedIndexedAccess": true,            /* Include 'undefined' in index signature results */
    "noImplicitOverride": true,                  /* Ensure overriding members in derived classes are marked with an 'override' modifier. */
    "noPropertyAccessFromIndexSignature": true,  /* Require undeclared properties from index signatures to use element accesses. */
  }
}
```

### DIFF
--- a/packages/generate/src/edgeql-js/generateOperatorTypes.ts
+++ b/packages/generate/src/edgeql-js/generateOperatorTypes.ts
@@ -300,23 +300,23 @@ export function generateOperators({
     if (typeof args[0] === "string" && overloadDefs.Prefix[args[0]]) {
       op = args[0];
       params = [args[1]];
-      defs = overloadDefs.Prefix[op];
+      defs = overloadDefs.Prefix[op]!;
     } else if (typeof args[1] === "string" && overloadDefs.Postfix[args[1]]) {
       op = args[1];
       params = [args[0]];
-      defs = overloadDefs.Postfix[op];
+      defs = overloadDefs.Postfix[op]!;
     }
   } else if (args.length === 3) {
     if (typeof args[1] === "string") {
       op = args[1];
       params = [args[0], args[2]];
-      defs = overloadDefs.Infix[op];
+      defs = overloadDefs.Infix[op]!;
     }
   } else if (args.length === 5) {
     if (typeof args[1] === "string" && typeof args[3] === "string") {
       op = \`\${args[1]}_\${args[3]}\`;
       params = [args[0], args[2], args[4]];
-      defs = overloadDefs.Ternary[op];
+      defs = overloadDefs.Ternary[op]!;
     }
   }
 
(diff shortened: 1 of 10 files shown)


packages/generate/src/edgeql-js/generateOperatorTypes.ts
```diff
@@ -300,23 +300,23 @@ export function generateOperators({
     if (typeof args[0] === "string" && overloadDefs.Prefix[args[0]]) {
       op = args[0];
       params = [args[1]];
-      defs = overloadDefs.Prefix[op];
+      defs = overloadDefs.Prefix[op]!;
```
jaclarke, line 303: I think it's more correct to add `| undefined` to the type of `defs`, instead of adding a non-null assertion here, since `op` may not be a valid name and there's a check for no valid defs being found below.

jaclarke (commented):
Thanks @NexZhu, this looks great.

packages/generate/src/syntax/cardinality.ts
```diff
@@ -160,7 +160,7 @@ export namespace cardutil {
   >(cards: Cards): mergeCardinalitiesVariadic<Cards> {
     if (cards.length === 0) throw new Error("Empty tuple not allowed");
     if (cards.length === 1) return cards[0] as any;
-    const [first, second, ...rest] = cards;
+    const [first, second, ...rest] = cards as unknown as [Cardinality, Cardinality, ...Cardinality[]];
```
jaclarke, line 163:
```suggestion
    const [first, second, ...rest] = cards as unknown as [
      Cardinality, Cardinality, ...Cardinality[]
    ];
```
(Make tslint happy)

NexZhu:
@jaclarke updated😃

jcamden:
Thanks @NexZhu  and @jaclarke. I could use this. Is there a release that includes this fix?

PR #518, opened 2022-12-31
Fix event handling in fetch-based connection classes

freeatnet:
Fetch-based connection classes:
* never set `connWaiter` event, which is inherited from `BaseRawConnection`;
* set, but don't clear `messageWaiter` after receiving data.

This results in a "client has been closed" error when calling `Client.close()` on an HTTP client.

This PR fixes the issue above.

### DIFF
--- a/packages/driver/src/fetchConn.ts
+++ b/packages/driver/src/fetchConn.ts
@@ -133,6 +133,8 @@ class BaseFetchConnection extends BaseRawConnection {
       this.messageWaiter.set();
     } catch (e) {
       this.messageWaiter.setError(e);
+    } finally {
+      this.messageWaiter = null;
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/driver/src/fetchConn.ts
```diff
@@ -133,6 +133,8 @@ class BaseFetchConnection extends BaseRawConnection {
       this.messageWaiter.set();
     } catch (e) {
       this.messageWaiter.setError(e);
+    } finally {
+      this.messageWaiter = null;
```
freeatnet, line 137: Moved clearing the `messageWaiter` into a `finally` block so that it's definitely and for sure non-null in the catch block. Does that work ok for you?

jaclarke: Yes, this looks good to me.

jaclarke:
Thanks @freeatnet.

PR #522, opened 2023-01-09
Add more e.for examples, guidance

raddevon:
These additions were inspired by direct request from a user:

> examples of e.for with other operations such as updates,  and more importantly examples of bulk inserting from a javascript/typescript array rather than a json array and some explanation of how to actually go about doing that, for example should we even be using e.for for that or should we use a javascript loop.


docs/for.rst
```diff
@@ -43,3 +43,309 @@ in as a ``json`` parameter, converted to a set of ``json`` objects with
 Note that any optional properties values must be explicitly set to ``null``.
 They cannot be set to ``undefined`` or ommitted; doing so will cause a runtime
```
jaclarke, line 44:
```suggestion
They cannot be set to ``undefined`` or omitted; doing so will cause a runtime
```

docs/for.rst
```diff
@@ -43,3 +43,309 @@ in as a ``json`` parameter, converted to a set of ``json`` objects with
...
+          })
+        ),
+      },
+      (params) => {
+        const movies = e.for(
+          e.op(
+            "distinct",
+            e.array_unpack(e.array_unpack(params.characters).movies)
+          ),
+          (movie) => {
```
jaclarke, line 124: I would rename this variable to something else like `movieTitle`, so it's obviously different from `movie` in the `unlessConflict`.

docs/for.rst
```diff
@@ -43,3 +43,309 @@ in as a ``json`` parameter, converted to a set of ``json`` objects with
...
+      (params) => { ...
+
+In raw EdgeQL, you can only have scalar types as parameters. We could mirror
+that here with something like this: ``e.params({characters: e.json})``, but
+this would then require us to cast all the values inside the JSON like
+``portrayed_by`` and ``name``.
+
+By doing it this way — typing ``characters`` with ``e.array`` and the character
+objects as named tuples by passing an object to ``e.tuple`` — all the data in
+the array will be properly cast for us.
```
jaclarke, line 201: It will also better typecheck the data you pass to `run` as well.

docs/for.rst
```diff
@@ -43,3 +43,309 @@ in as a ``json`` parameter, converted to a set of ``json`` objects with
...
+          e.for(e.array_unpack(params.characters), (character) => {
+            return e.insert(e.Character, {
+              name: character.name,
+              portrayed_by: character.portrayed_by,
+              movies: e.assert_distinct(
+                e.for(e.array_unpack(character.movies), (movieTitle) => {
+                  return e.select(movies, () => ({
+                    filter_single: { title: movieTitle },
+                  }));
+                })
```
jaclarke, line 147: Instead of a `for` loop selecting each movie individually, it might be easier to do:
```ts
e.select(movies, () => ({
  filter: e.op(movie.title, 'in', e.array_unpack(character.movies))
}))
```

docs/insert.rst
```diff
@@ -153,3 +154,241 @@ in as a ``json`` parameter, converted to a set of ``json`` objects with
       {title: 'Deadpool 3', release_year: null},
     ]
   });
+
+Here's a more complex example, demonstrating how to complete a nested insert
+with conflicts on the inner items. First, take a look at the schema for this
+database:
+
```
jaclarke, line 161: Instead of duplicating this on both pages, maybe on this page it could just say that you can do bulk inserts with `e.for`, and link to the full example on the `for` page?

jaclarke:
There's also a stray 'The' after the first sentence on the 'For Loops' page.

raddevon:
Great suggestions, @jaclarke. I believe I have addressed them all. Would you care to take another look when you have time?

docs/insert.rst
```diff
@@ -131,25 +132,5 @@ Or you can perform an upsert operation with an ``e.update`` in the ``else``.
...
-  });
-
-  const result = await query.run(client, {
-    items: [
-      {title: 'Deadpool', release_year: 2016},
-      {title: 'Deadpool 2', release_year: 2018},
-      {title: 'Deadpool 3', release_year: null},
-    ]
-  });
+You can use a :ref:`foor loop <edgedb-js-for>` to perform :ref:`bulk inserts
```
jaclarke, line 135:
```suggestion
You can use a :ref:`for loop <edgedb-js-for>` to perform :ref:`bulk inserts
```

jaclarke (approved):

PR #526, opened 2023-01-16
Expose codec to decode/encode decimal type into/from string

jaclarke:


gauthierrodaro:
Is there anything I could do to help merge this PR?

PR #533, opened 2023-02-04
Check if crypto.subtle is undefined

jmn:
`crypto.subtle` can be undefined in some cases.

https://nodejs.org/api/webcrypto.html#cryptosubtle


jaclarke:
As far as I can tell, in Node the `SubtleCrypto` api was added in v15, but only as `require('crypto').webcrypto.subtle`, while the global `crypto` wasn't introduced until v19, so if the global `crypto` exists then `crypto.subtle` must also exist?
Can you give an example case where `crypto` exists but not `crypto.subtle`?

PR #538, opened 2023-02-09
Update interfaces.rst

aarongeorge:
Fixed incorrect reference to queries


raddevon:
Thanks for catching this, @aarongeorge, and for going out of your way to fix it! 🤘

PR #551, opened 2023-02-26
add (unofficial) support for http client in `@edgedb/generate`

freeatnet:
This PR adds unofficial support for the HTTP client in `@edgedb/generate`. This allows users to run generation actions against an instance that is only available over HTTP. Discord [ref](https://discord.com/channels/841451783728529451/1060204573156978688/1060551724026310786).


jaclarke:
Thanks @freeatnet. LGTM

Issue #540, opened 2023-02-13
Generating queries for Typescript (`--target ts`) results in import error/conflict


tjjfv:
When generating query functions for Typescript (`npx @edgedb/generate queries --target ts`), an EdgeQL file `<path>/<name>.edgeql` results in a TS file `<path>/<name>.edgeql.ts` (as expected per [the docs](https://www.edgedb.com/docs/clients/js/generation#:~:text=queries%3A%20Scans%20your%20project%20for%20*.edgeql%20files%20and%20generates%20a%20file%20containing%20a%20strongly%2Dtyped%20function%20alongside%20each.).

However, to `import` the TS file it must be specified as `.../<name>.edgeql` (since by default[^1] the filename must not include a `.ts` extension) and that results in 'importing' from the EdgeQL file at `.../<name>.edgeql`, which results in an error.

[^1]: Overriding the default currently requires using the 5.0 beta with a [new compiler option](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#allowimportingtsextensions) that restricts other common compiler configurations.

tjjfv:
Having the generated files under a different extension (`*.eql.ts` or `*.e.ts`[^1]) is likely to be the only viable solution for avoiding the conflict.

The generator should do this by default so that it works without additional configuration by every user.

[^1]: Personally, my preference is `*.e.ts`.

scotttrinh:
Confirmed the issue here. The `--file` option doesn't suffer from this, so here are some additional fixes here:

1. Remove the in-place conversion and just generate a single file, similar to how the other two generators work.
2. Remove the `edgeql` infix so that `<name>.edgeql` becomes `<name>.ts`
   - I consider the proposed solution here to be a bit of a variation that allows you to configure how that replacement works, but I think we can/should standardize on how we do this to eliminate confusion and lower the maintenance burden for documentation.

---

Of those 2.5 solutions, I'm personally leaning toward the first solution since it's the most congruent with the other generators, has the least number of edge cases, and is the simplest thing to implement. It _is_ a breaking change though, so I definitely want to hear from folks who use the `queries` generator about what the advantages of having separate modules/files for each query are.

scotttrinh:
Found this little note from @1st1 about preferring the co-located, inline generation: https://github.com/edgedb/edgedb-js/issues/314#issuecomment-1148114305

[...]

PR #568, opened 2023-04-10
Add support for generated suffixes (#540)

rcoedo:
This PR adds a `--suffix` option to customize the generated output file name. This will allow to work around the problem described in #540.

### DIFF
--- a/packages/generate/src/cli.ts
+++ b/packages/generate/src/cli.ts
@@ -181,6 +181,18 @@ const run = async () => {
         }
         options.out = getVal();
         break;
+      case "--suffix":
+        if (
+          generator === Generator.Interfaces ||
+          generator === Generator.Queries
+        ) {
+          options.suffix = getVal();
+        } else {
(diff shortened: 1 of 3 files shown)


scotttrinh:
@rcoedo 

Thanks for the contribution! I'd like to discuss some other possible solutions to this issue in #540 before considering merging this, so feel free to chime in over there. I'll leave this PR open for now, but my feeling is that we're not likely to want this to be configurable, so I expect we'll _at least_ modify this to have some different suffix that is non-configurable.

rcoedo:
@scotttrinh Agreed. This is a less than ideal solution and more like a workaround/hotfix for a feature that is currently broken. I'm fine with leaving the PR open, but I'm also fine with closing it while we think about a better approach.

scotttrinh:
Closed in favor of #622 

Thanks so much @rcoedo for the PR and helping us to figure out a way forward.

PR #575, opened 2023-04-20
Bump ring buffer size + abort conn if ring buffer overflows

### LINKED ISSUE #571: `RingBufferError: RingBuffer(capacity=2048) is full` persists until restart
**Describe the bug**
Follow up from [discord](https://discord.com/channels/841451783728529451/849374705210490900/1098016205199114293)

We have occasionally hit the following exception on certain queries. After the query fails, the client then continues to throw this on all subsequent, otherwise unaffected queries.

```
RingBufferError: RingBuffer(capacity=2048) is full
      at RingBuffer.enq (/home/credal/app/node_modules/edgedb/dist/primitives/ring.js:74:19)
      at ReadMessageBuffer.feedEnqueue (/home/credal/app/node_modules/edgedb/dist/primitives/buffer.js:487:19)
      at ReadMessageBuffer.feed (/home/credal/app/node_modules/edgedb/dist/primitives/buffer.js:483:25)
      at RawConnection._onData (/home/credal/app/node_modules/edgedb/dist/rawConn.js:155:33)
      at TLSSocket.emit (node:events:527:28)
```

**Reproduction**
Not caused by specific code/query but by selecting very large JSON fields. 

**Expected behavior**

1. Client doesn't die when selecting very large JSON fields
2. If client dies on a given query, it does not affect subsequent, non-problematic queries

- OS:
- EdgeDB version (e.g. `2.0`): 2.12+815c3e2
- EdgeDB CLI version (e.g. `2.0`): 2.3.1+76a728f
- `edgedb-js` version (e.g. `0.20.10`;): 1.1.0
- Node/Deno version:

**Additional context**

jaclarke:
Should fix #571.

Currently if the ring buffer that incoming data is placed into overflows, an error is thrown for the currently running query but the ring buffer is never cleaned up, so subsequent queries will also fail.
For now the safest fix seems to be to just abort the connection if the ring buffer overflows, since cleaning up the buffer properly will probably require some refactoring to track the boundaries of the separate messages in the buffer.

Also bump the ring buffer size by 4x, to allow larger results, though this also is a temporary fix, and we should probably find a way to remove this query result size limitation.

### DIFF
--- a/packages/driver/src/primitives/buffer.ts
+++ b/packages/driver/src/primitives/buffer.ts
@@ -28,7 +28,7 @@ const BUFFER_INC_SIZE: number = 4096;
 /* Max number of recv buffers that can be queued for
  * reading.
  */
-const BUFFER_RING_CAPACITY: number = 2048;
+const BUFFER_RING_CAPACITY: number = 8192;
 
 const EMPTY_BUFFER = new Uint8Array(0);
 
--- a/packages/driver/src/rawConn.ts
+++ b/packages/driver/src/rawConn.ts
@@ -174,9 +174,8 @@ export class RawConnection extends BaseRawConnection {
       if (this.messageWaiter) {
         this.messageWaiter.setError(e);
         this.messageWaiter = null;
-      } else {
-        throw e;
       }
+      this._abortWithError(e);
     }
 
     if (pause) {


scotttrinh:
@jackfischer 

Working on another PR that I'd like to get into a release, and I'm hoping to get that release together (which will include this change) late in the coming week.

PR #574, opened 2023-04-20
Update node version matrix for current LTS

scotttrinh:
Node v14 is end-of-life at the end of the month, and Node v20 is out, so this moves our testing matrix window forward based on the current LTS schedule. I picked v18 as our common version to test against in the extra versions since v16 is technically in maintenance mode and will be EOL in 6 months.

As of today:
image

Source: https://github.com/nodejs/Release

<details>
<summary>Release schedule JSON</summary>

```json
{
  "v0.8": {
    "start": "2012-06-25",
    "end": "2014-07-31"
  },
  "v0.10": {
    "start": "2013-03-11",
    "end": "2016-10-31"
  },
  "v0.12": {
    "start": "2015-02-06",
    "end": "2016-12-31"
  },
  "v4": {
    "start": "2015-09-08",
    "lts": "2015-10-12",
    "maintenance": "2017-04-01",
    "end": "2018-04-30",
    "codename": "Argon"
  },
  "v5": {
    "start": "2015-10-29",
    "maintenance": "2016-04-30",
    "end": "2016-06-30"
  },
  "v6": {
    "start": "2016-04-26",
    "lts": "2016-10-18",
    "maintenance": "2018-04-30",
    "end": "2019-04-30",
    "codename": "Boron"
  },
  "v7": {
    "start": "2016-10-25",
    "maintenance": "2017-04-30",
    "end": "2017-06-30"
  },
  "v8": {
    "start": "2017-05-30",
    "lts": "2017-10-31",
    "maintenance": "2019-01-01",
    "end": "2019-12-31",
    "codename": "Carbon"
  },
  "v9": {
    "start": "2017-10-01",
    "maintenance": "2018-04-01",
    "end": "2018-06-30"
  },
  "v10": {
    "start": "2018-04-24",
    "lts": "2018-10-30",
    "maintenance": "2020-05-19",
    "end": "2021-04-30",
    "codename": "Dubnium"
  },
  "v11": {
    "start": "2018-10-23",
    "maintenance": "2019-04-22",
    "end": "2019-06-01"
  },
  "v12": {
    "start": "2019-04-23",
    "lts": "2019-10-21",
    "maintenance": "2020-11-30",
    "end": "2022-04-30",
    "codename": "Erbium"
  },
  "v13": {
    "start": "2019-10-22",
    "maintenance": "2020-04-01",
    "end": "2020-06-01"
  },
  "v14": {
    "start": "2020-04-21",
    "lts": "2020-10-27",
[...]

PR #576, opened 2023-04-20
Assign private `EdgeDBError._message` before `defineProperties`

scotttrinh:
When working through the failing tests in #574 , I found this:

Screenshot 2023-04-20 at 1 46 24 PM

> TypeError: Cannot assign to read only property '_message' of object

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties#parameters)

> writable
> true if and only if the value associated with the property may be changed with an [assignment operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#assignment_operators). Defaults to false.

Couldn't find any mention of this behavior changing between 16 and 18, but this _does_ seem strictly safer than what we had before. I went back and forth on whether or not to set `writable` explicitly on the `defineProperties` call itself, and decided to do the smallest change here and discuss the context for this change before committing to something more explicit. I couldn't find where `_query` or `_attrs` is ever getting set, but perhaps it used to be used somehow?

---

Separately, I had quite a time trying to run the test suite locally and had to make some drastic changes to the test setup to accommodate the standard install path for the server binary on macOS (`~/Library/Application Support/edgedb/portable/{version}/bin/edgedb-server`). I spent a while trying to find the most flexible solution, but it's honestly pretty hacky. I'll try to improve this as we go.

### DIFF
--- a/packages/driver/src/errors/base.ts
+++ b/packages/driver/src/errors/base.ts
@@ -15,12 +15,12 @@ export class EdgeDBError extends Error {
   ) {
     // @ts-ignore
     super(undefined, options);
+    this._message = message ?? "";
     Object.defineProperties(this, {
       _message: {enumerable: false},
       _query: {enumerable: false},
       _attrs: {enumerable: false}
     });
-    this._message = message ?? "";
   }
 
   get message() {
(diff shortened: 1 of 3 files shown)


scotttrinh:
Ahh, I see that we use `_attrs` and `_query` in error handling in the `BaseRawConnection` class. I'll set all of the properties to be writeable, but since they're private, I wonder if we should expose a public setter instead? I don't think it's worth going through too much trouble just now for that, but I'll put it on the short list of things to explore.

jaclarke (approved):

PR #578, opened 2023-04-25
Update dependencies

scotttrinh:
Going to be adding a bunch of performance benchmarking and I want to start on a clean slate with TypeScript 5.0+ and thought it was a good time to do a little housekeeping on the dev dependencies.

Some things to note:

- Root packages made it hard to get consistent yarn upgrades and they should only really be used for root scripts.
- TSLint has long been deprecated, so switching to ESLint's TypeScript setup. For now, this will run on CI, but we'll ignore any failures

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,0 +1,14 @@
+/* eslint-env node */
+module.exports = {
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier"
+  ],
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  root: true,
+  rules: {
+    "@typescript-eslint/no-unnecessary-type-constraint": "warn"
+  }
+};
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -9,6 +9,10 @@ module.exports = {
   plugins: ["@typescript-eslint"],
   root: true,
   rules: {
-    "@typescript-eslint/no-unnecessary-type-constraint": "warn"
+    "@typescript-eslint/no-unnecessary-type-constraint": "warn",
+    // Svelte doesn't correctly compile if imports of the generated /modules
+    // aren't imported as 'import type' in other parts of the generated
+    // querybuilder, so set this option to ensure we always do that
+    "@typescript-eslint/consistent-type-imports": "error"
   }
 };
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,6 +72,10 @@ jobs:
         run: |
           yarn lint
 
+      - name: ESLint (ignore failures for now)
+        run: |
+          yarn eslint || echo "ESLint still failing... Fine for now!"
+
       - name: Build
         run: |
           yarn workspaces run build
--- a/packages/driver/src/datatypes/datetime.ts
+++ b/packages/driver/src/datatypes/datetime.ts
@@ -233,10 +233,10 @@ export class LocalDate {
 
   toString(): string {
     const year =
-      this.year < 1000 || this.year > 9999
+      this.year < 0 || this.year > 9999
         ? (this.year < 0 ? "-" : "+") +
           Math.abs(this.year).toString().padStart(6, "0")
-        : this.year.toString();
+        : this.year.toString().padStart(4, "0");
     const month = this.month.toString().padStart(2, "0");
     const day = this.day.toString().padStart(2, "0");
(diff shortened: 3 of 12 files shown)


.eslintrc.cjs
```diff
@@ -0,0 +1,18 @@
...
+  ],
+  parser: "@typescript-eslint/parser",
+  plugins: ["@typescript-eslint"],
+  root: true,
+  rules: {
+    "@typescript-eslint/no-unnecessary-type-constraint": "warn",
+    // Svelte doesn't correctly compile if imports of the generated /modules
+    // aren't imported as 'import type' in other parts of the generated
+    // querybuilder, so set this option to ensure we always do that
+    "@typescript-eslint/consistent-type-imports": "error"
```
scotttrinh, line 16: This ESLint rule replaces the now deprecated `tsconfig` flag `"importsNotUsedAsValues": "error"`. See [the PR introducing the change to TypeScript](https://github.com/microsoft/TypeScript/pull/52203) and [the section in the release announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax).

TL;DR: The intention of the `importsNotUsedAsValues` flag was never really to be a linting rule to disallow importing a type without the `import type` syntax. It also has weird interactions with `isolatedModules` and `preserveValueImports` which had some drastic effects on the transpiled JavaScript.

.github/workflows/tests.yml
```diff
@@ -72,6 +72,10 @@ jobs:
         run: |
           yarn lint
 
+      - name: ESLint (ignore failures for now)
+        run: |
+          yarn eslint || echo "ESLint still failing... Fine for now!"
+
```
scotttrinh, line 78: Making this an always passing step in the CI for now, and plan on tackling issues as I spot them in files that we touch over the next few months so we can switch over fully away from TSLint.

packages/driver/src/datatypes/datetime.ts
```diff
@@ -233,10 +233,10 @@ export class LocalDate {
 
   toString(): string {
     const year =
-      this.year < 1000 || this.year > 9999
+      this.year < 0 || this.year > 9999
         ? (this.year < 0 ? "-" : "+") +
           Math.abs(this.year).toString().padStart(6, "0")
-        : this.year.toString();
+        : this.year.toString().padStart(4, "0");
```
scotttrinh, line 239: This was failing, but it doesn't seem related to any changes I made here. I'm not sure why this is now breaking, but perhaps it was always flaky in tests since it is randomly generated? At any rate, here is the same bit of code from the polyfill we use:

```ts
  if (year < 0 || year > 9999) {
    const sign = year < 0 ? '-' : '+';
    const yearNumber = MathAbs(year);
    yearString = sign + `000000${yearNumber}`.slice(-6);
  } else {
    yearString = `0000${year}`.slice(-4);
  }
```
https://github.com/js-temporal/temporal-polyfill/blob/a5eda13f7ce9d2f2df899c2ac56e92936ec8d1bf/lib/ecmascript.ts#L2679-L2685

I think this update matches that logic precisely.

packages/generate/tsconfig.json
```diff
@@ -1,21 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    // "target": "es2017",
-    // "strictPropertyInitialization": false,
-    // "baseUrl": ".",
```
scotttrinh, line 6: I'm not a big fan of commented out code, so I just removed all this extra stuff 🤷

packages/driver/src/datatypes/datetime.ts
jaclarke: Looks like the polyfill version got updated from 0.4.2 to 0.4.3 in the lockfile, and the new version contains this change. Eventually when the `Temporal` api is finalised and makes its way into browsers/node/etc. we'll be able to throw all these custom datetime types away 🎉.

jaclarke (approved):
This all looks good to me.

packages/driver/src/datatypes/datetime.ts
scotttrinh: Ahh, of course I should've checked the `yarn.lock` file not just the `package.json` 🤦 .

> Eventually when the Temporal api is finalised and makes its way into browsers/node/etc. we'll be able to throw all these custom datetime types away 🎉.

Yeah, what has been our policy for runtime support in the past? If something is so easily polyfilled like this, it feels like it's fine to put that on the consumer to polyfill, yeah? I guess the same question could be said for our `fetch` polyfill, so maybe we can just make some kind of executive decision about it.

PR #579, opened 2023-04-25
Fix with var extraction in `for` loops + fix `insert unless conflict` child expr tracking + add tests

jaclarke:
In some cases when an expression is re-used inside a `for` expr, the `with` var it gets automatically extracted into gets bound to the outside of the `for` expr, which if the `with` expr references the loop var it causes an error. See: https://discord.com/channels/841451783728529451/849377751370432573/1098317767289090149

This PR fixes how the common parent scope of all usages of an expr is computed, so the `with` vars are correctly bound to the `insert` expr in this example. Also fixes a bug where child exprs in the `unless conflict` clause of an `insert` were not being correctly tracked as children of the `insert` expr.

### DIFF
--- a/packages/generate/src/syntax/toEdgeQL.ts
+++ b/packages/generate/src/syntax/toEdgeQL.ts
@@ -209,10 +209,7 @@ export function $toEdgeQL(this: any) {
       // first, check if expr is bound to scope
       let withBlock = refData.boundScope;
 
-      // filter nulls
-      const parentScopes = [...refData.parentScopes].filter(
-        scope => scope !== null
-      ) as WithScopeExpr[];
+      const parentScopes = [...refData.parentScopes];
 
       // if expression is unbound
       if (!withBlock) {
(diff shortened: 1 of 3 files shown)


packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -385,14 +391,16 @@ function walkExprTree(
     // const arg = (expr as any)?.__parent__ || (expr as any)?.__name__;
     // if (arg) console.log(arg);
     // }
-    seenExpr.parentScopes.add(parentScope);
+    if (parentScope) {
+      seenExpr.parentScopes.add(parentScope);
+    }
```
scotttrinh, line 396: Just a question for my own context: Was there some reason we were inserting `null` scopes before?

packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -223,22 +220,31 @@ export function $toEdgeQL(this: any) {
           continue;
         }
 
-        // set withBlock to top-level parent scope
-        const resolvedParentScopes = parentScopes.map(
-          parentScope => withVars.get(parentScope)?.scope ?? parentScope
+        const [firstParentScopeChain, ...parentScopeChains] = parentScopes.map(
```
scotttrinh, line 223: Took me a few read throughs to grok what was happening here, but in my attempt to form a coherent question, I was able to figure it out 😆

packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -1048,7 +1060,7 @@ function renderEdgeQL(
     return `(${renderEdgeQL(expr.__expr__, ctx, false, true).slice(
       1,
       -1
-    )} ${clause.join("")})`;
+    )}${clause.join("")})`;
```
scotttrinh, line 1063: 🔍 👏

scotttrinh (approved):
Looks good!

packages/generate/src/syntax/toEdgeQL.ts
jaclarke: Ah, you're right, there was a reason. We do need to track if an expr is used outside of a scope, so we can throw an error if it needs extracting to a `with` var, since there'll be no place to put it.

jaclarke: Fixed

scotttrinh (approved):

PR #580, opened 2023-04-26
Use prettier with no config

scotttrinh:
Alright, this is a little gratuitous, but I think the point of the entire ecosystem adopting prettier is to have a rust-fmt/gofmt kind of experience where you just let the formatter do it's thing without arguing about how to configure it. It turns out, we hadn't been enforcing the _other_ rules for a while and there were already 31 files in violation anyway, so I also introduced a check with prettier to CI to ensure we are consistent going forward. I personally hate these big reformatting commits, and the `--ignore-revs` feature in git is a pain to setup and onboard people into, but I think the pros slightly outweigh the cons here.

I don't feel _too_ strongly about this, so if this just feels too noisy, I'm fine with just keeping things as is and fixing formatting as we touch older files.


jaclarke:
I think this makes sense. And I'm fine with merging this all now; I think it's probably less noisy than spreading it across future commits as we touch files, particularly if we use `--ignore-revs`. (I had somehow never come across this git feature before 😄).
Maybe we should also add a `.git-blame-ignore-revs` file?: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

jaclarke (approved):

lu-zen:
use [rome](https://rome.tools/) instead
> )

scotttrinh:
Going to wait for a few more things to land and then rebase this before merging.

PR #595, opened 2023-05-09
Add type intersection docs to select page

raddevon:
This is addressing a request from a Discord user to mention type intersections in the `select` documentation: https://discord.com/channels/841451783728529451/1046803633548841031/1047104928046403675


docs/select.rst
```diff
@@ -490,6 +490,19 @@ fact that they will only occur in certain objects.
   with ``e.is``, the ``id`` property will be filtered out of the polymorphic
   shape object.
 
+
+Type intersection
+-----------------
+
+You can use the type intersection operator to filter a link based on its type.
```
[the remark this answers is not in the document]
scotttrinh: Oh, and maybe link to the [type intersections](https://www.edgedb.com/docs/clients/js/objects#type-intersections) section of the Objects and Paths page? Maybe even the more detailed docs in the [Standard Library](https://www.edgedb.com/docs/stdlib/set#operator::isintersect)?

docs/select.rst
```diff
@@ -490,6 +506,11 @@ fact that they will only occur in certain objects.
   with ``e.is``, the ``id`` property will be filtered out of the polymorphic
   shape object.
 
+
+Type intersection
+-----------------
+
+
```
scotttrinh, line 513: ✂️

docs/select.rst
```diff
@@ -289,6 +291,20 @@ Filters on links
   }));
 
 
+You can also use the :ref:`type intersection
+<edgedb-js-objects-type-intersections>` operator to filter a link based on its
+type.
```
scotttrinh, line 296: Do you think we can add some more detail about when and why you might need to do this? I think the docs in the Standard Library do a decent job here, but it might be helpful to relate it to _this_ specific schema. Maybe something like:

```suggestion
You can also use the :ref:`type intersection
<edgedb-js-objects-type-intersections>` operator to filter a link based on its
type. For example, since `actor.roles` might be of type `Movie` or `TVShow`,
to only return `roles` that are `Movie`s, you would use the `.is` type
intersection operator:
```

scotttrinh (changes requested):
Just removing the extra heading is the only blocker here, you can take or leave the extra detail in the description based on your judgement.

docs/select.rst
raddevon: 🤦‍♂️

raddevon:
I like the extra detail too. I think we're in good shape now!

scotttrinh (approved):

PR #598, opened 2023-05-11
Properly quote enum values when generating query text

### LINKED ISSUE #577: EdgeQLSyntaxError: Unexpected keyword "IN" / mixing string type literals and type names is not supported
**Code**
The code causing the error.

```typescript
(async() => {
  let databaseQuery;

  switch((query.type as Record["type"])) {
    case "A":
    case "AAAA":
    case "CNAME":
    case "DNAME":
    case "NS":
    case "PTR": {
      databaseQuery = e.select(
        e.insert(e.PlainRecord, { ...query }),
        () => ({ ...e.PlainRecord["*"] })
      );
      break;
    }

    default: {
      break;
    }
  }

  try {
    response = await databaseQuery.run(client);
    return { detail: response };
  } catch(_) {
    return { detail: response };
  }
});
```

**Schema**

Your application schema.

```
module default {
  scalar type ClassType extending enum<"CH", "CS", "HS", "IN">;

  abstract type BaseRecord {
    required property class -> ClassType;
    required property created -> datetime {
      default := datetime_of_transaction();
      readonly := true;
    };
    required property name -> str;
    required property ttl -> float64;
    required property type -> RecordType;
    required property updated -> datetime {
      default := datetime_of_transaction();
    };
  }

  type PlainRecord extending BaseRecord {
    required property data -> str;
  }
}
```

**Generated EdgeQL**

```
SELECT (INSERT default::PlainRecord {
  name := "app.beachfront",
  type := default::RecordType.A,
  class := default::ClassType.IN,
  ttl := 300
}) {
  data,
  class,
  name,
  ttl,
  type,
  created,
  updated,
  id
}
```

**Error or desired behavior**

SS 2023-04-20 at 1 23 12 PM

- OS: macOS Ventura
- EdgeDB version (e.g. `2.0`): 2.5+5f12a50
- EdgeDB CLI version (e.g. `2.0`): 2.3.1+5d93f42
- `edgedb-js` version (e.g. `0.20.10`;): 1.1.0 (latest Deno version)
- `@edgedb/generate` version (e.g. `0.0.7`;): 1.1.0 (latest Deno version)
- Typescript version: 5.0.3
- Deno version: 1.32.2 (release, aarch64-apple-darwin)

scotttrinh:
We already had a mechanism for quoting values when inserting them into generated queries, we just were not calling it in the case of enums. Will do a little more auditing and open additional PRs if there are more places where we're missing this. I wish I could've thought of a better keyword that is an actual movie genre, but alas.

I also did a non-essential optimization here: Use `Set.has` instead of `Array.includes` to lookup reserved keywords. Probably a pretty small performance impact, but since `.toEdgeQL` gets called a lot, thought it was worth optimizing while touching this code. Any reason to think we'll have platform support issues here? It looks like we're using `Set`s in other places in the code already, but thought it was worth pointing out.

Closes #577

### DIFF
--- a/packages/driver/genReservedKeywords.js
+++ b/packages/driver/genReservedKeywords.js
@@ -6,7 +6,7 @@ const execFile = util.promisify(require("child_process").execFile);
 const prettier = require("prettier");
 
 (async () => {
-  const {stdout: grammar} = await execFile("edb", [
+  const { stdout: grammar } = await execFile("edb", [
     "gen-meta-grammars",
     "edgeql",
   ]);
@@ -16,15 +16,17 @@ const prettier = require("prettier");
 
   const reservedKeywords = resKeywords
     .split("\n")
-    .map(line => line.trim().replace(/,$/, "").slice(1, -1))
-    .filter(keyword => !!keyword);
+    .map((line) => line.trim().replace(/,$/, "").slice(1, -1))
+    .filter((keyword) => !!keyword);
 
   await fs.writeFile(
     path.join(__dirname, "../src/reflection/reservedKeywords.ts"),
-    `export const reservedKeywords = ${JSON.stringify(
-      reservedKeywords,
-      null,
-      2
-    )}\n`
+    prettier.format(
+      `export const reservedKeywords = new Set(${JSON.stringify(
+        reservedKeywords,
+        null,
+        2
+      )})\n`
+    )
   );
 })();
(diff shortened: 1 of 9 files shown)


jaclarke (approved):

freeatnet:
> I also did a non-essential optimization here: Use `Set.has` instead of `Array.includes` to lookup reserved keywords.

FYI, this causes an exception — `reservedKeywords.includes is not a function` — if `edgedb` is upgraded to 1.2.2 without upgrading `@edgedb/generate` to 0.1.1.

scotttrinh:
> FYI, this causes an exception — reservedKeywords.includes is not a function — if edgedb is upgraded to 1.2.2 without upgrading @edgedb/generate to 0.1.1.

Thanks for the report! We're going to be synchronizing version numbers a little better in the very near future which should help alleviate some of the surprise here.

PR #606, opened 2023-05-17
Allow assigning subtypes in queries

### LINKED ISSUE #592: Making overloaded link required and subtyped in "grandchild" of inherited type breaks query builder
**Code**

```typescript
e.insert(e.StrategyGamer, {
    faveGame: strategyGame,
//  ^^^^^^^^^^^^^^^^^^^^^^ link `faveGame` is not allowed even though it should be
    firstName: "A",
    lastName: "A",
});
```

**Schema**

```
abstract type Game {
    # ...
}
type FpsGame extending Game {
    # ...
}
type StrategyGame extending Game {
    # ...
}

abstract type Person {
    link faveGame -> Game; # 👈 optional link
    required property firstName -> str;
    required property lastName -> str;
}
abstract type Gamer extending Person {
    # ...
}
type StrategyGamer extending Gamer {
    overloaded required link faveGame -> StrategyGame; # 👈 overloaded link, required and specialised subtype
}
```

**Generated EdgeQL**

The EdgeQL does generate correctly. It's just TS which isn't happy with it.

```
INSERT default::StrategyGamer {
  faveGame := (
    INSERT default::StrategyGame {
      # ...
    }
  ),
  firstName := "A",
  lastName := "A"
}
```

**Error or desired behavior**

`faveGame` is not an allowed property when trying to insert a new `StrategyGamer`. But not only should it be _allowed_ - because `StrategyGamer` extends `Person` which has a `link faveGame -> Game` - but it should actually be _required_; since `StrategyGamer` overloads the link to make it required + more specialised.

This bug only seems to happen when the overloaded link is _both_ required _and_ refers to a subtype, but not when only one of those two changes are made.

**There is no `faveGame` in the allowed insert query fields, not even optional**
image

**But inserting a new `Gamer` - i.e. a direct inheritor of `Person` where `faveGame` is defined - _does_ correctly know about `faveGame`**
image

- OS: Windows 10 Home build 19045.2846
- EdgeDB version (e.g. `2.0`): 2.13+c171f92
- EdgeDB CLI version (e.g. `2.0`): 2.3.1+ef99779
- `edgedb-js` version (e.g. `0.20.10`;):
- `@edgedb/generate` version (e.g. `0.0.7`;): 0.1.0
- Typescript version: 5.0.4
- Node/Deno version: Node v18.15.0

scotttrinh:
Closes #592 

This adds a new property to the type that contains all of the subtypes of a given type, so that when testing assignment, instead of comparing objects structurally we compare just that the current type is a supertype of the assigning type by comparing the listed subtypes of the assigning object.

In other words, if a type chain is A -> B -> C -> D, then type A will reference all 4 types, and attempts to assign C (which contains C and D) will succeed.

Note that this will break if anyone was relying on structural typing before since if two unrelated types were being used before but shared the same structure, the old assignment type was typechecking. I think it's probably the case that it would've been a runtime error, so I'm not concerned that there are valid use cases, but it's worth having a think about it before running with this.


Arrow7000:
I can't really wrap my head around the type generation code, but this does seem to fix my use case!

image

Arrow7000:
There seems to be something wrong with the default exports though. I'm getting this, but the one at the bottom looks very off.

image

I'm guessing it's from this part of my schema? Looks like it was mis-parsed though.

Code_njZX8SUFej

Arrow7000:
I don't really understand how `TypeSet` works, but it seems wrong that it contains `Cardinality.AtMostOne | Cardinality.One | Cardinality.Empty` when this field only has a cardinality of exactly `Cardinality.One`?

image

scotttrinh:
Going to set this as a draft for now while I add a bunch of tests in a separate set of PRs that branch from this.

Arrow7000:
Hey @scotttrinh, just checking in on this PR. I don't know if this was ready to merge or not, was it waiting on anything else?

scotttrinh:
@Arrow7000 

Yeah, it's mostly waiting on me to have the time to work out the implications "downstream" from implementing what is essentially nominal typing within the TypeScript client. No ETA on this, but it's not ready to merge yet. We're hard at work on finish up some big stuff for EdgeDB 4.0 coming soon, but after that I'm going to circle back to a bunch of WIP stuff on the JS client side.

I appreciate your patience!

PR #609, opened 2023-05-19
Add examples of referencing non-default modules

raddevon:
I notice we already have an example of this in "Globals" just above, but it could be easy to miss if you're not interested in referencing globals in particular. I wanted to draw attention to a couple of examples of how to reference entities on non-`default` modules with the query builder. (Recycled the one in "Globals," but I think it's useful to call attention to it again here.)


scotttrinh (approved):
😍

PR #610, opened 2023-05-19
Use `node:assert/strict` instead of Jasmine `expect` in generator tests

scotttrinh:
Wrote a [quick-and-dirty jscodeshift code mod](https://github.com/scotttrinh/expect-to-assert) to do this automatically, but there were a few manual interventions I had to do. I will outline those inline.

## Motivation
1. BETTER TYPES! Most of the `assert` functions are actual TypeScript assertion functions, so we can eliminate a bunch of `!` and `?` and `as any` calls littered around the test code. I, for one, would rather have the tests have the tightest possible types since we're relying on the tests as automated checks.
1. I find it way easier to read and understand:
    ```ts
    assert.equal(a, b, "a is b");
    ```
    than
    ```ts
    // A is b
    expect(a).toBe(b);
    ```
2. The new assert diffs are fire, and having the message in the failing tests is a great quality of life improvement.
    New assert diff with message:
    Screenshot 2023-05-17 at 2 35 20 PM
    Jasmine-style expect:
    Screenshot 2023-05-17 at 2 35 35 PM

4. I found one case where we accidentally didn't make any assertion at all (`expect(a)`) 🙈 
5. There are so many different Jasmine-style assertions that it becomes a bit of a speed-bump when trying to choose which assertion to use in a given case.
6. Now that the [Node test runner](https://nodejs.org/api/test.html) is stable on the newest LTS versions, it would be great to drop Jest to make the development machinery more lean. 

## Misc Notes
- Other test runners (👀 `node:test`) require wrapping any `beforeAll`/`afterAll` calls in a `describe` block, and I honestly think it makes it a little clearer, so I just wrapped _every_ test in a generic `describe` block. Pretty much a no-op for Jest, but will make the eventual conversion a lot easier. In the future we can make multiple suites in a file to increase parallelism and group tests together.
- I did some slight refactoring in some places to match the eslint rules that I'm trying to implement slowly, so there are some changes related to that here.
- I didn't convert the driver tests in this same PR, mostly due to time, but I'll do that the next time I need to add a bunch of tests there.

### DIFF
--- a/packages/generate/test/select.test.ts
+++ b/packages/generate/test/select.test.ts
@@ -190,7 +190,6 @@ describe("select", () => {
     type q1 = $.BaseTypeToTsType<(typeof q1)["__element__"]>;
     tc.assert<tc.IsExact<no_shape, q1>>(true);
     assert.deepEqual(no_shape.__element__.__shape__, q1.__element__.__shape__);
-    // expect(no_shape.__element__.__shape__).toEqual({id: true});
 
     // allow override shape
     const override_shape = e.select(q1, () => ({
@@ -436,14 +435,6 @@ describe("select", () => {
       tc.IsExact<(typeof q9)["__cardinality__"], $.Cardinality.AtMostOne>
     >(true);
     assert.deepEqual(q9.__cardinality__, $.Cardinality.AtMostOne);
-
-    // const q10 = e.select(e.Villain, villain => ({
-    //   filter_single: e.op(villain.name, "=", e.cast(e.str, e.set()))
-    // }));
-    // tc.assert<tc.IsExact<typeof q10["__cardinality__"], $.Cardinality.Empty>>(
-    //   true
-    // );
-    // expect(q10.__cardinality__).toEqual($.Cardinality.Empty);
   });
 
   test("infer cardinality - object type filters", () => {
@@ -586,8 +577,9 @@ describe("select", () => {
     );
 
     const result = await query.run(client);
-    expect(result).toMatchObject(data.iron_man);
-    assert.deepEqual(result?.villains, [{ id: data.thanos.id }]);
+    assert.ok(result);
+    assert.deepEqual(result, { ...result, ...data.iron_man });
+    assert.deepEqual(result.villains, [{ id: data.thanos.id }]);
   });
 
   test("computables", async () => {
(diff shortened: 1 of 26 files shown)


packages/generate/test/select.test.ts
```diff
@@ -1,761 +1,731 @@
...
-  const result = await query.run(client);
-  expect(result).toEqual(
-    [data.cap, data.spidey, data.iron_man]
-      .map((h) => ({ id: h.id }))
-      .sort((a, b) => a.id.localeCompare(b.id))
-  );
-});
+    const result = await query.run(client);
+    assert.ok(result);
+    assert.deepEqual(result, { ...result, ...data.iron_man });
```
scotttrinh, line 581: Here's my solution to the loosy `toMatchObject`: merge the expectation with the result.

packages/generate/test/select.test.ts
```diff
@@ -1010,433 +1015,358 @@ SELECT __scope_0_defaultPerson {
...
-    title: true,
-    filter_single: e.op(movie.genre, "=", e.Genre.Action),
-  }));
+    await query.run(client);
+  });
 
-  await expect(query.run(client)).rejects.toThrow(CardinalityViolationError);
-});
+  test("filter_single expect error", async () => {
+    // @ts-expect-error filter_single only types exclusive props
```
scotttrinh, line 1337: eslint wants you to write a "reason" for the `ts-expect-error` clauses, which end up serving a similar function as the `assert` functions' message argument, although you don't see them. Still, kinda nice.

packages/generate/test/sets.test.ts
```diff
@@ -1,200 +1,208 @@
...
-    e.tuple({ a: 3, b: "asdf", c: e.int64(5) })
-  );
-  expect(q1.__element__.__kind__).toEqual(TypeKind.namedtuple);
-  expect(await q1.run(client)).toMatchObject([
-    { a: 1, b: "asdf", c: 214 },
-    { a: 3, b: "asdf", c: 5 },
-  ]);
-
-  expect(() => e.set(e.tuple({ a: 1 }), e.tuple({ a: "asfd" }))).toThrow();
-  expect(() => e.set(e.tuple({ a: 1 }), e.tuple({ a: "asfd", b: "qwer" })));
```
scotttrinh, line 184: Here is the test that accidentally did not have an assertion.

PR #619, opened 2023-05-30
Update JS/TS client documentation overview

### LINKED ISSUE #607: Improve JS/TS client overview documentation
- Requirements should be minimized.
- Installation instructions should be minimized.
- You should see code very early in the first page to get a sense of the look and feel of the library.
  - Idea: Show a complex EdgeQL query using all three generators
- Remove dated references to `v1.0` being recent.

scotttrinh:
Closes #607 

## Remove references to "new" things.

"New" is relative, but the things we're calling out here are several months old at this point.

## Get to the code a _lot_ sooner in the overview.

In general, collapse a lot of the details that you would gloss over when first wanting to evaluate the library itself (installation instructions, requirements, etc) and show the code!

Here are some screenshots of before and after:

#### Before
www edgedb com_docs_clients_js_index

#### After
localhost_3000_docs_clients_js_index


docs/index.rst
```diff
@@ -27,123 +27,166 @@ EdgeDB TypeScript/JS Client
...
-connection to your database and execute queries. If you prefer writing queries
-as strings, the Client API is all you need.
+    .. code-tab:: typescript
+      :caption: deno
 
-.. code-block:: javascript
+      import * as edgedb from "http://deno.land/x/edgedb/mod.ts";
 
-  const edgedb = require("edgedb");
+Requirements
```
raddevon, line 64: I think it makes sense to keep these above installation since that's how it works chronologically — meet the requirements, then install.

raddevon (changes requested):
Thanks for tackling this, Scott. I like the way you've simplified and updated. In particular, moving the note about the query builder vs. an ORM to the query builder page seems like the right move to me.

I left some comments, mostly regarding how we organize the examples. Please follow up if it doesn't make sense or if you think it's not the right way forward.

raddevon:
Well, something happened to most of those comments I left. GitHub seems to have eaten them. 😖 

I'll type them up again. Comments forthcoming!

docs/index.rst
scotttrinh: Yeah, that's fair, I was definitely optimizing for visual scanning where you land on this page and just want to install it and the requirements bit is something you might not ever look at since you'd probably only notice you don't meet the requirements if something isn't working.

Also, what do you think about how I've stated the requirements? Basically instead of saying minimum versions (which is hard to keep in mind while adding features), I've listed what versions we actually test that they work?

raddevon:
Here's the summary of my recommendations re: examples organization:

- Split the examples paragraph after talking about the client API and put just that example after that part of the paragraph
- Use only a string query in the client-only example (since that's what we mention in the paragraph and since it's the simplest way to use that method of querying; add a comment noting you can import a query file and use that query instead of hard-coding a query if you want, unless we absolutely don't want users to hard-code text queries, in which case we should make that explicit to explain why we're not showing a simpler example)
- Split each of the generator examples into separate example sub-sections
- Provide a brief intro sentence or two for each generator example explaining why and when you'd use that generator
- Repeat the "EdgeQL" tab across any of the examples that use it (query and interfaces generators if you decide to switch to a raw string for the client-only example)
- Rename the "EdgeQL" tab to reflect its path and filename to make it easier to reason about how the pieces fit together
- Provide the most basic versions of each generator command required to generate the files used for that example

I'll follow up with a more detailed rationale.

raddevon:
The recommendations were driven by two main problems I noticed:

1. Showing examples of different ways to do the same thing without context for which you should use could be confusing, hence the recommendation for splitting into sections and adding a sentence or two to each of the examples.
2. The things we've organized into a single set of tabs here are not alike. The first four are ways to write queries. The last one is a supporting file for the middle three but does not interact with the first one at all. Tabbed content should be either related to a single alternative or should show multiple alternative ways to do one thing. These tabs try to do both of those things, and it makes the examples difficult to understand. If we split the examples apart, ditch tabs where we no longer need them, and use tabs per example to show pieces of that example, I think it gets easier to parse.

docs/index.rst
raddevon: That makes sense. I think you've shortened the requirements enough that you don't lose much by bumping it up.

I like that approach! I might flip the perspective to instead focus on our user instead of us. I'll write a proper suggestion below.

raddevon: Actually, now I'm second guessing that. 😅  I would generally prefer to write from the user perspective where it's currently written from our perspective: "We test…" vs. "You should use…" (or something similar)

But actually I think what you're trying to do is to say that it may work under some other version. By saying "We test…" you've leaving that possibility open. Am I understanding your intent?

scotttrinh:
@raddevon 

Thanks for the feedback, lemme explore my intention to see if we can come to some outcome that incorporates the feedback.

In general, my hope is _not_ to explicitly teach the reader what each generator does, how to use it, when to use it, etc. It is to give a brief scannable example of what the code looks and feels like without overwhelming you with things to read and understand first. I want all of this to fit with a minimal amount of scrolling so that this overview page feels lightweight and approachable. The [DrizzleORM Quickstart](https://orm.drizzle.team/docs/quick-start) is a good example of what I'm going for here. 

> Use only a string query in the client-only example (since that's what we mention in the paragraph and since it's the simplest way to use that method of querying;[..snip..]

> Tabbed content should be either related to a single alternative or should show multiple alternative ways to do one thing. These tabs try to do both of those things, and it makes the examples difficult to understand.

Yeah, that's a fair critique. I was attempting to get a sense of reusability by putting the query in a separate logical place and showing _just_ the parts of the API that I'm trying to highlight to the user in each case.

---

What do you think of this approach: Have a code block that just shows the query in a const string, and the driver-only example first. After that, show a tabbed code sample that shows the query with the three different generators. I realize that the query builder wouldn't use that string, but I hope that's not that big of a deal. I'm attempting to balance two things: "look, it's easy to get started with just the driver and string queries!" against "we recommend using the query builder for maximum safety and developer experience!", so I don't want to push the query builder too far down the page or spend too much time on the queries and interfaces generators.

docs/index.rst
scotttrinh: Yeah, that's my intent: these are not minimum versions, they are the versions we promise to test and support. They may work on earlier versions, and there are probably currently known versions, but without testing on those versions there is no way for us to know if we've broken capability with those versions.

scotttrinh: To be honest, I could see just moving this whole section to the repo's README since it probably introduces more friction than it alleviates. Our support posture isn't that radical here, and it's not likely to save us from anything unless we have some kind of runtime/install-time checking in place.

raddevon: That sounds fine to me, @scotttrinh!

scotttrinh:
> Having said that, when I see "Overview," that means to me that you're going to show me a little bit of all the basics. It implies brevity in comparison to the full-fat documentation but not in any absolute sense.

It appears we have different perspectives and objectives for this section and there-in lies our difference in emphasis. In the end, I want some where for a user like me (experienced application developer) to get a quick sense of what it's like to use a tool to evaluate it, not to generate some running hello world code that has no connection to what it looks like to actually use the tool. The existing Query Builder page does a great job of this by putting a context-free bit of example code fully above the fold right up front to give you a taste, and I'm trying to bring that here. Maybe it's not possible to get a brief enough overview of all four types of queries that satisfies my own desire for "just a taste".

In that case, maybe we just keep most of what we have after the installation instructions and just update the client example to do the same query as the query builder example?

scotttrinh:
After a sync call with @raddevon , we've decided to explore a more fully-fleshed out Overview section.

I'm basically picking up from where the Quickstart ends and demonstrating the progression of the client -> query builder. I start with the Client API to seed and query the database. I then layer on the interfaces generator to give some type information to the Client API query. Next I show how the queries generator give us the types for free. Finally, I re-write the string query using the query builder.

I haven't yet touched the other pages and I think we can definitely clean them up now that the overview page is much more fleshed out.

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+    console.log(JSON.stringify(result, null, 2));
+  }
+
+  main();
+
+Interfaces
+^^^^^^^^^^
+
+Since we're using TypeScript, it would be nice to be able to type the return
+value of this query, so let's use our first generator, the :ref:`Queries
```
raddevon, line 200: Should `Queries` be capitalized here? Actually, should this maybe be `the interfaces generator` and link to `edgedb-js-interfaces` instead?

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
 
+  The first time you run the generator, you'll be prompted to add the generated
+  files to your ``.gitignore``. Confirm this prompt to automatically add a line
+  to your ``.gitignore`` that excludes the generated files.
 
+  For consistency, we recommend omitting the generated files from version
+  control and re-generating them as part of your deployment process. However,
+  there may be circumstances where checking the generated files into version
+  control is desirable, e.g. if you are building Docker images that must contain
+  the full source code of your application.
```
raddevon, line 307: 🤌

I absolutely love that you tell them why!

raddevon (approved):
This looks awesome! I have a few comments, but it's mostly trivial stuff.

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+
+First, move the query into a separate file called ``getMovie.edgeql``. Next,
+we'll run the ``queries`` generator:
+
+.. code-block:: bash
 
-The EdgeDB query builder provides a **code-first** way to write
-**fully-typed** EdgeQL queries with TypeScript. We recommend it for TypeScript
-users and JavaScript users who prefer writing queries as code.
+  $ npx @edgedb/generate queries --file
```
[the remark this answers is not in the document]
scotttrinh: The current implementation is actually broken (see #622 ), but furthermore this exports all potential queries to a centralized file such that you import all of the generated queries from the same file, similar to the interfaces and query builder generators.

I don't think it's worth really talking through this _too_ much since I [plan on changing the default to be the current `--file` behavior](https://github.com/edgedb/edgedb-js/issues/540#issuecomment-1545952603) in the very near future, so this will just go away soon.

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+  $ npx tsx query.ts
 
-The query builder
-=================
+Queries generator
+^^^^^^^^^^^^^^^^^
+
+Wouldn't it be great if we could write any arbitrary query and get a type-safe
+function that we could call? Good news, that's exactly what the next generator
+does! The :ref:`queries generator <edgedb-js-queries>` Scans your project for
```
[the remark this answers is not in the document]
scotttrinh: Yeah, I think the names are sloppy in general:

1. Query Builder -> `generate edgeql-js`
2. Query Files -> `generate queries`
3. Interfaces -> `generate interfaces`

I think I like "Queries Generator" better than "Query Generator" since it generates zero or more "queries" instead of one "query".

So maybe the pages should be:

Query Builder Generator
Queries Generator
Interfaces Generator

WDYT?

docs/querybuilder.rst
```diff
@@ -383,7 +225,7 @@ Nested shapes
   }));
 
   const result = await query.run(client);
-  // Array<{id: string; title: string, actors: Array<{name: string}>}>
+  // { id: string; title: string; actors: { name: string }[] }[]
```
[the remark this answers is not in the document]
scotttrinh: The format I'm using here is what you would see from the language server, so I'd rather keep that consistent:

image

I personally agree that `Array<T>` is easier to read, but this is the conventional/idiomatic way of writing these types. Maybe splitting it on multiple lines is easier to read?

```suggestion
  /*
    {
      id: string;
      title: string;
      actors: { name: string }[]
    }[]
  */
```

docs/index.rst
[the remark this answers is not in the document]
raddevon: Perfect!

docs/index.rst
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+
+.. note:: A word on module systems
+
+  Different build tools and runtimes have different specifications for how
+  modules are imported, and we support a wide-range of those styles. For
+  clarity, we will be sticking to standard TypeScript-style ESM module importing
+  without a file extension throughout this documentation. Please see your build
+  or environment tooling's guidance on how to adapt this style.
+
+Query
```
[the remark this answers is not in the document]
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+    // result will be inferred as Movie | null
+    const result = await getMovie(client);
+
+    console.log(JSON.stringify(result, null, 2));
+  }
+
+  main();
+
+Now, if you change the query to return different data, or take parameters, the
+type of the generated function will change. It'll be completely type safe!
```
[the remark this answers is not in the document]
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
-How do I get started?
----------------------
+What's next
+===========
 
-We recommend reading the :ref:`Client docs <edgedb-js-driver>` first. If you
-are happy writing your EdgeQL as plain strings, then that's all you need! If
-you're a TypeScript user, or simply prefer writing queries in a code-first
-way, continue on to the :ref:`Query builder <edgedb-js-qb>` docs.
+We recommend reading the :ref:`Client docs <edgedb-js-driver>` first and getting
```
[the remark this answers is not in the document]
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+familiar with configuring the client. You'll find important APIs like
+``withGlobals`` and connection details there. After that, depending on your
+preferences, look through the :ref:`Query Builder <edgedb-js-qb>` documentation
```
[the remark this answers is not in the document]
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

docs/querybuilder.rst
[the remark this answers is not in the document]
scotttrinh: ⚡ [`5b58c22` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/5b58c223b934794dd63cb329d8135b1f4dea9363)

scotttrinh:
Alright, @raddevon I'd love one more look. I made some pretty big changes to the "Generators" page and added back some transaction-specific use to the query builder docs (would love to make a separate page for this soon!)

docs/index.rst
```diff
@@ -27,123 +27,316 @@ EdgeDB TypeScript/JS Client
...
+
+  main();
+
+We can now seed the database by running this script with ``tsx``
+
+.. code-block:: bash
+
+  $ npx tsx seed.ts
+
+Feel free to explore the database in the EdgeDB UI, where you will find the new
```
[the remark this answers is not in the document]
scotttrinh: ⚡ [`505add5` (#619)](https://github.com/edgedb/edgedb-js/pull/619/commits/505add59fee1d81d9ff44045c27f9f6a6267ea6e)

1st1 (approved):

docs/generation.rst
```diff
@@ -3,122 +3,91 @@
...
+    :caption: Deno
 
-**Deno**
+    $ deno run \
+      --allow-all \
+      --unstable \
+      https://deno.land/x/edgedb/generate.ts <generator> [options]
 
-.. code-block:: bash
+The value of generators should be one of the following:
```
raddevon, line 42:
```suggestion
The value of ``<generator>`` should be one of the following:
```

docs/generation.rst
```diff
@@ -3,122 +3,91 @@
...
 
-  $ deno run --allow-all --unstable https://deno.land/x/edgedb/generate.ts <generator> [options]
+.. list-table::
+   :class: funcoptable
 
+   * - ``edgeql-js``
+     - Provides a **code-first** way to write **fully-typed** EdgeQL queries
+       with TypeScript. We recommend it for TypeScript users, or anyone who
```
raddevon, line 49: I would also be explicit that this is generating the query builder. The other two seem obvious to me since the names match up, but a fresh user might otherwise not connect this one to the query builder.

Something like "Generates the query builder which provides a code-first…"

docs/generation.rst
```diff
@@ -3,122 +3,91 @@
...
+       interfaces* that correspond to each object type. This is useful for
+       writing typesafe code to interact with EdgeDB.
+     - :ref:`docs <edgedb-js-interfaces>`
 
 Connection
 ^^^^^^^^^^
 
-Generating the query builder requires a
-connection to an active EdgeDB database. It does **not** simply read your local ``.esdl`` schema files. Generators rely on the database to introspect the schema and analyze queries. Doing so without a database connection would require implementing a full EdgeQL parser and static analyzer in JavaScript—which we don't intend to do anytime soon.
+Generating the query builder requires a connection to an active EdgeDB database.
```
raddevon, line 67: Is this true only for the query builder or for all generators?

docs/generation.rst
```diff
@@ -3,122 +3,91 @@
...
+analyzer in JavaScript—which we don't intend to do anytime soon.
 
 .. note::
 
-  Make sure your development database is up-to-date with your latest schema before running a generator!
+  Make sure your development database is up-to-date with your latest schema
+  before running a generator!
 
-If you're using ``edgedb project init``, this is automatically handled for you. Otherwise, you'll need to explicitly pass connection information via environment variables or CLI flags, just like any other CLI command. See :ref:`Client Libraries > Connection <edgedb_client_connection>` for guidance.
+If you're using ``edgedb project init``, this is automatically handled for you.
```
raddevon, line 78: Be explicit about what "this" refers to here. I think you're referring to "connecting to your instance"?

raddevon (approved):
Awesome! A few more, again small, comments, but it looks great!

docs/generation.rst
scotttrinh: True for all: we introspect the database in all three generators.

PR #620, opened 2023-05-30
improve error handling with HTTP-SCRAM auth

freeatnet:
This PR improves error handling in HTTP-SCRAM auth module by rigorously checking that the responses provide all the expected data and exposing the underlying server responses when those checks fail.

### DIFF
--- a/packages/driver/src/httpScram.ts
+++ b/packages/driver/src/httpScram.ts
@@ -16,39 +16,6 @@ import {
 
 const AUTH_ENDPOINT = "/auth/token";
 
-function utf8ToB64(str: string): string {
-  return encodeB64(utf8Encoder.encode(str));
-}
-
-function b64ToUtf8(str: string): string {
-  return utf8Decoder.decode(decodeB64(str));
-}
-
-/**
- * Parses SCRAM-SHA-256 authentication parameters from a string.
- * @param authParamsString a string of comma-separated key-value pairs
- * @returns an object with `sid` and `data` properties, decoded string or `null` when not present
- */
-function parseSidAndData(paramsStr: string): {
-  sid: string | null;
-  data: string | null;
-} {
-  const params = new Map(
-    paramsStr.length > 0
-      ? paramsStr
-          .split(",")
-          .map((attr) => attr.split(/=(.+)?/, 2)) // split on first '=' only; nb, `.split("=", 2)` doesn't do that
-          .map(([key, val]) => [key.trim(), val.trim()])
-      : []
-  );
-
-  const sid = params.get("sid") ?? null;
-  const rawData = params.get("data");
-  const data = rawData ? b64ToUtf8(rawData) : null;
-
-  return { sid, data };
-}
-
 export async function HTTPSCRAMAuth(
   baseUrl: string,
   username: string,
@@ -94,7 +61,7 @@ export async function HTTPSCRAMAuth(
     throw new ProtocolError(`authentication failed: ${body}`);
   }
 
-  const { sid, data: serverFirst } = parseSidAndData(authParams);
+  const { sid, data: serverFirst } = parseScramAttrs(authParams);
   if (!sid || !serverFirst) {
     throw new ProtocolError(
       `authentication challenge missing '${!sid ? "sid" : "data"}' attribute`
@@ -126,7 +93,7 @@ export async function HTTPSCRAMAuth(
     throw new ProtocolError(`authentication failed: ${body}`);
   }
 
-  const { data: serverFinal, sid: sidFinal } = parseSidAndData(authInfoHeader);
+  const { data: serverFinal, sid: sidFinal } = parseScramAttrs(authInfoHeader);
   if (!sidFinal || !serverFinal) {
     throw new ProtocolError(
       `authentication info missing '${!sidFinal ? "sid" : "data"}' attribute`
@@ -144,4 +111,32 @@ export async function HTTPSCRAMAuth(
 
   const authToken = await serverFinalRes.text();
   return authToken;
+}
+
+function utf8ToB64(str: string): string {
+  return encodeB64(utf8Encoder.encode(str));
+}
+
+function b64ToUtf8(str: string): string {
+  return utf8Decoder.decode(decodeB64(str));
+}
+
+function parseScramAttrs(paramsStr: string): {
+  sid: string | null;
+  data: string | null;
+} {
+  const params = new Map(
+    paramsStr.length > 0
+      ? paramsStr
+          .split(",")
+          .map((attr) => attr.split(/=(.+)?/, 2)) // split on first '=' only; nb, `.split("=", 2)` doesn't do that
+          .map(([key, val]) => [key.trim(), val.trim()])
+      : []
+  );
+
+  const sid = params.get("sid") ?? null;
+  const rawData = params.get("data");
+  const data = rawData ? b64ToUtf8(rawData) : null;
+
+  return { sid, data };
 }
--- a/packages/driver/src/httpScram.ts
+++ b/packages/driver/src/httpScram.ts
@@ -24,6 +24,31 @@ function b64ToUtf8(str: string): string {
   return utf8Decoder.decode(decodeB64(str));
 }
 
+/**
+ * Parses SCRAM-SHA-256 authentication parameters from a string.
+ * @param authParamsString a string of comma-separated key-value pairs
+ * @returns an object with `sid` and `data` properties, decoded string or `null` when not present
+ */
+function parseSidAndData(paramsStr: string): {
+  sid: string | null;
+  data: string | null;
+} {
+  const params = new Map(
+    paramsStr.length > 0
+      ? paramsStr
+          .split(",")
+          .map((attr) => attr.split(/=(.+)?/, 2)) // split on first '=' only; nb, `.split("=", 2)` doesn't do that
+          .map(([key, val]) => [key.trim(), val.trim()])
+      : []
+  );
+
+  const sid = params.get("sid") ?? null;
+  const rawData = params.get("data");
+  const data = rawData ? b64ToUtf8(rawData) : null;
+
+  return { sid, data };
+}
+
 export async function HTTPSCRAMAuth(
   baseUrl: string,
   username: string,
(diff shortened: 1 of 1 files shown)


packages/driver/src/httpScram.ts
```diff
@@ -24,6 +24,31 @@ function b64ToUtf8(str: string): string {
   return utf8Decoder.decode(decodeB64(str));
 }
 
+/**
+ * Parses SCRAM-SHA-256 authentication parameters from a string.
+ * @param authParamsString a string of comma-separated key-value pairs
+ * @returns an object with `sid` and `data` properties, decoded string or `null` when not present
+ */
```
scotttrinh, line 31: Couple of small nits:

1. No need to add comments and JSDocs to an internal function.
2. Our style is to put these internal functions after the exported functions since they play more of a supporting/abstracting role.

packages/driver/src/httpScram.ts
```diff
@@ -24,6 +24,31 @@ function b64ToUtf8(str: string): string {
...
+function parseSidAndData(paramsStr: string): {
+  sid: string | null;
+  data: string | null;
+} {
+  const params = new Map(
+    paramsStr.length > 0
+      ? paramsStr
+          .split(",")
+          .map((attr) => attr.split(/=(.+)?/, 2)) // split on first '=' only; nb, `.split("=", 2)` doesn't do that
```
scotttrinh, line 40: It would be good to add a test, if reasonable, to the suite that exercises when this doesn't work especially since this changes the existing behavior. Wouldn't want to regress on this if indeed we need the `[first, ...rest]` kind of behavior here.

packages/driver/src/httpScram.ts
```diff
@@ -41,27 +66,42 @@ export async function HTTPSCRAMAuth(
...
-    console.log(serverFirstRes);
-    throw new Error(`Server doesn't support HTTP SCRAM authentication`);
+
+  // The first request must have status 401 Unauthorized and provide a
+  // WWW-Authenticate header with a SCRAM-SHA-256 challenge.
+  // See: https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L153-L157
+  const authenticateHeader = serverFirstRes.headers.get("WWW-Authenticate");
+  if (serverFirstRes.status !== 401 || !authenticateHeader) {
+    const body = await serverFirstRes.text();
+    throw new ProtocolError(`authentication failed: ${body}`);
```
scotttrinh, line 76: I'm not sure I think the body text is the only (or indeed the most helpful) thing we can log here. And in fact, I'm concerned about leaking data to logs or having to even parse the body in these error cases.

I think it'd be helpful to think about some structured way to represent the failed response that has a relatively static size, or at least a reasonably finite size. Something along these lines:

```suggestion
    const headerKeys = Array.from(serverFirstRes.headers().keys());
    const status = serverFirstRes.status;
    throw new ProtocolError(`Authentication failed with status ${status}. Headers: ${headerKeys}`);
```

What do you think?

packages/driver/src/httpScram.ts
```diff
@@ -41,27 +66,42 @@ export async function HTTPSCRAMAuth(
...
   }
-  if (!firstAttrs.has("sid") || !firstAttrs.has("data")) {
+
+  // The server may respond with a 401 Unauthorized and `WWW-Authenticate: SCRAM-SHA-256` with
+  // no parameters if authentication fails, e.g., due to an incorrect username.
+  // See: https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L112-L120
+  const authParams = authenticateHeader.split(/ (.+)?/, 2)[1] ?? "";
+  if (authParams.length === 0) {
+    const body = await serverFirstRes.text();
+    throw new ProtocolError(`authentication failed: ${body}`);
```
scotttrinh, line 94: I think in this case adding the `authenticateHeader` text is probably the most helpful thing here instead of (or in addition to?) the body text. Maybe also log the header keys?

packages/driver/src/httpScram.ts
```diff
@@ -41,27 +66,42 @@ export async function HTTPSCRAMAuth(
...
+  }
+
+  const { sid, data: serverFirst } = parseSidAndData(authParams);
+  if (!sid || !serverFirst) {
     throw new ProtocolError(
-      `server response doesn't contain '${
-        !firstAttrs.has("sid") ? "sid" : "data"
-      }' attribute`
+      `authentication challenge missing '${!sid ? "sid" : "data"}' attribute`
```
scotttrinh, line 100: Falsy is a reasonable check in the `if` conditional, but I think it might be helpful to log the stringified value of both attributes here since `null` vs `""` might be helpful when debugging.

packages/driver/src/httpScram.ts
```diff
@@ -41,27 +66,42 @@ export async function HTTPSCRAMAuth(
...
+  if (!authenticateHeader.startsWith("SCRAM-SHA-256")) {
+    throw new ProtocolError(
+      `unsupported authentication scheme: ${authenticateHeader}`
+    );
   }
-  if (!firstAttrs.has("sid") || !firstAttrs.has("data")) {
+
+  // The server may respond with a 401 Unauthorized and `WWW-Authenticate: SCRAM-SHA-256` with
+  // no parameters if authentication fails, e.g., due to an incorrect username.
+  // See: https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L112-L120
```
scotttrinh, line 90: Love these helpful comments! I'll probably update them to the relevant parts of [RFC 7804](https://datatracker.ietf.org/doc/html/rfc7804#section-5) since they're less likely to line-number sensitive and we'd like to think of the client and server being decoupled at least the protocol level here.

packages/driver/src/httpScram.ts
```diff
@@ -24,6 +24,31 @@ function b64ToUtf8(str: string): string {
   return utf8Decoder.decode(decodeB64(str));
 }
 
+/**
+ * Parses SCRAM-SHA-256 authentication parameters from a string.
+ * @param authParamsString a string of comma-separated key-value pairs
+ * @returns an object with `sid` and `data` properties, decoded string or `null` when not present
+ */
+function parseSidAndData(paramsStr: string): {
```
scotttrinh, line 32: Non-blocking naming thought: I wonder if `parseScramAttrs` is maybe a bit more descriptive since it gives you so context as to _what_ `sid` and `data` are?

scotttrinh (commented):
A couple of small questions, and suggestions, but I love the direction here. Thanks for taking the time to add extremely useful comments and fixing the `split` related issues here.

packages/driver/src/httpScram.ts
freeatnet: As a user of EdgeDB, I'd prefer it if the errors were specific to the conditions that caused them. Unfortunately, edgedb server doesn't return enough information in the headers to determine the cause of the error, so we need to look at the response body. This is especially true in the second request, where three different issues: a bad session ID (which could be a client error or the result of server restart/request being routed to a different instance), a session timeout (client error), and invalid credentials (client misconfiguration) — all produce the same status/headers combo. In all cases though, edgedb-generated errors are succinct and don't include any sensitive data, so I think it's reasonably safe to include them into the error message.

However, I agree that there's a risk of getting an unwieldy response body here from some intermediate HTTP gateway spewing an HTML error page. We can guard against that by truncating the error message. Additionally, we could check the content type before reading the response body — it looks like edgedb returns `Content-Type: text/plain` in error responses.

packages/driver/src/httpScram.ts
freeatnet: I was thinking of doing the same! Unfortunately, I couldn't find an example of this specific behaviour in the RFC — did I miss it?

packages/driver/src/httpScram.ts
freeatnet: I believe the only value `authenticateHeader` could have here is `SCRAM-SHA-256` (for values starting with anything but `SCRAM-SHA-256` we would've thrown on [line 84](https://github.com/edgedb/edgedb-js/pull/620/files/d3ade318144f83e73d010a83f2c3cbe9f9ee9ec0#diff-58b536d06e962df6fbe8c6ddaffd1a334bd0001566bc4a5f606b11e99e2dba2cR84), and the rest of the header was non-empty, we wouldn't be in this branch). However, the error message may either be [`"Client selected an invalid SASL authentication mechanism"`](https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L39-L45) or [`"Authentication failed"`](https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L115-L120) here.

packages/driver/src/httpScram.ts
scotttrinh: Yeah, I think this just falls under the server returning a response that doesn't conform to the standard and thus should be rejected rather than something explicitly spelled out in the spec itself.

I think this is fine in lieu of having a more specified thing to point at!

packages/driver/src/httpScram.ts
scotttrinh: > However, I agree that there's a risk of getting an unwieldy response body here from some intermediate HTTP gateway spewing an HTML error page. We can guard against that by truncating the error message. Additionally, we could check the content type before reading the response body — it looks like edgedb returns Content-Type: text/plain in error responses.

Yeah, I don't want to get too caught up in guarding this since we're replacing a literal `console.log` here (🙈 ), but it would be good to avoid buffering the whole response to protect against some DDoS/accidental bug where the streaming response trickles in or it spews gigs of text down the stream. I haven't used `ReadableStream` yet, but I think something like this could work (export from the `utils.ts` module?):

```ts
export async function streamToFixedBuffer(
  size: number,
  stream: ReadableStream
): Promise<Buffer> {
  const buf = Buffer.alloc(size);
  let bytesRead = 0;

  for await (const chunk of stream) {
    const remainingSpace = size - bytesRead;
    buf.set(chunk.slice(0, remainingSpace), bytesRead);
    bytesRead += remainingSpace;

    if (bytesRead === size) {
      break;
    }
  }

  return buf;
}
```

[CodeSandbox](https://codesandbox.io/p/sandbox/naughty-morse-7x9eop?file=%2Findex.mjs%3A19%2C21)

Then used like:

```suggestion
    const body = await streamToFixedBuffer(1024, serverFirstRes.body);
    throw new ProtocolError(`authentication failed: ${body}`);
```

---

Apologies for falling into this rabbit hole, this is something I want to improve in general and it's helpful to have someone to bounce ideas off of here.

freeatnet (commented):
Oops. Forgot to post these.

packages/driver/src/httpScram.ts
scotttrinh: Thanks for the detailed explanation and investigation. I think what you have here is clearly correct and I'd really rather keep the change you made and harden it by testing it, but I realize this isn't straightforward to test from the outside of this module. This PR conversation can be our documentation about the issue until such a time as we add some tests that actually exercise this logic.

🙏

packages/driver/src/httpScram.ts
scotttrinh: Let's ship what we have here. I want to improve tracing/logging in general so we can do something more fundamentally helpful and general at the point to mitigate any issues then.

scotttrinh: FWIW, I was able to come up with a fairly compact version of `streamToFixedBuffer` using a `reader` (and fix a bug 🙈 ):

```ts
export async function streamToFixedBuffer(
  size: number,
  stream: ReadableStream
): Promise<Buffer> {
  const reader = stream.getReader();
  const buf = Buffer.alloc(size);
  let bytesRead = 0;

  while (bytesRead < size) {
    const { done, value } = await reader.read();
    if (done) {
      return buf;
    }

    const remainingSpace = size - bytesRead;
    const byteSlice = value.slice(0, remainingSpace);
    buf.set(byteSlice, bytesRead);
    bytesRead += byteSlice.length;

    if (bytesRead === size) {
      return buf;
    }
  }
}
```

Neustradamus:
Good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #631, opened 2023-06-06
Benchmark query builder language server performance

scotttrinh:
Starting point for being able to benchmark multiple versions of the query builder against multiple versions of TypeScript's language service binary.


scotttrinh:
Closing in favor of #829

Issue #599, opened 2023-05-11
Fix deno release flow and publish two packages


scotttrinh:
We are currently not automatically publishing releases to deno.land during our release. We're also only publishing a single package that represents the driver, so we should also publish the generator package.

jaclarke:
@scotttrinh Just in case this is useful:

The rough plan I had for splitting edgedb-deno into two packages was to re-organise the repo into two root directories: `driver` and `generate`, remove the current webhook (https://github.com/edgedb/edgedb-deno/settings/hooks), and replace that with a github action that would similarly be triggered on tag creation and mimic github's webhook call, but would decide which webhook (`https://api.deno.land/webhook/gh/edgedb?subdir=driver` or `https://api.deno.land/webhook/gh/edgedb_generate?subdir=generate`) to call based on the tag prefix.

Since it's very possible that this won't work for some reason, the fallback idea was to just create another repo for the edgedb-deno generate package, and use the normal webhook. But I felt like trying to keep it to one repo first was less stuff to duplicate and maintain.

scotttrinh:
@jaclarke 

> that would similarly be triggered on tag creation and mimic github's webhook call, but would decide which webhook

This is a clever solution! I'm definitely worried that since this isn't documented, that they'll have some kind of shared token to ensure that the call is coming from GitHub that would make it harder to publish. I'll do some more digging here.

My current plan is to investigate having a single version that all packages share which will solve the short-term issue with the tagging structure and ambiguity in Deno. It's worth understanding the deno.land publishing flow a little better anyway, though, so I'll see what I can find about this.

denizdogan:
While waiting for this to be changed, is there anyone I could bribe to publish v1.2.2 to deno.land?

scotttrinh:
@denizdogan 

I'm going to try to manually publish a release!

Out of curiosity is it possible to use github directly instead of deno.land? I imagined one of the advantages of "use fully qualitified URLs as packages" was that one isn't tied to a package repository and could import directly from any public URL? Mostly, I'm just wondering how "stuck" deno users are as we're working out the publishing story here.

scotttrinh:
@denizdogan 

> If I may ask, what are the customizations that your Deno builder does?

https://github.com/edgedb/edgedb-js/blob/master/compileForDeno.ts

> Could some of it be replaced with just [Denoify](https://docs.denoify.land/)? 

[...]

PR #632, opened 2023-06-08
WIP: Improve and simplify Deno compatibility

denizdogan:
**Work In Progress** – I'm aiming to remove _all_ "custom" Deno build logic here in favor of Denoify.

[x] packages/driver
[ ] packages/generate
[ ] ???

- I've taken some liberties with things like changing file naming conventions in favor of Denoify's default pattern, where `foobar.deno.ts` is the Deno replacement for `foobar.ts`. Previously, it's been `foobar.node.ts` and `foobar.deno.ts` explicitly.
- A bunch of minor cosmetic changes to appease the Deno linter.
- Ended up removing `driver/.../globals.deno.ts` because I couldn't figure out what it was for...?
- Explicitly import Node.js builtins, e.g. `import { Buffer } from "buffer"` because Denoify needs it to do its thing. Ideally, I'd want to use `node:buffer` style imports, Denoify actually turns that into `node:node:buffer`, so I'm holding off on that.


denizdogan:
Started doing this after discussion in #599, please let me know what you think so far, especially if it looks like I'm unknowingly about to break compatibility with something important.

denizdogan:
I should add that the reason why I've added `deno_dist` is that it seems like it's kind of required in order to publish this to deno.land? If there's no simple way around it, that kind of sucks, but I'm sure we can figure it out somehow.

PR #638, opened 2023-06-09
Drop the RingBuffer

1st1:
* it's a micro-optimization
* it causes our users pain
* it's unclear what should be the capacity of the ring buffer so it's easier to just replace it with a regular Array


scotttrinh (approved):
Cool. An alternative is to make it like configurable or some much larger size, right? `Array` seems fine and typical of what I would expect from a database driver, tbh.

1st1:
> Cool. An alternative is to make it like configurable or some much larger size, right? 

Yes, but that would also be wasteful use of resources. It used to be visible in benchmarks, now I very much doubt so (JITs are faster than they were 5 years ago)

scotttrinh:
@1st1 Some tests failing now, I'll see if I can troubleshoot on Monday and get this ready for a release.

PR #640, opened 2023-06-10
Lowercase org/instance name when computing Cloud instance DNS

elprans:
Nebula switched to doing this, and so the clients must follow.


nsidnev (approved):

scotttrinh (approved):

scotttrinh:
I'll get the tests passing as well.

elprans:
> I'll get the tests passing as well.

Failures seem like flakes.

scotttrinh:
> Failures seem like flakes.

There was one failure earlier before the force push that seemed significant, but looks like you already caught it.

PR #648, opened 2023-06-13
Support `ext::pgvector::vector`

scotttrinh:
Mostly copies the equivalent PRs in `edgedb/edgedb`:
- https://github.com/edgedb/edgedb-python/pull/439
- https://github.com/edgedb/edgedb-python/pull/440


packages/driver/test/client.test.ts
```diff
@@ -397,6 +399,113 @@ test("fetch: int64 as bigint", async () => {
   }
 });
 
+if (!isDeno) {
```
scotttrinh, line 402: To be continued, Deno!

scotttrinh: Note to future self: the tests are actually passing in the Deno test runner, but the "describe" block itself is being reported as failing.

```
 FAIL  test/deno.test.ts (120.209 s)
  ● Console

    console.log
      running 36 tests from ./test/client.test.ts
      query: basic scalars ... ok (125ms)
      fetch: bigint ... ok (52ms)
      fetch: decimal as string ... ok (59ms)
      fetch: int64 as bigint ... ok (42ms)
      fetch: ext::pgvector::vector ...
        valid: Float32Array ... ok (196ms)
        valid: JSON ... ok (168ms)
        invalid: empty ... ok (6ms)
        invalid: invalid argument ... ok (5ms)
      fetch: ext::pgvector::vector ... FAILED (2s)
```

PR #653, opened 2023-06-16
Fix pgvector support in @edgedb/generate

jaclarke:


packages/generate/test/pgvector.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        }[]
+      >
+    >(true);
+  });
+
+  test("params", async () => {
+    const query = e.params({ vec: e.ext.pgvector.vector }, ($) => e.select($));
+
+    const args = {
+      vec: Float32Array.from(
```
scotttrinh, line 53: Should we also test `number[]`? Maybe test that selecting either is deep equal?

packages/generate/test/testRunner.ts
```diff
@@ -26,11 +29,37 @@ import {
...
+    });
+    console.log(`\nRunning tests...`);
+    await runCommand(
+      "yarn",
+      [
+        "test:ts",
+        ...(version.major < 3
+          ? [
+              `--testPathIgnorePatterns="${[
+                "pgvector",
```
scotttrinh, line 54: Love this! We might even be able to throw flags into the test name to get this less specific, but this works great.

scotttrinh (approved):
Great! No blockers.

packages/generate/test/pgvector.test.ts
jaclarke: Yes, added

PR #657, opened 2023-06-18
fix(generate): write the generated query file to the source's path

### LINKED ISSUE #656: Queries generator writes files to project root
**Describe the bug**

`@edgedb/generate queries` outputs files in the project root folder. This can cause clashes or, for TypeScript projects, be outside the compiled files directory.

**Reproduction**

1. Set up a project with edgedb and `@edgedb/generate`, add a deeply nested query, e.g., at `src/server/auth/adapters/edgedb/queries/createUser.edgeql`.
2. Run `npx @edgedb/generate queries` from the project root.
3. Observe the command running successfully:

```
$ npx @edgedb/generate queries
Generating functions from .edgeql files...
Detected tsconfig.json, generating TypeScript files.
   To override this, use the --target flag.
   Run `npx @edgedb/generate --help` for full options.
Detected project root via edgedb.toml:
   /path/to/project/t3-template-with-edge
Connecting to database...
Analyzing .edgeql files...
Generating files for following queries:
   ./createUser.query.ts
```

**Expected behavior**

`createUser.query.ts` is placed in the same directory as the source query file, i.e.,`src/server/auth/adapters/edgedb/queries/`.

**Observed behavior**

`createUser.query.ts` is placed at the EdgeDB project root path.

- OS: macOS
- EdgeDB version (e.g. `2.0`): 2.16+183b29a
- EdgeDB CLI version (e.g. `2.0`): EdgeDB CLI 3.1.1+5798f31
- `edgedb-js` version (e.g. `0.20.10`;): "edgedb": "1.3.0", "@edgedb/generate": "0.2.0"
- Node/Deno version: 18.16

**Additional context**

Looks like changes in https://github.com/edgedb/edgedb-js/pull/622/files#diff-d6282c6c02acf0ae7f8578cc1ba9d0a376a489b8ea7dafa007c2353f2e596b06 broke path passing. The following should work:

```diff
  const queryFileName = adapter.path.basename(params.path);
  const baseFileName = queryFileName.replace(/\.edgeql$/, "");
-  const outputBaseFileName = `${baseFileName}.query`;
+  const outputDirname = adapter.path.dirname(params.path);
+  const outputBaseFileName = adapter.path.join(outputDirname, `${baseFileName}.query`);
```

freeatnet:
Fixes #656.

### DIFF
--- a/packages/generate/src/queries.ts
+++ b/packages/generate/src/queries.ts
@@ -198,7 +198,11 @@ function generateFiles(params: {
 }[] {
   const queryFileName = adapter.path.basename(params.path);
   const baseFileName = queryFileName.replace(/\.edgeql$/, "");
-  const outputBaseFileName = `${baseFileName}.query`;
+  const outputDirname = adapter.path.dirname(params.path);
+  const outputBaseFileName = adapter.path.join(
+    outputDirname,
+    `${baseFileName}.query`
+  );
 
   const method =
     params.types.cardinality === Cardinality.ONE


scotttrinh (approved):
Nice catch! I swore I tested this but perhaps I broke it doing some refactoring? Thanks for the quick fix.

scotttrinh:
I'll get it formatted and merged on Monday and do a release.

freeatnet:
@scotttrinh formatting fixed!

As a side-note: is there a chance to get prettier [integrated into eslint](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) for this repo?

scotttrinh:
> As a side-note: is there a chance to get prettier [integrated into eslint](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) for this repo?

I have [the basic `extends` setup](https://github.com/edgedb/edgedb-js/blob/a05d11c7e13bd6659eff702310558ba8a7f65e11/.eslintrc.cjs#L6) but haven't setup the other config, thanks for the tip, I'll get that setup, too.

scotttrinh:
@freeatnet 

If you don't mind, I'd love your help bikeshedding the eslint setup a bit: Looking at the docs, it seems like the recommended configuration that you linked to ends up turning on the `prettier/prettier` rule which runs prettier from within eslint. I'm a mild -1 on this for the following reasons, but I'd love to hear some feedback from someone who has used this setup and likes it:

1. ESLint rule violations are something that I want to fix, so if I see a little squiggly, I spend time investigating. If it's just a prettier violation, I feel like it's a little bit distracting since I—
2. Have my editor auto-format on save.
3. The runtime overhead of running prettier in eslint, but I'm not sure what kind of actual impact it has.

Some alternatives to having it integrated into eslint:

1. Precommit hook
2. Some kind of GitHub Action that commits formatting
3. Anything else I'm missing here?

---

I'm open to trying it for a bit to see, but my initial reaction to using it for just a few minutes is that it gets in the way more than it helps since my editor makes it seem like something is wrong as I'm typing only to discover that it's just a formatting issue.

freeatnet:
@scotttrinh Tbh, I don't have deep arguments for this setup. I spent a bit of time figuring out a resilient setup a couple years back — and I've found it's easy to get consistent results in a team of developers with (slightly) different environments using this setup. On VSCode, install [the official ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and enable `source.fixAll.eslint`. Done.

I previously used ran Prettier separately via its own IDE plugins. I found that it resulted in inconsistencies based on the IDE plugin version/config and conflicts with code-modifying ESLint rules (e.g., [`import`-sorting rules](https://github.com/freeatnet/t3-with-dx/blob/main/.eslintrc.cjs#L59-L85)). Running Prettier via ESLint solved both for me. The plugins may have improved since!

I find that I don't really look at squiggles until I've saved the file. With `fixAll`-on-save, everything that can be auto-resolved goes away, so the only squiggles remaining are those that can't be solved automatically.

And re: alternatives… I do usually have a pre-commit hook as the second-to-last line of defence (the last being the CI), but I feel like formatting and linting should happen mostly in real time. I've tried using a GitHub Action that formats code before, but it was a huge mess (you basically end up working together on the same branch).

scotttrinh:
> On VSCode, install [the official ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and enable source.fixAll.eslint. Done.

Awesome, yeah, I think this is the problem I want to avoid: I don't think we can or should automatically run eslint's `fix`. At least not right away while we're actively migrating to it, and configuring the rules that matter.

> I find that I don't really look at squiggles until I've saved the file.

Yeah, maybe my issue is that I should just configure my editor to only show errors after save to avoid this, but I find it breaks my personal flow to: save constantly since saving moves things around, and see errors while I'm trying to think through what I'm typing at the moment.

> I usually have a pre-commit hook as the second-to-last line of defense (the last being the CI), but I feel like formatting and linting should happen mostly in real-time. I've tried using a GitHub Action that formats code before, but it was a huge mess (you basically end up working together on the same branch).

Yeah, I sympathize with this: Zod, for instance, has a pretty [heavy pre-commit hook](https://github.com/colinhacks/zod/blob/master/.husky/pre-commit) that generate Deno and does a bunch of other things and it can be a little painful since I like a lot of smaller commits. I wonder if moving this into a `pre-push` commit is like the least gross version of this. Doesn't get in your way while you're working, and saves you from wasting cycles on hitting issues in CI.

How does adding a pre-push hook feel as a first step toward smoothing out the friction of formatting/linting issues?

freeatnet:
@scotttrinh I feel like pre-push is slightly too late. IMO, every (or nearly every) commit pushed should pass all CIs. What I meant by pre-commit vs real-time is that real-time IDE-based feedback is definitely preferable to me — I don't love the idea of committed code being different from code last seen by the developer. (But aside from that, yeah, building and testing feel too heavy for a pre-commit.)

Re: auto-fixing with ESLint: is there a particular reason you'd like to avoid it?

scotttrinh:
> IMO, every (or nearly every) commit pushed should pass all CIs.

Gotcha. Yeah, I often commit works in progress that would not pass CI, so maybe that's a fundamental difference here.

I think, ultimately, the requirements for tooling here are:
1. Do not waste CI time and review cycles on testing things that do not pass lint and format rules
2. Do not require that contributors install any specialized tools or have to know how to set up their editor specifically for this
3. Little to no out-of-band requirements to get tooling working.

In that spirit, I think pre-push check (as opposed to a pre-push fix) might be the lowest common denominator here, even if it maintains some of the existing friction. I, too, find it not ideal if the code being committed isn't the code that the contributor saw in their editor.

> auto-fixing with ESLint: is there a particular reason you'd like to avoid it?

Especially in this code base, there are rules that we want to specifically break that make the code more readable _in certain cases_, but that we might not want to totally disable. For instance, there is a rule to disallow trivial types that can be inferred from default assignments. In some cases, I think it's very preferable to keep the explicit type, so I'd rather not add a blanket fix when it would be better to add an ignore.

Once we've fully migrated to eslint, I might find myself sliding closer to auto-fixing, but linting rules are not quite as fire-and-forget as formatting in my experience, and we currently have a _lot_ of rules that I wouldn't want to just apply fixes for blindly without reviewing.

Another thing is that I'd prefer that contributors are submitting smaller patches and files full of formatting and linting fixes are a little bit harder to review. Formatting, fine, I trust prettier. But linting is something I would want to review more closely, and unless someone is submitting a fix specifically for linting, I'd rather minimize that.

---

Thanks for taking the time to bikeshed around this, I think I've found a more firm position based on our convo, even if maybe we haven't actual found consensus. 🙌

freeatnet:
@scotttrinh Sounds like a great place to start! Also, maybe putting these details and any helpful commands in the "Contributing" section of the readme would be helpful?

scotttrinh:
> maybe putting these details and any helpful commands in the "Contributing" section of the readme would be helpful?

Great idea! Will do.

PR #664, opened 2023-06-22
Avoid generate package migration in server setup

scotttrinh:
We want to control migrations of the generate project manually, so instead of initializing the project in the setup, we'll just create an instance and query that for our server version check.


scotttrinh:
Going to explore just exporting the server version in the `setup-edgedb` action itself to avoid having to do this.

scotttrinh:
Incorporated into #653

PR #667, opened 2023-06-23
Add example: filtered delete with `in` operator

scotttrinh:
I was pairing with @DijanaPavlovic on doing this exact thing, and it took us both a few minutes to wrap our heads around this so thought it was worth adding an example to the docs.

Added a test to make sure I didn't screw this up!

### DIFF
--- a/docs/querybuilder.rst
+++ b/docs/querybuilder.rst
@@ -447,7 +447,7 @@ Compose queries
 
 All query expressions are fully composable; this is one of the major
 differentiators between this query builder and a typical ORM. For instance, we
-can ```select`` an ``insert`` query in order to fetch properties of the object we
+can ``select`` an ``insert`` query in order to fetch properties of the object we
 just inserted.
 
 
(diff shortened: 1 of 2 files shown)


DijanaPavlovic:
Hey, @scotttrinh  this is great... I'm just not sure what is the intent in general throughout the docs, there is `delete` inside QueryBuilder and `Delete` like a separate page with only one simple example... @raddevon

DijanaPavlovic (approved):

scotttrinh:
> there is delete inside QueryBuilder and Delete like a separate page with only one simple example...

Yeah, I agree that it's confusing and could be cleaned up a lot. Maybe we move the "cheatsheet" examples into their respective top-level page? Seems like that's easier to reference then going to the bottom of the query builder docs.

Thoughts @raddevon

raddevon:
@scotttrinh @DijanaPavlovic I like the idea of moving the cheatsheet examples into their respective sections. I think our goal is to ultimately do something like that with all of the top-level cheatsheets (Guides > Cheatsheets) as well, so this would make the JS docs consistent with the rest once that effort is completed.

PR #669, opened 2023-06-26
Handle >2 union members in links

scotttrinh:
We do not need to generate a separate object type for a type that is a union of other types since we will have already generated the individual types and the use of that type within another type will be handled by the declaration of the consumer of the union.

TODO:
- [x] Add 2.x back to the testing matrix
- [x] Get 2.x passing
- [x] Test (and handle?) aliases that are unions
   - You cannot make an alias union type. There are ways for an alias to have a type that is a union (making it a computed set of each type), but the reflection query is likely the same, so I'll consider this fine for now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,8 +6,6 @@ on:
       - master
       - ci
   pull_request:
-    branches:
-      - master
 
 jobs:
   test:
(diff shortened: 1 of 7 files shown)


packages/generate/test/params.test.ts
```diff
@@ -274,7 +274,7 @@ SELECT (SELECT __param__test)`
       tuple: args,
     });
 
-    assert.deepEqual(Object.values(complexResult.tuple), Object.values(args));
+    assert.deepEqual(complexResult.tuple, args);
```
scotttrinh, line 277: This changed for no good reason, but our macos tests were failing with a test timeout until I did this update. 🫠

Issue #635, opened 2023-06-09
TSError: src/baseConn.ts:533:9 - error TS2454: Variable 'inTypeId' is used before being assigned.


denizdogan:
Running `yarn test` in `packages/driver` causes the error. It seems like an easy fix, but maybe I'm the only one running into it? I'm not sure why the code is even written in a way that allows this to happen, but at the same time the code is so old, maybe TS was different back then.

```
~/Code/edgedb/edgedb-js/packages/driver master*
❯ yarn test
yarn run v1.22.19
$ npx --node-options='--experimental-fetch' jest --detectOpenHandles
Error: Jest: Got error running globalSetup - /Users/deniz/Code/edgedb/edgedb-js/packages/driver/test/globalSetup.ts, reason: [TSError: src/baseConn.ts:533:9 - error TS2454: Variable 'inTypeId' is used before being assigned.

533     if (inTypeId == null || outTypeId == null) {
            ~~~~~~~~
src/baseConn.ts:533:29 - error TS2454: Variable 'outTypeId' is used before being assigned.

533     if (inTypeId == null || outTypeId == null) {
                                ~~~~~~~~~
src/baseConn.ts:619:9 - error TS2454: Variable 'cardinality' is used before being assigned.

619     if (cardinality == null || outCodec == null || inCodec == null) {
            ~~~~~~~~~~~]
    at runGlobalHook (/Users/deniz/Code/edgedb/edgedb-js/node_modules/@jest/core/build/runGlobalHook.js:117:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async runJest (/Users/deniz/Code/edgedb/edgedb-js/node_modules/@jest/core/build/runJest.js:327:5)
    at async _run10000 (/Users/deniz/Code/edgedb/edgedb-js/node_modules/@jest/core/build/cli/index.js:344:7)
    at async runCLI (/Users/deniz/Code/edgedb/edgedb-js/node_modules/@jest/core/build/cli/index.js:199:3)
    at async Object.run (/Users/deniz/Code/edgedb/edgedb-js/node_modules/jest-cli/build/run.js:124:37)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

scotttrinh:
This is working for me locally, and obviously this runs in CI, so I'm not sure why you're encountering that issue. Are you sure you're on a clean checkout and at `master` with no changes?

[...]

PR #670, opened 2023-06-27
Allow testing features of different versions

scotttrinh:
Restructures the generator integration tests as separate top-level yarn workspace projects. This allows:

- Each project to have a different schema
- Easily test new features that are only available on newer channels

These are the "channels" I've defined here:
- `legacy`: versions no longer supported in LTS
- `lts`: The current LTS version
- `stable`: The current "latest"/"stable" version
- `nightly`: The latest nightly release

Since 3.x is the first LTS release and I _do_ want to support 2.x at least through 4.x, I've mapped version 2 to LTS even though it isn't strictly LTS. We also decided to continue to support all legacy versions at least to ensure we're connecting and can do some basic query building.

Each channel will run the tests from earlier channels:

| Channel | legacy | lts | stable | nightly |
|---|---|---|---|---|
| legacy | ✅ | ❌ | ❌ | ❌ |
| lts | ❌ | ✅ | ❌ | ❌ |
| stable | ❌ | ✅ | ✅ | ❌ |
| nightly| ❌ | ✅ | ✅ | ✅ |

The intent here is that _most_ of the tests live in `integration-lts` and any new features that are not available in LTS are tested in either `integration-stable` or `integration-nightly` depending on where they're available. That means that there is a chance that stable and nightly might have 0 tests and potentially an empty schema file.

## Drawbacks

- CI time is much higher now: each integration test suite takes ~1.5 minutes even if there are no tests due to server startup-cleanup time and (to a lesser extent) running the generators.
- When new server versions are released, we will need to shuffle schema definitions and test cases "down", assuming there were any changes that affect edgedb-js.
- It might not be obvious what kind of tests should live in the `packages/generate/test` directory: this should be a smoke test and any kind of unit testing we might want to do on the generator code.


.github/workflows/tests.yml
```diff
@@ -99,14 +99,22 @@ jobs:
...
+
+      - name: Run query builder integration tests stable
+        if: ${{ matrix.edgedb-version == 'stable' || matrix.edgedb-version == 'nightly' }}
+        run: |
+          yarn workspace @edgedb/integration-stable test:ci
+
+      - name: Run query builder integration tests nightly
+        if: ${{ matrix.edgedb-version == 'nightly' }}
+        run: |
+          yarn workspace @edgedb/integration-nightly test:ci
```
scotttrinh, line 117: Here are the relevant changes to the workflow that sequence the tests based on which "channel" CI is running.

package.json
```diff
@@ -6,15 +6,15 @@
...
+    "@typescript-eslint/eslint-plugin": "^5.60.1",
+    "@typescript-eslint/parser": "^5.60.1",
+    "eslint": "^8.43.0",
     "eslint-config-prettier": "^8.8.0",
     "prettier": "^2.8.8",
     "tslint": "^6.1.3",
     "tslint-config-prettier": "^1.18.0",
     "tslint-plugin-prettier": "^2.3.0",
-    "typescript": "^5.0.4"
+    "typescript": "5.0.4"
```
scotttrinh, line 17: Did a little bit of dependency bumping since I was adding new packages to avoid pinning the new packages to old versions.

Ran into a couple of small issues that I'll sort out separately:

1. `@js-temporal/polyfill` v0.4.4 now gives slightly different results for one particular range test.
2. `typescript` v5.1.x changed some behavior w.r.t. to `undefined` and `void` which has the side effect shown in #635 

Both are totally fixable, but I didn't want to add noise to this PR.

packages/generate/dbschema/default.esdl
```diff
@@ -1,209 +1,3 @@
-using extension pgvector;
-
 module default {
 
-  scalar type Genre extending enum<"Horror", "Action", "RomCom", "Science Fiction", "Select">;
```
scotttrinh, line 5: Most of this file moved to `integration-lts` except for the pgvector related stuff which ended up in `integration-stable`.

packages/generate/package.json
```diff
@@ -21,20 +21,18 @@
...
-    "@types/jest": "^29.5.1",
-    "@types/node": "^20.1.7",
-    "conditional-type-checks": "^1.0.5",
-    "edgedb": "^1.1.0",
-    "esbuild": "^0.15.7",
-    "globby": "^13.1.2",
+    "@types/jest": "^29.5.2",
+    "@types/node": "^20.3.2",
+    "conditional-type-checks": "^1.0.6",
+    "edgedb": "^1.3.2",
```
scotttrinh, line 27: I couldn't actually find an explanation of what the version string for a workspace actually does, but I thought it made sense to bump this to latest 🤷

packages/generate/package.json
```diff
@@ -21,20 +21,18 @@
...
+    "esbuild": "^0.18.10",
+    "globby": "^13.2.0",
     "jest": "^29.5.0",
-    "superjson": "^1.7.5",
+    "superjson": "^1.12.4",
     "ts-jest": "^29.1.0",
-    "typescript": "^5.0"
-  },
-  "dependencies": {
-    "chokidar": "^3.5.3"
```
scotttrinh, line 36: We aren't actually using `chokidar` anywhere, and definitely not at runtime!

packages/integration-stable/testRunner.ts
```diff
@@ -0,0 +1,40 @@
+import createClient from "../driver/src/index.node";
```
scotttrinh, line 1: BTW, some of these test setup files can probably move into some shared place like `driver/test/testUtil`, but I wanted to err on the side of copy-paste for now. Open to doing it now if this feels too WET.

scotttrinh:
A quick note that we discussed adding some decent smoke test for 1.x and 2.x to make sure we're not loosing support as our release cadence picks up speed. Going to do that here, so taking this back to draft.

scotttrinh:
Leaving a comment about this for future self:

One idea that @jaclarke had to speed this up is to dynamically copy different schemas and tests into place at runtime based on the running version of the server instead of having separate projects. That is still worth exploring, but I'm going to consider that as a separate optimization to what exists here already.

packages/generate/package.json
scotttrinh: Yeah, I assume that's what we're trying to do since these are dev dependencies.

PR #671, opened 2023-06-30
Add subsections to the e.for bulk insert docs

scotttrinh:
Break down the bulk insert section for easier reading for both people and machines.

### DIFF
--- a/docs/for.rst
+++ b/docs/for.rst
@@ -188,8 +188,8 @@ you pass to the query's ``run`` method.
 
 .. _edgedb-js-for-bulk-inserts-conflicting-data:
 
-Inserting the conflicting data
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Inserting the inner conflicting data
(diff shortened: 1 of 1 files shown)


raddevon:
Beautiful! My only minor qualm: "parent data" sounds a bit off. I can get there, but I wonder if there's a better name for it.

- "outer data?"
- "containing data?"

I don't love those either to tell the truth. Maybe we just step around it by putting the headings in terms of the example: "inserting the movies" and "inserting the characters?"

scotttrinh:
I'm glad you also found that part objectionable!

> "inserting the movies" and "inserting the characters?"

Yeah, I thought about that, but then when thinking about how they would be linked by ChatGPT to explain potentially related problems, I felt like those headings didn't really explain what _part_ of this problem this was addressing. Maybe "outer data" is our best bet? 🤔

raddevon:
Good call on the heading vs. how ChatGPT would contextualize it. Yeah, I think "outer data" would be good.

raddevon (approved):

Issue #672, opened 2023-06-30
(Generate queries) Add ignore require-await rule for deno exports


PastelStoic:
Due to the output structure, the default deno linter complains about the generated queries not using await. Adding 
`// deno-lint-ignore-file require-await`
to each generated file would prevent this error.

scotttrinh:
Thanks for the report! PRs welcome, but I put it in my backlog if no one else gets to it in the next few weeks.

davepar:
Just a friendly ping that this is still an issue. The merged PR is adding a `deno-lint-ignore-file` for the file, but it's after an `import type` statement. This causes deno to ignore it. The line doesn't have to be at the top, but does need to be before any statements. See the [deno docs](https://docs.deno.com/runtime/manual/tools/linter#files) for details.

I can try this minor fix if nobody else is looking at it.

PastelStoic:
Now that I look more closely - is the `async` keyword actually needed? A promise is returned with or without it.

scotttrinh:
Fixed in #812

PR #675, opened 2023-07-03
Added linting rule for deno

PastelStoic:
Simple fix for #672. I'm slightly worried about the newline literal causing some weird pitfall with windows somehow.

### DIFF
--- a/packages/generate/src/queries.ts
+++ b/packages/generate/src/queries.ts
@@ -275,7 +275,7 @@ export function ${functionName}(client: Executor${
       return [
         {
           path: `${outputBaseFileName}.ts`,
-          contents: tsImpl,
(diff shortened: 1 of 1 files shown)


scotttrinh:
@PastelStoic 

Thanks for the suggestion here! Perhaps what we want is a `denoImpl` that just imports the `tsImpl` to make that assignment a little easier to grok. Do you want to take a swing at that? Otherwise, I can make the change myself when merging.

PastelStoic:
> @PastelStoic
> 
> Thanks for the suggestion here! Perhaps what we want is a `denoImpl` that just imports the `tsImpl` to make that assignment a little easier to grok. Do you want to take a swing at that? Otherwise, I can make the change myself when merging.

Done.

scotttrinh (approved):
🎉

PR #686, opened 2023-07-10
Add any existing `ext::*` scalar to a type's `__casttype__`

### LINKED ISSUE #685: `ext::pgvector` functions are not able to use scalars extend from `ext::pgvector::vector`
**Code**

```typescript
const q = e.select(e.Foo, (foo) => ({
  cosine_distance: e.ext.pgvector.cosine_distance(
    foo.embedding,
    target_embedding // Float32Array
  ),
});
```

**Schema**

```
module default {
  scalar type Embedding extending ext::pgvector::vector<1536>;

  type Foo {
    embedding: Embedding;
  }
}
```

scotttrinh:
Closes #685 

The type system resolves ancestor types to scalars calling them "material scalars" and we were not getting any `ext::*` scalars in our subquery that resolves these.

Some additional updates to make error messages a bit more useful since they currently print `[object Object]` instead of anything useful 😬

### DIFF
--- a/packages/driver/src/reflection/queries/types.ts
+++ b/packages/driver/src/reflection/queries/types.ts
@@ -127,7 +127,7 @@ export async function getTypes(
       material_scalars := (
         SELECT ScalarType
         FILTER
-          (.name LIKE 'std::%' OR .name LIKE 'cal::%')
+          (.name LIKE 'std::%' OR .name LIKE 'cal::%' OR .name LIKE 'ext::%')
           AND NOT .is_abstract
       )
 
--- a/packages/driver/src/reflection/queries/types.ts
+++ b/packages/driver/src/reflection/queries/types.ts
@@ -128,7 +128,9 @@ export async function getTypes(
         SELECT ScalarType
         FILTER
           (.name LIKE 'std::%' OR .name LIKE 'cal::%' OR .name LIKE 'ext::%')
-          AND NOT .is_abstract
+          AND NOT .abstract
+          AND NOT EXISTS .enum_values
+          AND NOT EXISTS (SELECT .ancestors FILTER NOT .abstract)
       )
 
     SELECT Type {
--- a/packages/driver/src/reflection/queries/types.ts
+++ b/packages/driver/src/reflection/queries/types.ts
@@ -126,11 +126,9 @@ export async function getTypes(
 
       material_scalars := (
         SELECT ScalarType
-        FILTER
-          (.name LIKE 'std::%' OR .name LIKE 'cal::%' OR .name LIKE 'ext::%')
-          AND NOT .abstract
-          AND NOT EXISTS .enum_values
-          AND NOT EXISTS (SELECT .ancestors FILTER NOT .abstract)
+        FILTER NOT .abstract
+           AND NOT EXISTS .enum_values
+           AND NOT EXISTS (SELECT .ancestors FILTER NOT .abstract)
       )
 
     SELECT Type {
--- a/packages/generate/src/syntax/funcops.ts
+++ b/packages/generate/src/syntax/funcops.ts
@@ -127,9 +127,15 @@ export function $resolveOverload(
   throw new Error(
     `No function overload found for ${
       funcName.includes("::")
-        ? `'e.${funcName.split("::")[1]}()'`
+        ? `'e.${funcName.split("::").join(".")}()'`
         : `operator '${funcName}'`
-    } with args: ${args.map((arg) => `${arg}`).join(", ")}`
+    } with args: ${[...positionalArgs, ...Object.values(namedArgs ?? {})]
+      .filter(Boolean)
+      .map(
+        (arg) =>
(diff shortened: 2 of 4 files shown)


packages/generate/src/syntax/funcops.ts
```diff
@@ -127,9 +127,15 @@ export function $resolveOverload(
   throw new Error(
     `No function overload found for ${
       funcName.includes("::")
-        ? `'e.${funcName.split("::")[1]}()'`
+        ? `'e.${funcName.split("::").join(".")}()'`
```
scotttrinh, line 130: Supports submodules including `ext::pgvector` in this case. It was printing `e.pgvector()` instead of `e.ext.pgvector.cosine_distance()`.

Note: we're not dropping `std` and `math` here any more 🤔

packages/integration-stable/setupTeardown.ts
```diff
@@ -15,7 +15,7 @@ export async function setupTests() {
 }
 
 async function cleanupData(client: Client) {
-  await client.execute(`reset schema to initial`);
+  await client.execute(`delete PgVectorTest`);
```
scotttrinh, line 18: Whoops! I erroneously thought `reset schema to initial` just deleted all the user defined objects, but it actually... resets the schema to the initial state before any migrations were run. 🤦 

Still need to figure out a query that will delete all user-defined types!

packages/driver/src/reflection/queries/types.ts
```diff
@@ -127,7 +127,7 @@ export async function getTypes(
       material_scalars := (
         SELECT ScalarType
         FILTER
-          (.name LIKE 'std::%' OR .name LIKE 'cal::%')
+          (.name LIKE 'std::%' OR .name LIKE 'cal::%' OR .name LIKE 'ext::%')
```
jaclarke, line 130: This works for now, but potentially a future extension may add non base-type scalars. So maybe the filter should check that the type has no non-abstract ancestors?

scotttrinh: > So maybe the filter should check that the type has no non-abstract ancestors?

Isn't that what `AND NOT .is_abstract` is doing on line 131? Or do you think there is some way to narrow the search space in `ext::` specifically?

packages/generate/src/syntax/funcops.ts
jaclarke: Good catch. I guess the previous logic was to strip out the `std` or `default` as they get aliased into the top level and we didn't have nested modules, but even that's broken because of the `cal` and `math` modules (and any non default user modules) 🤦.

scotttrinh: Yeah, I couldn't quickly think of a maintainable way to update this to still remove `std` and `default` other than hardcoding a filter in here, but maybe that's fine?

packages/driver/src/reflection/queries/types.ts
jaclarke: That's checking that the type itself isn't abstract. I meant checking the ancestors are all abstract as well like this:
```
select schema::ScalarType {
  name,
} filter not .abstract
  and not exists .enum_values
  and not exists (select .ancestors filter not .abstract)
```

packages/generate/src/syntax/funcops.ts
jaclarke: Yeah, I think it's fine to always show the full module + function name in error messages. You can probably also shadow the top level aliases of the stdlib functions anyway, and I think it's not worth trying to handle all that correctly just for an error message.

scotttrinh: Actually, I wonder if it's even worth trying to typescript-ify the function names at all 🤔 I think it would be maybe more clear if the error message was pointing to the actual EdgeQL function that was called rather than trying to construct what it probably looks like at the call site? In other words, just use `funcName` here. Thoughts?

packages/driver/src/reflection/queries/types.ts
scotttrinh: Oh yeah, maybe I don't understand after all: where would this query go?

jaclarke: Instead of the current `material_scalars` query.

scotttrinh: Oh, so don't even filter on the name at all? We don't need the shape here, right? At the very least we also want at least the `id` since that's what we actually consume. I'll make the change and maybe add a test for `std`, `cal`, `ext`.

scotttrinh: @jaclarke ❓ [`992f2f0` (#686)](https://github.com/edgedb/edgedb-js/pull/686/commits/992f2f0e4cb35cccfb148794d77589d25d96a830)

Does this seem right? I left the name filtering in, but do you think that's unnecessary?

jaclarke (approved):

PR #691, opened 2023-07-18
Improve ClientConnectionError message

janwirth:
In response to

https://discord.com/channels/841451783728529451/1113918639985401896/1127633227729866873

### DIFF
--- a/packages/driver/src/conUtils.ts
+++ b/packages/driver/src/conUtils.ts
@@ -632,7 +632,9 @@ async function parseConnectDsnAndArgs(
     } else {
       throw new errors.ClientConnectionError(
         "Found 'edgedb.toml' but the project is not initialized. " +
-          "Run `edgedb project init`."
+          "If you are developing, run `edgedb project init`." + 
+          "If you are running this project in production, use a DSN." +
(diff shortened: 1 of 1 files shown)


packages/driver/src/conUtils.ts
```diff
@@ -632,7 +632,9 @@ async function parseConnectDsnAndArgs(
     } else {
       throw new errors.ClientConnectionError(
         "Found 'edgedb.toml' but the project is not initialized. " +
-          "Run `edgedb project init`."
+          "If you are developing, run `edgedb project init`." + 
+          "If you are running this project in production, use a DSN." +
+          "Read more here: [https://github.com/edgedb/edgedb-js#configuring-the-connection](https://github.com/edgedb/edgedb-js#configuring-the-connection)"
```
scotttrinh, line 637: We're missing some whitespace here. Feel free to reformat this with string literals and have it all on one line.

Also just put the URL directly in since few, if any, of the places where you would see this error render Markdown.

scotttrinh (changes requested):
Thanks for hoping on this! A few small changes requested.

packages/driver/src/conUtils.ts
janwirth: Okay, great catch! Thank you.

PR #698, opened 2023-07-25
Fix `createHttpClient` not using `secretKey` when connecting to cloud instances

jaclarke:


diksipav (approved):

scotttrinh (approved):
Cool. Changes look good. I'll look at beefing up our tests to cover this config stuff since it's starting to get a little more interesting now.

PR #700, opened 2023-07-26
Respect an explicit false for updateIgnoreFile

CarsonF:
I run this command in docker build to generate the files.
It's not the first run; I've already configured my repo.

Also, for some reason this logic causes the docker build to exit early instead of finishing the generation.
I didn't look too closely into that, because I never want this logic to run.


CarsonF:
@scotttrinh awfully quiet here for a one liner. On the fence?

scotttrinh (approved):
Yeah, this seems fine. If you're explicitly opting into this, we shouldn't overwrite your option. 👍

PR #701, opened 2023-07-26
Allow custom suffix for generated query files

CarsonF:
Happen to call the option something else.
It doesn't apply to the other generators, so I didn't put in the common `CommandOptions` type.

I know I didn't connect to CLI handling either.
I don't need that since I call the function directly, but I can connect if you'd like.


CarsonF:
From my perspective, EdgeDB is assuming a lot, and maybe over stepping, to think that all `*.query.ts` files are generated edgedb queries and should be ignored.

This was the smallest diff so I started here. I'll look into a larger refactor instead.

CarsonF:
Closing in favor of #706

PR #702, opened 2023-07-26
Allow custom generator command to be given for generated header comment

CarsonF:
We use yarn not npm, so I don't want my devs seeing to use `npx` instead of `yarn dlx`.
Furthermore, I have a custom script to run these generators so I'd like to reference that.


scotttrinh:
Actually, I'm not personally attached to adding the instructions in the header, and I'm a very soft -1 on adding more options in general. Feel free to refactor this to just remove the explicit instructions altogether!

scotttrinh (approved):

PR #703, opened 2023-07-26
 Remove/move exit calls from generators to cli.ts

CarsonF:
Currently I have `watch: true` only to prevent the function from calling exit. Very weird lol.
It appears this functionality is stubbed, so dropping this seems fine. Even when implemented I'd question if this would be needed.

### DIFF
--- a/packages/generate/src/cli.ts
+++ b/packages/generate/src/cli.ts
@@ -367,29 +367,33 @@ Run this command inside an EdgeDB project directory or specify the desired targe
     exitWithError(`Failed to connect: ${(e as Error).message}`);
   }
 
-  switch (generator) {
-    case Generator.QueryBuilder:
-      await generateQueryBuilder({
-        options,
-        client,
-        root: projectRoot,
-      });
-      adapter.process.exit();
-      break;
-    case Generator.Queries:
-      await generateQueryFiles({
-        options,
-        client,
-        root: projectRoot,
-      });
-      break;
-    case Generator.Interfaces:
-      await runInterfacesGenerator({
-        options,
-        client,
-        root: projectRoot,
-      });
-      break;
+  try {
+    switch (generator) {
+      case Generator.QueryBuilder:
+        await generateQueryBuilder({
+          options,
+          client,
+          root: projectRoot,
+        });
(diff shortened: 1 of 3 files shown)


scotttrinh:
Can you help me understand why you don't want to call `exit`, are you calling this programmatically?

I think the approach I would take here would be to move the exit call into the caller (`cli.ts`) after the switch that calls the various generator functions so that the generators don't have to be concerned with the running process.

CarsonF:
Yes calling programmatically.
Ha yeah, I agree. After I PR'ed this I saw other spots we were exiting on.

After we resolve the other PRs I'll circle back to this and move the exits to cli.ts

CarsonF:
It looks like the only overall logical difference now is the queries generator doesn't have an explicit `exit(0)` call.
This should be fine, but I haven't tested....
maybe it was there because of some wonky unhandled async ops...
probably not though, probably just left over code based on the state of these files lol

packages/generate/src/cli.ts
```diff
@@ -367,29 +367,33 @@ Run this command inside an EdgeDB project directory or specify the desired targe
...
-      break;
+  try {
+    switch (generator) {
+      case Generator.QueryBuilder:
+        await generateQueryBuilder({
+          options,
+          client,
+          root: projectRoot,
+        });
+        adapter.process.exit();
```
scotttrinh, line 378: I think we should explicitly exit from the process in the CLI for all generators.

Can we move this to a `finally` block?

scotttrinh (changes requested):

scotttrinh:
> maybe it was there because of some wonky unhandled async ops...
> probably not though, probably just left over code based on the state of these files lol

I'm generally fine with hearing about it through bug reports or on Discord and fixing it the right way: racing a Promise with a timeout or something more explicit like that.

packages/generate/src/cli.ts
CarsonF: Ah I missed that one. I'll just drop it down to after the switch block. The finally isn't needed as the catch exits.

scotttrinh (approved):

PR #704, opened 2023-07-26
Create client in one spot and pass into generators

CarsonF:
This DRYs the code and also allows a client to be re-used when calling multiple functions programmatically


scotttrinh (approved):
Love it!

PR #705, opened 2023-07-27
ESM compatibility for generated query builder

CarsonF:
With these files copied wholesale to the user's codebase, they need to be compatible with ESM syntax.
This shouldn't conflict with commonjs functionality.


scotttrinh:
I don't think this works as-is (see the failing tests). I'm under the impression that the different targets should do the required transformations, and if that's not working correctly, we should fix that rather than modifying the import paths since they don't actually exist at those locations.

CarsonF:
Yes I'm dumb. I just needed to switch to `mts` target. Thanks!

PR #710, opened 2023-08-02
Add basic support for protocol v2

jaclarke:


packages/driver/src/codecs/registry.ts
```diff
@@ -294,6 +309,13 @@ export class CodecsRegistry {
 
       case CTYPE_SHAPE:
       case CTYPE_INPUT_SHAPE: {
+        if (t === CTYPE_SHAPE && isProtoV2) {
+          // @ts-expect-error: reserved for future use
+          const isEphemeralFreeShape = frb.readBoolean();
```
scotttrinh, line 314: We can now use the `_` prepend to avoid this TS error instead:

```suggestion
          const _isEphemeralFreeShape = frb.readBoolean();
```

Likewise for the others.

scotttrinh:
Closed in favor of #1091

PR #711, opened 2023-08-02
Add multirange support

jaclarke:
Depends on #710 

Todo:
- [ ] Add tests/docs for `MultiRange` type. (Do we want to add multirange normalisation similar to python and dart implementations?)
- [ ] Update querybuilder to support multiranges


jaclarke:
Superceded by #733

PR #716, opened 2023-08-14
Fix typescript errors for generated query builder

progrestian:
Trying to use the query builder together in a SvelteKit project throws some errors.
The fix is to just do what the compiler suggested. 

Found another user hit with the bug on the main repo: https://github.com/edgedb/edgedb/issues/5486


scotttrinh (approved):
Yep! I've enabled an eslint rule that help catch these too, and have been fixing them when I encounter them. Thanks for the contribution!

PR #721, opened 2023-08-23
Pass crypto into functions instead of direct imports to remove `require('crypto')` imports in browser client

### LINKED ISSUE #570: Edge runtime compatibility (Vercel, Cloudflare workers...)
Hi all,

I tried using the HTTP client to make EdgeDB work onto Vercel's [edge runtime](https://beta.nextjs.org/docs/rendering/edge-and-nodejs-runtimes) as there are non-negligible gains (on cold boots). But with no success.

It seems the query-builder uses node's crypto module.
Is there any plan on supporting the edge runtime sometime in the future ?

Bests

jaclarke:
When using edgedb-js in the nextjs edge runtime environment, the existence of `require('crypto')` in the browser client causes errors at build time, as nextjs can't tell it will never actually get called at runtime.

This PR refactors any direct imports of the crypto related functions, to instead pass them to functions that need them (like scram), so the browser client can provide alternate crypto without any `require('crypto')` existing in the codepath.

Should fix #570 (but we should probably add some actual tests to check edge runtime/cloudflare worker/etc.. compatibility)


scotttrinh (approved):
Changes make sense! Not sure how best to actually test Edge runtimes in CI, but we can investigate that as a separate effort and just manually test this with our own usage for now.

PR #725, opened 2023-09-04
Fix cardinality error

sharifzadesina:
Hello,

This is a simple PR aiming to fix cardinality error causing by the sample query provided in docs.


scotttrinh (approved):
Thanks for catching!

sharifzadesina:
Closing the pull request since it seems the examples are just fine and I forgot to add `constraint exclusive;` to my `title` property, with `constraint exclusive;`, we no longer need `limit 1` and the queries provided in examples will just work.

PR #733, opened 2023-09-13
Add minimal support for 4.x

scotttrinh:
A few new features have been developed in 4.0 that needed some updates here to get the driver and query builder to work:

1. `multirange` which is basically a collection type of ranges.
2. Full-text search has introduced a new pseudo-type called `anyobject` which is a subset of `anytype`.

There was also a change to cardinality inference that made certain polymorphic queries an error condition now, and that has been addressed here as well.


bastien8060:
Would this fix the 'Cannot find 'anytype' in multirange<std::anypoint>' error while introspecting the database schema, when generating query builder? (EdgeDB 4.x alpha)

scotttrinh:
@bastien8060 

> Would this fix the 'Cannot find 'anytype' in multirangestd::anypoint' error while introspecting the database schema, when generating query builder? (EdgeDB 4.x alpha)

That's right: it's been failing for some time now with that message, and I was hoping there would be an easy (yet wrong) workaround for this ahead of actually implementing support, but alas, nothing just fell into my lap yesterday.

Hope to get multirange support for 4.x-alpha in the next few weeks though, stay tuned!

bastien8060:
@scotttrinh Awesome! I'll be keeping an eye out :) Currently depending on it, so I'll just revert to the client.query methods for now!

scotttrinh:
@bastien8060 You're depending on 4.0-alpha?

bastien8060:
@scotttrinh Actually sort of, so my application is rather latency critical, and it seems the first query is much faster on 4.0-alpha than on 3.x, but still not quite there imo. And it's funny, but it's not an issue I noticed on the python client. I might submit an issue on the JS client I think—

On the JS client, the first request always takes a couple seconds, regardless of how long the client has been initialized for. 

```js
const baseClient = createClient();

const configClient = baseClient.withConfig({
  "session_idle_transaction_timeout": Duration.from({ seconds: 0 }),
});

const client = await configClient.ensureConnected();
```

integration-tests/lts/select.test.ts
```diff
@@ -310,14 +310,15 @@ describe("select", () => {
   test("* in polymorphic", async () => {
     const q = e.select(e.Person, () => ({
       ...e.is(e.Hero, e.Hero["*"]),
+      name: true,
```
scotttrinh, line 313: This change reflects the new behavior of 4.x where the parent type's cardinality is considered when using polymorphic queries. In this case, `Person` has a required `name`, but the query can return `Person` records that are not `Hero`s, which would return a `null` name, which is invalid for `Person`.

packages/driver/src/datatypes/range.ts
```diff
@@ -63,3 +63,25 @@ export class Range<
...
+
+  *[Symbol.iterator]() {
+    for (const range of this._ranges) {
+      yield range;
+    }
+  }
+
+  toJSON() {
+    return {
+      ranges: this._ranges,
```
jaclarke, line 84: I think this should probably just return the array, to mirror edgedb's json representation of multiranges.

jaclarke: Also probably should copy the array so you can't mutate it. (Or maybe this should be done in the constructor? or both?)

scotttrinh: Cool, I think I just copied this from the python impl, but that makes sense!

packages/driver/src/reflection/analyzeQuery.ts
```diff
@@ -144,5 +144,13 @@ function walkCodec(
     ctx.imports.add("Range");
     return `Range<${subCodec.tsType}>`;
   }
+  if (codec instanceof MultiRangeCodec) {
+    const subCodec = codec.getSubcodecs()[0];
+    if (!(subCodec instanceof ScalarCodec)) {
+      throw Error("expected multirange subtype to be scalar type");
+    }
+    ctx.imports.add("MultiRange");
+    return `MultiRange<${subCodec.tsType}>`;
```
jaclarke, line 153: You can have ranges of `LocalDate/Time`, so we also need to check if the subCodec type needs adding to `ctx.imports`

jaclarke:
While we're messing with ranges, could we also add this small fix to the range constructor: https://github.com/edgedb/edgedb-js/pull/711/files#diff-96a05a9179de11464b9faa257493af517dcd41f1d1b188b059a748dbd55e463aR29

packages/driver/src/reflection/analyzeQuery.ts
scotttrinh: This mirrors what we're doing in `Range`, do you think that's also broken? Looking at how `ArrayCodec` is done here, maybe I need to wrap this and the `Range` subcodec in a recursive call to `walkCodec`?

jaclarke: Yeah, I think you're right, `Range` also looks like it has the same problem. Either `walkCodec` or since we know the subCodec is going to be a scalar we could just inline the `codec.importedType` check, both ways seem fine to me.

scotttrinh: I'll stick with `walkCodec` just in case other behavior falls out and that seems to match how we're doing other types with subcodecs. Thanks for catching this!

jaclarke (approved):

PR #741, opened 2023-09-22
Cloudfalre change Node Imports to supported node: prefix

Jumaron:
Currently it is not possible to use Edgedb-js natively with Cloudflares Pages with there Edge network (Cloudflare Workers).
The Problem of this, are most likely connected to the imports, which cannot be found for the Workers since we need to add coresponding to Cloudflare: https://developers.cloudflare.com/workers/runtime-apis/nodejs/ 
The node: prefix.

The main point of this, is also to discuss future improvement on the integretation of Cloudflare Pages, which is one of the most popular Edge deployment methodes.

### DIFF
--- a/packages/driver/src/adapter.node.ts
+++ b/packages/driver/src/adapter.node.ts
@@ -1,13 +1,13 @@
-import * as crypto from "crypto";
-import { promises as fs } from "fs";
-import * as net from "net";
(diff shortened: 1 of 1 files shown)


hrasoa:
Hi, any news on this ?

PR #750, opened 2023-10-18
Add low-level auth lib

jaclarke:


packages/auth-core/src/pkce.ts
```diff
@@ -0,0 +1,14 @@
+import crypto from "node:crypto";
+
+export function createVerifierChallengePair(): {
+  verifier: string;
+  challenge: string;
+} {
+  const verifier = crypto.randomBytes(32).toString("hex");
```
scotttrinh, line 7: While this is strictly fine (hex is a subset of `base64url`), I think we should use `base64url` here. You probably copied this verbatim from my flawed first implementation, but I caught this discrepancy when writing the docs.

```suggestion
  const verifier = crypto.randomBytes(32).toString("base64url");
```

scotttrinh (changes requested):
A few piece of minor feedback, and one thing we ought to change now: `hex` -> `base64url`.

packages/auth-nextjs/src/app/index.ts
```diff
@@ -0,0 +1,551 @@
...
+                `'onEmailPasswordSignIn' auth route handler not configured`
+              );
+            }
+            let tokenData: TokenData;
+            try {
+              const [email, password] = _extractParams(
+                await _getReqBody(req),
+                ["email", "password"],
+                "email or password missing from request body"
+              );
```
scotttrinh, line 278: We also need a `verify_url` here, otherwise it falls back to the built-in UI verify.

packages/auth-core/src/core.ts
```diff
@@ -0,0 +1,291 @@
...
+  async resetPasswordWithResetToken(resetToken: string, password: string) {
+    return this._fetch("reset-password", "post", undefined, {
+      provider: "builtin::local_emailpassword",
+      reset_token: resetToken,
+      password,
+    }) as Promise<TokenData>;
+  }
+
+  async getProvidersInfo() {
+    // TODO: cache this data somehow?
```
MiroslavPetrik, line 190: You can memoize easily on a [private class property](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDOBXANsKBeKBvAUFfUAJgIbDEBcUARgPY3ITEB2UOAvgNxQD03UokfjSgAnCGDHwITNMAAW0AMbFFCwkVLEcORcmLx4UAILY8BXlAkBLAG6loNMMCs0mxZFABmViMkJn8AGJlVQgAfko4JFQObQIoAHMIYAARTQAKAEpTePirTyh0+St4ADpglQVs3FzaxVd4eghS5BoE9IAiEIUoOStgDszY2tyxYEQRFmKyitDhkbY4kfqmRoYWts7u6ABbEvhB+dqLHZpFAGsvZND1EjIA3JX4NEkUNEwsO4p+EUQITge8Qs3mQHm2gII03K2ww2C+lGAv3+sQh+DGEymfRm2yOUDYrEWOgaaGIsKYEAA7sYstpiKUkqkMtkcHSGWkyDTWcl2cQaUA)

packages/auth-core/src/core.ts
```diff
@@ -0,0 +1,291 @@
...
+    const providers = (await this.client.query(`
+    with module ext::auth
+    select cfg::Config.extensions[is AuthConfig].providers {
```
MiroslavPetrik, line 193: This throws when the `using extension auth` was not applied to the schema.

```
 InvalidReferenceError: type 'AuthConfig' does not exist
```

packages/auth-nextjs/src/app/index.ts
jaclarke: I can't see that we use `verify_url` anywhere in the `authenticate` endpoint? (https://github.com/edgedb/edgedb/blob/master/edb/server/protocol/auth_ext/http.py#L473)

packages/auth-core/src/core.ts
jaclarke: Yes, while the caching bit is easy, this note was more of a reminder to add it once we figure out a way to invalidate the cached provider info when the config is updated.

packages/auth-nextjs/src/app/index.ts
scotttrinh: Oh, sorry, I meant to put this on the `/signup` endpoint: https://github.com/edgedb/edgedb/blob/57b82842d02e2c946754baa22248b2d381b85b0e/edb/server/protocol/auth_ext/http.py#L429-L431

scotttrinh: I see now that it's the third argument to the `signupWithEmailPassword` 👍

scotttrinh (approved):
New changes look good!

packages/auth-core/src/core.ts
```diff
@@ -0,0 +1,289 @@
...
+  async getProvidersInfo() {
+    // TODO: cache this data when we have a way to invalidate on config update
+    let providers: {
+      _typename: string;
+      name: string;
+      display_name: string | null;
+    }[];
+    try {
+      providers = await this.client.query(`
+      with module ext::auth
```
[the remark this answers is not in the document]
jaclarke: That's a good idea. Thanks.

scotttrinh (approved):

packages/auth-core/src/core.ts
```diff
@@ -175,39 +175,27 @@ export class Auth {
...
+      }>(`
+      with
+        module ext::auth,
+        providers := (select cfg::Config.extensions[is AuthConfig].providers)
+      select {
+        oauth := providers[is OAuthProviderConfig] {
+          name,
+          display_name
+        },
+        emailPassword := exists providers[is EmailPasswordProviderConfig]
```
scotttrinh, line 191: Not a blocker: it seems like we should still maybe return the `name` here so we don't have a magic const string in the code?

jaclarke: I'm not sure that would have much use, since I decided to not expose that detail elsewhere as the api's are specific to the email/password provider anyway. But internally, yes, it's probably a good idea to use the emailPasswordProvider name from the generated consts file, instead of how it's currently hardcoded.

packages/auth-core/src/core.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  static async create(client: edgedb.Client) {
+    const connectConfig: ResolvedConnectConfig = (
+      await (client as any).pool._getNormalizedConnectConfig()
+    ).connectionParams;
+
+    const [host, port] = connectConfig.address;
+    const baseUrl = `${
+      connectConfig.tlsSecurity === "insecure" ? "http" : "https"
+    }://${host}:${port}/db/${connectConfig.database}/ext/auth`;
```
MiroslavPetrik, line 34: @jaclarke  There is missing slash `/`

```js
new URL("foo", "http://localhost:3000/auth").toString()
// 'http://localhost:3000/foo'
new URL("foo", "http://localhost:3000/auth/").toString()
// 'http://localhost:3000/auth/foo'
```

This causes wrong builtin urls to be created (beside other)

image

jaclarke: Yep, you're right. Thanks for catching that.

Issue #376, opened 2022-06-29
Deno + EdgeDB solution issue


Jakkalsie:
I'm trying to use Deno and EdgeDB together in a new framework called Fresh, but I'm running into the TLS issue [EdgeDB-Deno](https://github.com/edgedb/edgedb-deno#tls). After reading the solution proposed, I'm still not able to connect to the client.

I think I am doing something wrong, but not sure what. Can you maybe offer a more in depth explanation or an example I can follow?

Some direction would be helpful.
Best regards

1st1:
cc @jaclarke

jaclarke:
The readme was out of date (I've just updated it now). Deno now supports tls alpn (behind the `--unstable` flag), so you shouldn't need to set any edgedb server flags anymore for edgedb-deno to work.

I've just tried out deno's Fresh framework, following the getting started instructions (https://fresh.deno.dev/#getting-started), and for me it works with edgedb-deno by just adding the `--unstable` flag to the 'start' task.

jaclarke:
https://github.com/edgedb/edgedb-examples/issues/76

Jakkalsie:
Thank you for the feedback, that is perfect!

Jakkalsie:
There must something fundamental I'm missing, but after adding `--unstable` to my `deno.json` so that it looks like this:
```json
{
  "tasks": {
    "start": "deno run -A --watch=static/,routes/ --unstable dev.ts"
  },
  "importMap": "./import_map.json"
}
```

and this in my main `index.tsx` page
```js
/** @jsx h */
import { h } from "preact";
import { tw } from "@twind";
import { Handlers, PageProps } from "https://deno.land/x/fresh@1.0.0/server.ts";
import * as edgedb from "https://deno.land/x/edgedb@v0.20.10/mod.ts";

export const handler: Handlers = {
  async GET(_, ctx) {
    const client = edgedb.createClient();

    const result = await client.query(`select 2 + 2;`);

    return ctx.render(result);
  },
};

export default function Home({ data }: PageProps) {
  console.log(data);
}
```

It keeps throwing the following error:

```
A client connection error occurred; reconnecting because of "waitUntilAvailable=30000". ClientConnectionClosedError: network error: ConnectionRefused: No connection could be made because the target machine actively refused it. (os error 10061)
    at RawConnection._onError (https://deno.land/x/edgedb@v0.20.10/_src/rawConn.ts:110:20)
[...]

PR #769, opened 2023-11-08
Fix auth urls

jaclarke:


jaclarke:
> Ahh, nice catch. Without the trailing slash, the new "path" replaces the last path segment, yeah?

Yes, MiroslavPetrik caught it: https://github.com/edgedb/edgedb-js/pull/750#discussion_r1386685436. I think I broke it in the refactor to use `URL`s.

jaclarke:
I'll add some tests for this once I figure out how to mock smtp.

packages/auth-core/src/core.ts
```diff
@@ -31,7 +31,7 @@ export class Auth {
     const [host, port] = connectConfig.address;
     const baseUrl = `${
       connectConfig.tlsSecurity === "insecure" ? "http" : "https"
-    }://${host}:${port}/db/${connectConfig.database}/ext/auth`;
+    }://${host}:${port}/db/${connectConfig.database}/ext/auth/`;
```
MiroslavPetrik, line 34: here I found one more problem, Next does not want to connect to host of `localhost`, so I have

    const hostFix = host === "localhost" ? "127.0.0.1" : host;


```suggestion
    }://${hostFix }:${port}/db/${connectConfig.database}/ext/auth/`;
```

MiroslavPetrik: image

original error,
found the solution in 
https://stackoverflow.com/questions/74165121/next-js-fetch-request-gives-error-typeerror-fetch-failed

Not sure if its not only local, my node version is

$ node -v
v18.18.2

scotttrinh: I think it's related to this issue: https://github.com/nodejs/node/pull/44731

I believe this is fixed in Node v20 (see https://github.com/nodejs/node/pull/44731), but maybe there is something we can do to opt-into the same behavior in our codebase here. I'll look into that.

MiroslavPetrik: Just tested with node v20.9.0, and it still fails

scotttrinh: What is your operating system? If Node v20 is still failing in the same way, it seems like the DNS resolver is not returning `127.0.0.1` for `localhost` lookup which is definitely odd.

MiroslavPetrik: I'm on Windows 11. Solved it with (for now)
```
import dns from "node:dns";

dns.setDefaultResultOrder("ipv4first");
```
it also impacted the default edgedb client connection. You are right with the DNS resolver, it prioritizes the ipv6, as in the screenshot of the error we see `::1`. 

I don't know where to change the priority in the OS.

scotttrinh: @MiroslavPetrik perhaps this comment might help: https://github.com/nodejs/node/issues/40537#issuecomment-1706257550

MiroslavPetrik: Thanks, will save that as alternative option for developing on windows 11.

Still it seems temporary, just noticed that even the hosts file no longer uses localhost mapping (they are commented out) as that somehow interferes with ipv6 adoption, so maybe the edgedb could listen on ipv6 as well? 
I don't know much about networking. Just expanding on what I just read

jaclarke: Yes, it's possible to make edgedb listen on ipv6 as well. There was a similar issue previously with Deno resolving localhost to the ipv6 `::1` address: https://github.com/edgedb/edgedb-js/issues/376#issuecomment-1173840632

PR #773, opened 2023-11-09
Release edgedb-js v1.4.1

jaclarke:


scotttrinh:
I want to ensure the release updates for publishing the auth packages is working before we publish another version, just in case we accidentally break these. I'll merge once it seems like it's working.

scotttrinh (approved):

PR #776, opened 2023-11-10
Convert selecting a named tuple into a free object

scotttrinh:


packages/generate/src/syntax/select.ts
```diff
@@ -996,6 +1050,15 @@ export function select(...args: any[]) {
 
   let expr = exprPair[0];
   const shapeGetter = exprPair[1];
+  if (expr.__element__.__kind__ === TypeKind.namedtuple) {
+    expr = $for(expr as TypeSet<NamedTupleType>, (forScope) => {
+      const shape: any = {};
+      for (const k of Object.keys(forScope.__element__.__shape__)) {
+        shape[k] = op("distinct", (forScope as any)[k]);
```
scotttrinh, line 1057: This `distinct` operator is a hack for dealing with https://github.com/edgedb/edgedb/issues/6530.

PR #784, opened 2023-11-15
Use the configured passwordResetUrl in `auth-nextjs`

MiroslavPetrik:
When requesting the reset password email, the URL from config is used, but was not really set during `auth` initialization.
The `resetUrl` param is not supported, so I've removed it from the type.

### DIFF
--- a/packages/auth-nextjs/src/app/index.ts
+++ b/packages/auth-nextjs/src/app/index.ts
@@ -466,7 +466,7 @@ export class NextAppAuth extends NextAuth {
         }
       },
       emailPasswordSendPasswordResetEmail: async (
-        data: FormData | { email: string; resetUrl: string }
+        data: FormData | { email: string }
       ) => {
         if (!this.options.passwordResetUrl) {
           throw new Error(`'passwordResetUrl' option not configured`);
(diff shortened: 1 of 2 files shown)


MiroslavPetrik:
@scotttrinh I see that only the links for the builtin signin/signout are exposed via helper methods. Will there be such methods for the [rest of the urls, like the password reset email?](https://github.com/edgedb/edgedb/blob/57b82842d02e2c946754baa22248b2d381b85b0e/edb/server/protocol/auth_ext/http.py#L107C13-L107C13)

I can imagine, that for custom UI I would ship initially just the Sign in and sign out screens, and for the password flows I would fallback to the default ui. That way, I can progressively ship just the screens which users use the most and fallback to default for the infrequent flows.

scotttrinh:
> I see that only the links for the builtin signin/signout are exposed via helper methods. Will there be such methods for the [rest of the urls, like the password reset email?](https://github.com/edgedb/edgedb/blob/57b82842d02e2c946754baa22248b2d381b85b0e/edb/server/protocol/auth_ext/http.py#L107C13-L107C13)

I don't think that's currently planned since those links have been considered internal and subject to change. Perhaps in the future we might be alright with stabilizing those internal routes, but in the meantime you're welcome to link to them as long as you realize that they're subject to breaking changes in minor versions.

> I can imagine, that for custom UI I would ship initially just the Sign in and sign out screens, and for the password flows I would fallback to the default ui. That way, I can progressively ship just the screens which users use the most and fallback to default for the infrequent flows.

Yeah, I can see the value in that approach, although I think it's pretty likely that the cost of building those few screens yourself is probably worth it to match the design and feel of the sign in and sign up pages. It would feel jarring for them to be so different, and if they were very similar, I'm not sure it makes sense that you would do something so similar but make it custom.

scotttrinh (approved):

PR #785, opened 2023-11-16
Multiple fixes/improvements to auth-nextjs

jaclarke:
Depends on https://github.com/edgedb/edgedb/pull/6469

### DIFF
--- a/packages/auth-nextjs/src/app/index.ts
+++ b/packages/auth-nextjs/src/app/index.ts
@@ -29,7 +29,11 @@ type ParamsOrError<Result extends object, ErrorDetails extends object = {}> =
 
 export interface CreateAuthRouteHandlers {
   onOAuthCallback(
-    params: ParamsOrError<{ tokenData: TokenData; isSignUp: boolean }>
+    params: ParamsOrError<{
+      tokenData: TokenData;
+      provider: BuiltinOAuthProviderNames;
+      isSignUp: boolean;
+    }>
   ): void;
   onEmailPasswordSignIn(params: ParamsOrError<{ tokenData: TokenData }>): void;
   onEmailPasswordSignUp(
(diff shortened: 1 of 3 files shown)


packages/auth-nextjs/src/app/index.ts
```diff
@@ -9,26 +9,53 @@ import { cookies } from "next/headers";
...
+  type BuiltinProviderNames,
+  type TokenData,
+};
 
-type ParamsOrError<Result extends object> =
-  | ({ error: null } & Result)
-  | ({ error: Error } & { [Key in keyof Result]?: undefined });
+type ParamsOrError<Result extends object, ErrorDetails extends object = {}> =
+  | ({ error: null } & { [Key in keyof ErrorDetails]?: undefined } & Result)
+  | ({ error: Error } & ErrorDetails & { [Key in keyof Result]?: undefined });
```
scotttrinh, line 28: Not a blocker: does moving the payload into a separate key make it easier for this to be a simple discriminated union?

Something like:
```suggestion
type ParamsOrError<Result extends object, ErrorDetails extends object = {}> =
  | { error: null; params: Result }
  | { error: Error; details: ErrorDetails };
```

packages/auth-nextjs/src/shared.ts
```diff
@@ -65,8 +74,12 @@ export class NextAuthSession {
 
   async isLoggedIn() {
     if (!this.authToken) return false;
-    return (await this.client.querySingle(
-      `select exists global ext::auth::ClientTokenIdentity`
-    )) as boolean;
+    try {
+      return (await this.client.querySingle(
+        `select exists global ext::auth::ClientTokenIdentity`
+      )) as boolean;
```
scotttrinh, line 80: I know this isn't new and is definitely not a blocker, but can we move the `boolean` into the generic position rather than using a cast here?

```suggestion
      return await this.client.querySingle<boolean>(
        `select exists global ext::auth::ClientTokenIdentity`
      );
```

scotttrinh (approved):
A few small initial thoughts. No blockers.

PR #791, opened 2023-11-27
Update temporal shapes to be compatible with Temporal API instances

CarsonF:
This is essentially just removing/"hiding" private fields, so that TS doesn't try to use those fields when checking type compatibility.

Previously, this failed type check, and now it doesn't
```ts
const ld: edgedb.LocalDate = Temporal.PlainDate.from('2023-11-27');
```

There are `instanceof` checks on our temporal classes in some places, so this isn't a complete solution but it does help ease TS friction.

### DIFF
--- a/packages/driver/src/codecs/datetime.ts
+++ b/packages/driver/src/codecs/datetime.ts
@@ -28,6 +28,7 @@ import {
   DateDuration,
   LocalDateFromOrdinal,
   LocalDateToOrdinal,
+  localDateInstances,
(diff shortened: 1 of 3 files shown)


scotttrinh (approved):
Looks great! Thanks for the extra `any` squashing here as well, I'd love to do the same mechanical transformations to the other codecs sometime soon. I'm pleasantly surprised it didn't end up having any downstream impacts.

PR #793, opened 2023-11-28
Merge default types into modules with the same name (query builder)

CarsonF:
Given the schema
```
module default {
  type User {
    required status: User::Status;
  }
}
module User {
  type Status extending enum<Active, Disabled>;
}
```
In EdgeQL we can do
```
insert User {
  status := User::Status.Active
}
```
However, with the query builder the `User` type was not flattened because there's a module named `User`.
So we had to do
```ts
e.insert(e.default.User, {
  status: e.User.Status.Active,
});
```
With this PR we can do
```ts
e.insert(e.User, {
  status: e.User.Status.Active,
});
```
Which more closely matches EdgeQL


scotttrinh:
Do you mind adding a quick test case to the LTS schema [here](https://github.com/edgedb/edgedb-js/blob/master/integration-tests/lts/dbschema/default.esdl)? You can add a new dummy type and dummy module with a type and add a low-effort test to the select tests there to ensure all of the type inference works out? If not I can stick it in my todo list.

CarsonF:
Absolutely, and thanks for the starting point. Wanted to get the concept in front of you before putting more work in.

packages/generate/src/edgeql-js/generateIndex.ts
```diff
@@ -122,7 +127,10 @@ export function generateIndex(params: GeneratorParams) {
         allowFileExt: true,
       });
 
-      index.writeln([r`${genutil.quote(moduleName)}: _${internalName},`]);
+      const valueStr = defaultSpreadTypes.has(moduleName)
+        ? `{ ..._${internalName}, ..._default.${moduleName} }`
```
CarsonF, line 131: Should this be frozen? Following suit of https://github.com/edgedb/edgedb-js/blob/d99c022a02b746585af7b706d3f7ba9fadc81d22/packages/generate/src/syntax/path.ts#L433-L433

PR #795, opened 2023-11-30
Support new functional `if..then..else` syntax

### LINKED ISSUE #794: Support new functional `if..then..else` operator syntax
Released in 4.x: https://www.edgedb.com/docs/stdlib/set#operator::if..then..else

scotttrinh:
We still map this to the old style in the generated EdgeQL for now. Since the reflection doesn't give us separate operator definitions, we don't have an easy way to reflect which form is available to you.

We can make this distinction in the future by providing an optional `__form__` expression property and generating the appropriate EdgeQL based on that, but for now since we don't have an easy way to tell ifthe EdgeDB server supports it without hardcoding some notion of that into the generated logic, seems best to not do this yet.

Closes #794

### DIFF
--- a/packages/generate/src/edgeql-js/generateOperatorTypes.ts
+++ b/packages/generate/src/edgeql-js/generateOperatorTypes.ts
@@ -108,146 +108,155 @@ export function generateOperators({
         overloadIndex++;
       }
 
-      if (opDef.description) {
-        overloadsBuf.writeln([
-          t`/**
+      const operatorForms = opName === "std::if_else" ? ["PYTHON", "FUNCTIONAL"] : ["NORMAL"];
+      for (const operatorForm of operatorForms) {
+        if (opDef.description) {
+          overloadsBuf.writeln([
+            t`/**
 * ${opDef.description.replace(/\*\//g, "")}
 */`,
-        ]);
-      }
+          ]);
+        }
 
-      overloadsBuf.writeln([dts`declare `, t`function op<`]);
+        overloadsBuf.writeln([dts`declare `, t`function op<`]);
 
-      const anytypes = opDef.anytypes;
-      const anytypeParams: string[] = [];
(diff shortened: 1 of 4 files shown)


packages/generate/src/edgeql-js/generateOperatorTypes.ts
```diff
@@ -108,146 +108,155 @@ export function generateOperators({
...
+              if (opName === "std::if_else") {
+                if (operatorForm === "PYTHON") {
+                  overloadsBuf.writeln([
+                    t`${args[0]}, op: "if", ${args[1]}, op2: "else", ${args[2]}`,
+                  ]);
+                } else {
+                  overloadsBuf.writeln([
+                    t`op: "if", ${args[1]}, op2: "then", ${args[0]}, op3: "else", ${args[2]}`,
+                  ])
+                }
```
scotttrinh, line 200: Forgive the indentation noise here: here's the actual change.

packages/generate/src/edgeql-js/generateOperatorTypes.ts
```diff
@@ -314,10 +323,18 @@ export function generateOperators({
...
       params = [args[0], args[2], args[4]];
       defs = overloadDefs.Ternary[op];
     }
+  } else if (args.length === 6) {
+    // Functional-style if-then-else
+    if (typeof args[0] === "string" && typeof args[2] === "string" && typeof args[4] === "string") {
+      op = \`\${args[0]}_\${args[4]}\`;
+      params = [args[3], args[1], args[5]];
+      defs = overloadDefs.Ternary[op];
+    }
```
scotttrinh, line 337: Note that this effectively maps both forms into the same operator which gets rendered as the same old `if..else` python style form.

In the future we can thread through a `__form__` property in the expression to output different forms here based on the input form or server version or capabilities, etc.

PR #797, opened 2023-12-01
Index `toEdgeQL` for doc search

### LINKED ISSUE #696: docs: Make sure `toEdgeQL` and similar APIs are documented and indexed as APIs
For example, `toEdgeQL` does not appear in search because it's not documented structurally (or, at least, not referenced in by an `:index:` directive).

scotttrinh:
Closes #696

### DIFF
--- a/docs/querybuilder.rst
+++ b/docs/querybuilder.rst
@@ -170,6 +170,7 @@ that later.
 
 Converting to EdgeQL
 --------------------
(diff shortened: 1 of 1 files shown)


raddevon (approved):
Similar to my comments on the other one, we should probably also add `querybuilder` at the top of this page so that it also appears in the results for that search query.

PR #798, opened 2023-12-01
Index `unlessConflict` for doc search

scotttrinh:


raddevon (approved):
As part of this, I think maybe we should also add `querybuilder` to [the main query builder page](https://www.edgedb.com/docs/clients/js/querybuilder) since otherwise, this page appears for `querybuilder` and it does not.

Slightly off-topic, but looking at this brought it to my attention. I think that page, which is titled "Query builder generator" is actually more about the query builder. It covers the generator also, but it has recipes for using the query builder itself which are not at all related to the generator. Should we change the page title?

PR #805, opened 2023-12-08
EdgeDB Auth: Express

### LINKED ISSUE #803: EdgeDB Auth: Express
Publish an auth helper package for [Express](https://expressjs.com/).

scotttrinh:
Closes #803 

Add an auth helper library for Express. Unlike some of the other helper libraries, this one stays a little lower level to fit in with the Express ecosystem, but still focuses you on just handling errors and successes rather than the plumbing.


packages/auth-express/readme.md
```diff
@@ -0,0 +1,398 @@
...
+
+## Using the session middleware: `createSessionMiddleware`
+
+We provide a middleware factory that will attach an `ExpressAuthSession` object to your request object, which you can use to make authenticated queries, or protect routes.
+
+```ts
+app.use(auth.createSessionMiddleware());
+```
+
+You can then use the `expressAuth.SessionRequest` type for your route's request parameter, which adds a `session` key to your request object.
```
haikyuu, line 62:
```suggestion
You can then use the `expressAuth.SessionRequest` type for your route's request parameter, which adds a `session` key to your request object type.
```

packages/auth-express/readme.md
```diff
@@ -0,0 +1,397 @@
...
+      res.redirect("/");
+    },
+  ],
+});
+
+app.use(emailPasswordRouter);
+// This creates the following routes:
+// - POST /auth/email-password/signin
+// - POST /auth/email-password/signup
+// - POST /auth/email-password/verify
```
jaclarke, line 199:
```suggestion
// - GET /auth/email-password/verify
```
This should be a get request right? since it's the link that gets sent in the email.

packages/auth-express/readme.md
```diff
@@ -0,0 +1,397 @@
...
+    "/signup",
+    auth.emailPassword.signUp(
+      // URL of the verify endpoint configured below
+      "http://localhost:3000/auth/email-password/verify"
+    ),
+    (req: expressAuth.AuthRequest, res) => {
+      res.redirect("/onboarding");
+    }
+  )
+  .post(
```
jaclarke, line 271: Same as above

packages/auth-express/readme.md
```diff
@@ -0,0 +1,397 @@
...
+  callback: [
+    (req: expressAuth.CallbackRequest, res, next) => {
+      if (req.isSignUp) {
+        return res.redirect("/onboarding");
+      }
+
+      res.redirect("/");
+    },
+    (error: any, req, res, next) => {
+      res.redirect(`/error?${encodeURIComponent(error.message)}`);
```
jaclarke, line 343:
```suggestion
      res.redirect(`/signin?error=${encodeURIComponent(error.message)}`);
```

packages/auth-express/src/index.ts
```diff
@@ -0,0 +1,568 @@
...
+      this.emailPassword.signUp(
+        new URL(`${routerPath}/verify`, this.options.baseUrl).toString()
+      ),
+      ...stacks.signUp
+    );
+    router.get("/verify", this.emailPassword.verify, ...stacks.verify);
+    router.post(
+      "/send-password-reset-email",
+      this.emailPassword.sendPasswordResetEmail(
+        new URL(`${routerPath}/reset-password`, this.options.baseUrl).toString()
```
jaclarke, line 145: Shouldn't this url be the user's reset password page, that contains a form that will then post to this `/reset-password` url?

packages/auth-express/src/index.ts
```diff
@@ -0,0 +1,568 @@
...
+        next();
+      } catch (err) {
+        next(err);
+      }
+    },
+    sendPasswordResetEmail:
+      (passwordResetUrl: string) =>
+      async (req: AuthRequest, res: ExpressResponse, next: NextFunction) => {
+        try {
+          if (!this.options.passwordResetPath) {
```
jaclarke, line 468: We don't seem to use this option, so maybe we don't need it for the express integration, and the user can just pass it directly when they're setting up the middleware?

packages/auth-express/src/index.ts
scotttrinh: oop, good catch! I forgot to test the reset flow, probably some other bugs there 🙈

packages/auth-express/readme.md
scotttrinh: This is the built-in UI so any errors that make it outside of the typical error handling in the built-in UI are probably fatal, so I intended this to redirect to back to the actual application. Good catch about missing the query parameter tho 🙈

jaclarke (approved):

PR #806, opened 2023-12-11
Refactor auth-nextjs to separate client and server for pages api + improve support for 'actions' in pages router

jaclarke:


packages/auth-nextjs/src/pages/client.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export { type NextAuthOptions, type BuiltinProviderNames };
+
+export default function createNextPagesClientAuth(options: NextAuthOptions) {
+  return new NextPagesClientAuth(options);
+}
+
+export class NextPagesClientAuth extends NextAuthHelpers {
+  async emailPasswordSignIn(
+    router: NextRouter,
```
scotttrinh, line 16: Is there a reason we cannot just use the imported `next/router` here directly from the library? Or pass it in the constructor?

scotttrinh (approved):
No blockers, just a question about making the router passing a little less verbose.

packages/auth-nextjs/src/pages/client.ts
jaclarke: Turns out we can. I had always thought you had to use the router object from the `useRouter` hook, since the `router` object import directly didn't seem to work, but it seems it does work if you use the default import 🤷:  https://github.com/vercel/next.js/discussions/17046

scotttrinh (approved):

Issue #807, opened 2023-12-13
Create `npm init/create` package for project scaffolding


scotttrinh:
Create an [`npm init`](https://docs.npmjs.com/cli/v10/commands/npm-init) "initializer" to easily scaffold projects.

jackfischer:
Would also make reporting minimized JS/TS related bugs easier.

scotttrinh:
> Would also make reporting minimized JS/TS related bugs easier.

Great idea, I hadn't considered having a "minimal" preset/template here, but that makes sense. Could even have it provide a simple `index.ts` that makes a trivial query.

PR #809, opened 2023-12-14
Add an initializer package for quickly creating a new EdgeDB project

scotttrinh:
For now, this just sets up a _super_ barebones project structure and is missing all of the most important parts of setting up each framework. However, this is working well enough to demonstrate the basic idea and we can now build "recipes" in parallel for each framework.

Note:

Going to explore the latest module syntax in this package since I think we need to update it everywhere else to support `node16` style module imports. So far, so good.

https://github.com/edgedb/edgedb-js/assets/1682194/9bccbda3-50eb-4a2b-a439-eebc4724fb90

image


scotttrinh:
Partially addresses #807 , but I don't want to close that issue when this is merged since there is a lot we want to do before calling that issue "done".

packages/create/src/recipes/base/_eslint.config.js
```diff
@@ -0,0 +1,24 @@
+import { FlatCompat } from "@eslint/eslintrc";
+import js from "@eslint/js";
+
+const compat = new FlatCompat();
+
+export default [
```
scotttrinh, line 6: Spicy: going to use the new "flat config" style of ESLint to avoid having to do a big overhaul in the future once ESLint 9 lands with this style on by default.

Open to anyone saying this is a bad idea 😂

PR #810, opened 2023-12-14
Add auth remix

diksipav:


packages/auth-remix/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "files": [
+    "/dist"
+  ],
+  "scripts": {
+    "build": "tsc --project tsconfig.json"
+  },
+  "devDependencies": {
+    "@types/node": "^20.8.4",
+    "edgedb": "^1.3.6",
+    "tailwindcss": "^3.3.5",
```
jaclarke, line 22:
```suggestion
```
I think this dependency was only needed for the example.

packages/auth-remix/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "typescript": "^5.2.2"
+  },
+  "peerDependencies": {
+    "edgedb": "^1.3.6"
+  },
+  "dependencies": {
+    "@edgedb/auth-core": "^0.1.0-beta.1",
+    "@octokit/core": "^5.0.2",
+    "@remix-run/node": "^2.3.1",
```
jaclarke, line 31:
```suggestion
```
I don't think either of these deps are needed now either.

packages/auth-remix/src/client.ts
```diff
@@ -0,0 +1,78 @@
...
+export function createClientAuth(options: RemixAuthOptions) {
+  return new RemixClientAuth(options);
+}
+
+export class RemixClientAuth {
+  protected readonly options: Required<
+    Omit<RemixAuthOptions, OptionalOptions>
+  > &
+    Pick<RemixAuthOptions, OptionalOptions>;
+
```
jaclarke, line 23:
```suggestion

/** @internal */
```

packages/auth-remix/src/client.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!this.authToken) return false;
+    try {
+      return await this.client.querySingle<boolean>(
+        `select exists global ext::auth::ClientTokenIdentity`
+      );
+    } catch {
+      return false;
+    }
+  }
+}
```
jaclarke, line 78: I don't think this needs to be part of the client exports? The edgedb client can only be used on the server side anyway.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+  type emailPasswordProviderName,
+} from "@edgedb/auth-core";
+
+import {
+  type RemixAuthOptions,
+  RemixClientAuth,
+  RemixAuthSession,
+} from "./client";
+import { Octokit } from "@octokit/core";
+import { redirect, json } from "@remix-run/node";
```
jaclarke, line 16:
```suggestion
import { redirect, json } from "@remix-run/server-runtime";
```
Octokit isn't needed, and I think it's better to use imports from remix's shared server-runtime package, rather than the runtime specific `@remix-run/node` package.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+    >
+  ): Promise<Response>;
+  onSignout(): Promise<Response>;
+}
+
+export class RemixServerAuth extends RemixClientAuth {
+  private readonly client: Client;
+  private readonly core: Promise<Auth>;
+
+  constructor(client: Client, options: RemixAuthOptions) {
```
jaclarke, line 64:
```suggestion
  /** @internal */
  constructor(client: Client, options: RemixAuthOptions) {
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+
+            return redirect(
+              pkceSession.getOAuthUrl(
+                provider,
+                redirectUrl,
+                `${redirectUrl}?isSignUp=true`
+              ),
+              {
+                headers: new Headers({
+                  "Set-Cookie": `${this.options.pkceVerifierCookieName}=${pkceSession.verifier}; HttpOnly;`,
```
jaclarke, line 132: Maybe we should use a cookie lib like this one: https://www.npmjs.com/package/cookie, instead of manually building the strings here?

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+            headers.append(
+              "Set-Cookie",
+              `${this.options.authCookieName}=${tokenData.auth_token}; HttpOnly; SameSite=Lax; Path=/`
+            );
+            headers.append(
+              "Set-Cookie",
+              `${this.options.pkceVerifierCookieName}=`
+            );
+
+            return oathCallbackCall(onOAuthCallback, headers, {
```
jaclarke, line 186: Maybe this function should have a different name? since it's not just used for oauth.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+              });
+            }
+            const code = searchParams.get("code");
+            const verificationEmailSentAt = searchParams.get(
+              "verification_email_sent_at"
+            );
+
+            if (!code) {
+              if (verificationEmailSentAt) {
+                return onBuiltinUICallback({
```
jaclarke, line 215: If we're checking that the responses returned to callbacks are always redirects, we need to wrap all these callback calls in that check.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+              provider: searchParams.get("provider") as BuiltinProviderNames,
+              isSignUp,
+            });
+          }
+          case "builtin/signin":
+          case "builtin/signup": {
+            const pkceSession = await this.core.then((core) =>
+              core.createPKCESession()
+            );
+            return new Response(null, {
```
jaclarke, line 266: Could use the `redirect` helper function here as well.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+  }
+
+  async emailPasswordSignUp(
+    req: Request,
+    data?: { email: string; password: string }
+  ): Promise<{ tokenData: TokenData; headers: Headers }>;
+  async emailPasswordSignUp<Res extends Response>(
+    req: Request,
+    cb: (params: ParamsOrError<{ tokenData: TokenData }>) => Res | Promise<Res>
+  ): Promise<Res>;
```
jaclarke, line 353: Since we're following remix's behaviour, where if you return something that isn't a response, we wrap it in `json()`, these overloads need to be updated to:

```suggestion
  async emailPasswordSignUp<Res>(
    req: Request,
    cb: (params: ParamsOrError<{ tokenData: TokenData }>) => Res | Promise<Res>
  ): Promise<Res extends Response ? Res : TypedResponse<Res>>;
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+
+        return { tokenData: null, headers };
+      },
+      req,
+      dataOrCb,
+      cb
+    );
+  }
+
+  async emailPasswordResendVerificationEmail(
```
jaclarke, line 415: This action doesn't technically need the request object, but maybe it should still have it for consistency?

jaclarke: Also when we fix this: https://github.com/edgedb/edgedb/issues/6503, then the action will need to set the pkce verifier cookie, so we should probably just make this action match the others (with the callback overloads, etc) now anyway.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+
+        if (!verifier) {
+          throw new Error("no pkce verifier cookie found");
+        }
+
+        let resetToken = new URL(req.url).searchParams.get("reset_token");
+
+        if (Array.isArray(resetToken)) {
+          resetToken = resetToken[0];
+        }
```
jaclarke, line 582: Don't think this is needed, `searchParams.get` always returns `string | null`. It's only an array if you use `searchParams.getAll`

jaclarke: Also why is the `reset_token` in the searchParams and not the body?

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+      },
+      req,
+      dataOrCb,
+      cb
+    );
+  }
+
+  async emailPasswordSendPasswordResetEmail(
+    req: Request,
+    data?: { email: string; password: string }
```
jaclarke, line 484:
```suggestion
    data?: { email: string; }
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+
+        return { tokenData };
+      },
+      req,
+      dataOrCb,
+      cb
+    );
+  }
+
+  async signout(
```
jaclarke, line 612:
```suggestion
  async signout(): Promise<{headers: Headers}>
  async signout<Res>(
    cb: () => Res | Promise<Res>
  ): Promise<Res extends Response ? Res : TypedResponse<Res>>
  async signout(
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+        select ext::auth::EmailFactor filter .identity = identity
+      ).email
+    } unless conflict on .identity`,
+      {
+        identity_id: tokenData.identity_id,
+        username: username,
+      }
+    );
+  }
+}
```
jaclarke, line 650:
```suggestion
```
This is only needed in the example.

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+  const cookies = request.headers.get("Cookie");
+
+  return cookies
+    ? cookies.split(";").reduce((cookies, cookie) => {
+        const [name, val] = cookie.split("=");
+        cookies[name.trim()] = val.trim();
+        return cookies;
+      }, {} as { [key: string]: string | undefined })
+    : {};
+}
```
jaclarke, line 662: As above, we should probably replace this with a proper cookie parsing library like:  https://www.npmjs.com/package/cookie

packages/auth-remix/package.json
diksipav: hey I use redirect & json from @remix-run/node, but deleted `@remix-run/server-runtime` since I don't use it anywhere anymore

diksipav: ah I saw your comment below about using @remix-run/server-runtime instead of /node, will check it and update

packages/auth-remix/src/server.ts
diksipav: yes, do u think its better to use authRouteHandlerCbCall, and for the other one to use cbCall instead of callbackCall?

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,795 @@
...
+      },
+      req,
+      dataOrCb,
+      cb
+    );
+  }
+
+  async emailPasswordResetPassword(
+    req: Request,
+    data?: { email: string; password: string }
```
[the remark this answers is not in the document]
diksipav: @jaclarke do we need reset_token here if we get it from url search paarms?

packages/auth-remix/src/server.ts
diksipav: I updated this to use reset_token from the data

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,863 @@
...
+              throw new Error(
+                `'onOAuthCallback' auth route handler not configured`
+              );
+            }
+            const error = searchParams.get("error");
+            if (error) {
+              const desc = searchParams.get("error_description");
+              return onOAuthCallback({
+                error: new Error(error + (desc ? `: ${desc}` : "")),
+              });
```
jaclarke, line 169:
```suggestion
              return cbCall(onOAuthCallback, {
                error: new Error(error + (desc ? `: ${desc}` : "")),
              });
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,863 @@
...
+            }
+            const code = searchParams.get("code");
+            const isSignUp = searchParams.get("isSignUp") === "true";
+            const verifier =
+              parseCookies(req)[this.options.pkceVerifierCookieName];
+            if (!code) {
+              return onOAuthCallback({
+                error: new Error("no pkce code in response"),
+              });
```
jaclarke, line 178:
```suggestion
              return cbCall(onOAuthCallback, {
                error: new Error("no pkce code in response"),
              });
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,863 @@
...
+                tokenData.auth_token,
+                { httpOnly: true, sameSite: "lax", path: "/" }
+              )
+            );
+            headers.append(
+              "Set-Cookie",
+              cookie.serialize(this.options.pkceVerifierCookieName, "", {
+                maxAge: 0,
+              })
+            );
```
jaclarke, line 207: I think all 'Set-Cookie' headers should have `path: "/"`.

diksipav: ``If the server omits the Path attribute, the user agent will use the "directory" of the request-uri's path component as the default value.`` 

Hm I agree

jaclarke (approved):

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+    `select exists ext::auth::UIConfig`
+  );
+
+  const session = auth.getSession(request);
+  const isLoggedIn = await session.isLoggedIn();
+  const username = isLoggedIn
+    ? (await session.client.querySingle<string>(
+        `select global currentUser.name`
+      )) || ""
+    : "";
```
scotttrinh, line 123: This is a bit of a personal nit, but I think it makes sense to fallback to `null` instead of the empty string, and maybe assume the `name` property is required to make this less ambiguous ("is it empty because i'm not logged in, or because it's set to the empty string?")

```suggestion
  const username = isLoggedIn
    ? await session.client.queryRequiredSingle<string>(
        `select global currentUser.name`
      )
    : null;
```

If you think it makes more sense to keep `name` optional, I'd still suggest the `null` rather than empty string as the fallback, so:

```suggestion
  const username = isLoggedIn
    ? await session.client.querySingle<string>(
        `select global currentUser.name`
      )
    : null;
```

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+export default function Index() {
+  const { builtinUIUrl, isLoggedIn, username } = useLoaderData<typeof loader>();
+
+  return (
+    <main>
+      <h1>Home</h1>
+      {isLoggedIn ? (
+        <>
+          <h2>
+            You are logged in as<span>{username}</span>
```
scotttrinh, line 145:
```suggestion
You are logged in{username ? ` as <span>${username}</span>` : ""
```

—?

scotttrinh (approved):
Readme looks good! Small suggestion around the `name` fallback.

diksipav:
I couldn't fix the default imports, I got into other errors when tried to implement Scott suggestion. Will work on that later today.

packages/auth-remix/readme.md
diksipav: Hey @scotttrinh, I think it's good to make username required and to fallback to null, but just to be sure I got this, the end result in your suggestion is the same for empty string or null? So using null is more like a good practice preference or for someone debugging, we show here to the user `You are logged in ` in case username is null but somehow user is logged in.

scotttrinh: Yeah, the result is the same, I was just handling the case where the name was not required.

If I was doing this in a real app, I would make some discriminated union that helped the type checker know that if a user was logged in, the name is not null, but not necessary for this example I think.

scotttrinh: Btw, the suggestion was initially to fix the missing space before the name but I ended up also suggesting changes for optional name.

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+   });
+   ```
+
+   The currently available auth handlers are:
+
+   - `onOAuthCallback`
+   - `onBuiltinUICallback`
+   - `onEmailPasswordSignIn`
+   - `onEmailPasswordSignUp`
+   - `onEmailPasswordReset`
```
jaclarke, line 73:
```suggestion
```
These auth route handlers don't exist. They are actions instead.

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+   - `emailPasswordResendVerificationEmail`
+   - `emailPasswordSendPasswordResetEmail`
+   - `emailPasswordResetPassword`
+   - `signout`
+   - `isPasswordResetTokenValid(resetToken: string)`: Checks if a password reset token is still valid.
+   - `getOAuthUrl(providerName: string)`: This method takes the name of an OAuth provider (make sure you configure providers you need in the auth ext config first using CLI or EdgeDB UI) and returns a link that will initiate the OAuth sign in flow for that provider. You will also need to configure the `onOAuthCallback` auth route handler.
+
+## Usage
+
+Now you have auth all configured and user's can signin/signup/etc. you can use the `auth.getSession()` method in your app pages to retrieve an `AuthSession` object. This session object allows you to check if the user is currently logged in with the `isLoggedIn` method, and also provides a `Client` object automatically configured with the `ext::auth::client_token` global, so you can run queries using the `ext::auth::ClientTokenIdentity` of the currently signed in user.
```
jaclarke, line 102:
```suggestion
Now you have auth all configured and user's can signin/signup/etc. you can use the `auth.getSession()` method in your app pages to retrieve an `AuthSession` object. This session object allows you to check if the user is currently logged in with the `isSignedIn` method, and also provides a `Client` object automatically configured with the `ext::auth::client_token` global, so you can run queries using the `ext::auth::ClientTokenIdentity` of the currently signed in user.
```

packages/auth-remix/src/server.ts
```diff
@@ -0,0 +1,874 @@
...
+  public readonly client: Client;
+
+  /** @internal */
+  constructor(client: Client, private readonly authToken: string | undefined) {
+    this.client = this.authToken
+      ? client.withGlobals({ "ext::auth::client_token": this.authToken })
+      : client;
+  }
+
+  async isLoggedIn() {
```
jaclarke, line 40:
```suggestion
  async isSignedIn() {
```

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+import { client, auth } from "~/services/auth.server";
+import { transformSearchParams } from "~/utils";
+
+export const loader = async ({ request }: LoaderFunctionArgs) => {
+  const builtinUIEnabled = await client.queryRequiredSingle<boolean>(
+    `select exists ext::auth::UIConfig`
+  );
+
+  const session = auth.getSession(request);
+  const isLoggedIn = await session.isLoggedIn();
```
jaclarke, line 118:
```suggestion
  const isLoggedIn = await session.isSignedIn();
```

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+  const username = isLoggedIn
+    ? (await session.client.querySingle<string>(
+        `select global currentUser.name`
```
jaclarke, line 121: I know this is just an example, but could it be confusing to use a custom global here? since `currentUser` isn't part of the auth ext schema, but looks like it could be?

jaclarke: Also presumably in this case, checking `isLoggedIn` first would be redundant, as `currentUser` would be null if not signed in.

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+    : "";
+
+  let params = new URL(request.url).searchParams;
+
+  return json({
+    builtinUIEnabled,
+    builtinUIUrl: auth.getBuiltinUIUrl(),
+    isLoggedIn,
+    username,
+    params: transformSearchParams(params),
```
jaclarke, line 132: `params` doesn't seem to be used anywhere, and if you did want to use search params in the component, I think the preferred way would be to use the `useSearchParams()` hook.

packages/auth-remix/readme.md
```diff
@@ -0,0 +1,157 @@
...
+    ? (await session.client.querySingle<string>(
+        `select global currentUser.name`
+      )) || ""
+    : "";
+
+  let params = new URL(request.url).searchParams;
+
+  return json({
+    builtinUIEnabled,
+    builtinUIUrl: auth.getBuiltinUIUrl(),
```
jaclarke, line 129: Instead of passing the `builtinUIUrl` from the loader, you can also use the `getBuiltinUIUrl` method on the client auth object directly in the component.

packages/auth-remix/readme.md
diksipav: yes yes, I copy pasted :/

scotttrinh (approved):
Looks good! Thanks for the extra effort on getting the release action working 🙏

PR #812, opened 2023-12-15
Removed unneeded async keyword

PastelStoic:
Functions are currently generated using the async keyword, despite not running async. They return a promise directly. This change slightly reduces the size of generated code and stops linters from yelling about unneeded async.


scotttrinh (approved):
Looks good. FWIW, I think consumers of codegen should try to avoid having their linting/formatting tools on generated code, but that's certainly just a suggestion.

PR #814, opened 2023-12-15
Update the `next-auth` readme to add details about pages router support + api reference

jaclarke:


diksipav:
Hey @jaclarke I’m still a bit confused, the things we have in the client are methods on the client class, and then also on the server class so we call them with `auth.getBuiltinUIUrl` for example , I still don’t get a complete idea why would someone need these links without full functionality… 

I wrote a readme, pls check it (I mostly copied the flow and info that you provided in next example).
@scotttrinh added more detailed info for Express https://github.com/edgedb/edgedb-js/blob/master/packages/auth-express/readme.md. 

Should I also go into more details explaining (explaining error handling, that callbacks should preferably return response/redirect etc).

PR #820, opened 2024-01-04
Add remix create recipe

diksipav:
Remix template files and index file that copies template files into the newly created project and create/update this project's package.json.


packages/create/src/recipes/remix/index.ts
```diff
@@ -0,0 +1,56 @@
...
+      type: "module",
+      scripts: {
+        dev: "remix dev --manual",
+        build: "remix build",
+        lint: "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
+        start: "remix-serve ./build/index.js",
+        typecheck: "tsc",
+      },
+      dependencies: {
+        ...(useEdgeDBAuth ? { "@edgedb/auth-nextjs": "^0.1.0-beta.1" } : {}),
```
scotttrinh, line 31:
```suggestion
        ...(useEdgeDBAuth ? { "@edgedb/auth-remix": "^0.1.0-beta.1" } : {}),
```

Also, I wonder if there is a version specifier that will install whatever version is actually published 🤔

scotttrinh (changes requested):

packages/create/src/recipes/remix/index.ts
diksipav: oh this already got merged ( I enabled auto merge), will create another PR now

PR #826, opened 2024-01-04
Add express recipe for `@edgedb/create`

scotttrinh:
Borrowing liberally from our example app (https://github.com/edgedb/edgedb-examples/tree/main/express-auth) making it much simpler.

Do we think it's worth having the `todos` part of the API defined here, or maybe we should just have a single route that does `select 42;` since this is _not_ an example but rather a starter for an actual app? @raddevon any thoughts on that since we talked a bunch about it in the context of the example app?


raddevon:
> Do we think it's worth having the `todos` part of the API defined here, or maybe we should just have a single route that does `select 42;` since this is _not_ an example but rather a starter for an actual app?

What about a single `select 42;` route along with a comment giving the URL to the example app?

raddevon (approved):

PR #838, opened 2024-01-16
Add an example of `select(update())` to the docs

### LINKED ISSUE #837: Allow data return on update (Similar to postgresql/prisma)

Implement an option to return the updated document or fields, similar to postgres.
```
UPDATE courses
SET published_date = '2020-07-01'
WHERE course_id = 2
RETURNING *;
```

I know EdgeQL allow for such scenario like this:
```
select (
  update posts
  filter .id = <uuid>"abd01c82-b3eb-11ee-9242-1788c2aec42a"
  set {
    like_counter := .like_counter + 1
  }
) {
  id,
  like_counter
}
```

I find myself updating counters and other various fields and having to query the database again. It's quite cumbersome and annoying.

The point I remember reading about EdgeDB is that a lot run in the DB itself, but we still need to run two distinctive queries for a "simple" update. :thinking: 

**Suggested way of doing this:**
```
const updateLikes = await e
    .update(e.posts, (seq) => ({
        filter: e.op(seq.id, "=", e.uuid(testid)),
        set: {
            like_counter: e.op(seq.like_counter, "+", 1),
        },
        returns: {
            like_counter: true
        },
    })).run(client);
```
would return
```
[
  {
    "id": "f0e8fd98-b3cd-11ee-a1c8-2b46a3bed554", // Right now, the ID is always return, lets keep it that way
    "like_counter": 5
  }
]
```

There could also be a way to _splat_ things up
```
const updateLikes = await e
    .update(e.posts, (seq) => ({
        filter: e.op(seq.id, "=", e.uuid(testid)),
        set: {
            like_counter: e.op(seq.like_counter, "+", 1),
        },
        returns: {
            '*': true
        },
    })).run(client);
```

Current time taken on localhost:
```
Total            195 msec
Update value     190 msec
Fetch new value  5 msec
```

If there is already a way of doing this, please consider updating the docs :smile:

scotttrinh:
Closes #837

### DIFF
--- a/docs/update.rst
+++ b/docs/update.rst
@@ -41,6 +41,24 @@ You can conditionally update a property by using an :ref:`optional parameter
     }))
   );
 
+Note that ``e.update`` will return just the `{ id: true }` of the updated object. If you want to select further properties, you can wrap the update in a `e.select` call.
+
+.. code-block:: typescript
+
+  e.params({ id: e.uuid, title: e.optional(e.str) }, (params) =>
+    e.select(
+      e.update(e.Movie, (movie) => ({
+        filter_single: { id: params.id },
+        set: {
+          title: e.op(params.title, "??", movie.title),
+        },
+      })),
+      (movie) => ({
+        title: movie.title,
+      }),
+    ),
+  );
+
 
 Updating links
 ^^^^^^^^^^^^^^
--- a/docs/update.rst
+++ b/docs/update.rst
@@ -41,23 +41,22 @@ You can conditionally update a property by using an :ref:`optional parameter
     }))
   );
 
-Note that ``e.update`` will return just the `{ id: true }` of the updated object. If you want to select further properties, you can wrap the update in a `e.select` call.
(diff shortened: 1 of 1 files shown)


docs/update.rst
```diff
@@ -41,23 +41,22 @@ You can conditionally update a property by using an :ref:`optional parameter
     }))
   );
 
-Note that ``e.update`` will return just the `{ id: true }` of the updated object. If you want to select further properties, you can wrap the update in a `e.select` call.
+Note that ``e.update`` will return just the `{ id: true }` of the updated object. If you want to select further properties, you can wrap the update in a `e.select` call. This is still just a single query to the database.
```
raddevon, line 44:
```suggestion
Note that ``e.update`` will return just the ``{ id: true }`` of the updated object. If you want to select further properties, you can wrap the update in a ``e.select`` call. This is still just a single query to the database.
```

raddevon (approved):
Needs double backticks on some inline code, but aside from that, it looks great!

PR #839, opened 2024-01-22
Add edgedb auth helper for svelte projects

diksipav:


packages/auth-sveltekit/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@edgedb/auth-sveltekit",
+  "description": "Helper library to integrate the EdgeDB Auth extension with Sveltekit.",
+  "version": "0.1.0-alpha.1",
+  "author": "EdgeDB <info@edgedb.com>",
+  "type": "module",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/edgedb/edgedb-js.git",
+    "directory": "packages/auth-svelte"
```
jaclarke, line 10:
```suggestion
    "directory": "packages/auth-sveltekit"
```

packages/auth-sveltekit/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "typescript": "^5.2.2"
+  },
+  "peerDependencies": {
+    "edgedb": "^1.3.6"
+  },
+  "dependencies": {
+    "@edgedb/auth-core": "^0.1.0-beta.1",
+    "@sveltejs/kit": "^2.0.0",
+    "svelte": "^4.2.7",
+    "vite": "^5.0.3",
```
jaclarke, line 32: `svelte` and `vite` should probably be dev dependencies, as they're only needed to make sveltekit's postinstall script happy, and I don't think we want to define deps here that could conflict with the user's installed versions.

jaclarke: Also not sure if `@sveltejs/kit` should be a peer dependency (+ dev dep) instead? @scotttrinh

scotttrinh: Yeah, I that's how we structured the others: https://github.com/edgedb/edgedb-js/blob/316a398f9dbb413a880440086df7b5bebc7e68b5/packages/auth-nextjs/package.json#L23-L35. I think the only dependency should be on `@edgedb/auth-core` itself.

packages/auth-sveltekit/readme.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+2. Initialize the server auth helper by passing an EdgeDB `Client` object to `createServerAuth()`, along with configuration options. This will return a `SvelteServerAuth` object which you can use across your app on the server side.
+
+   ```ts
+   // src/lib/server/auth.ts
+
+   import createServerAuth from "@edgedb/auth-sveltekit/server";
+   import { createClient } from "edgedb";
+   import { options } from "./auth.client";
```
jaclarke, line 44:
```suggestion
   import { options } from "$lib/auth";
```

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,735 @@
+import {
+  redirect,
+  type Cookies,
+  type RequestEvent,
+  type ActionFailure,
+} from "@sveltejs/kit";
+import * as cookie from "cookie";
```
jaclarke, line 7: I'm not sure this is needed? Sveltekit looks like it has it's own cookie handling functions, and the only place this seems to be used is in the `getSession` which I think could take a `RequestEvent` object instead, that has a `cookies` object.

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,735 @@
...
+    dataOrCb?:
+      | { email: string; password: string }
+      | ((params: ParamsOrError<{ tokenData: TokenData }>) => CallbackReturn),
+    cb?: (params: ParamsOrError<{ tokenData: TokenData }>) => CallbackReturn
+  ): Promise<
+    | {
+        tokenData: TokenData;
+      }
+    | CallbackReturn
+  > {
```
jaclarke, line 385: I don't think we need to copy the all of this response wrapping code for the form actions from remix. As far as I can tell sveltekit is closer to nextjs, in that it lets you set cookies via the `cookies` object on the `RequestEvent` independently of what the user returns from the action function. So these form actions can be simplified to just taking the `RequestEvent` and any data they need and directly returning the `tokenData` (or whatever)/thowing an error.

packages/auth-sveltekit/src/server.ts
diksipav: oh yes, I forgot to refactor this part, tnx

packages/auth-sveltekit/src/server.ts
diksipav: I agree about RequestEvent, but are we sure we don't wanna support callbacks?

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+  const core = Auth.create(client);
+  const config = getConfig(options);
+
+  return {
+    createServerRequestAuth: ({ event }: { event: RequestEvent }) =>
+      new ServerRequestAuth(client, core, event, options),
+    createAuthRouteHook:
+      (handlers: AuthRouteHandlers): Handle =>
+      ({ event, resolve }) => {
+        const pathname = new URL(event.request.url).pathname;
```
jaclarke, line 70: The `RequestEvent` object already has a URL object, so this could just be:
```suggestion
        const pathname = event.url.pathname;
```

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+
+        if (pathname.startsWith(`/${config.authRoutesPath}`)) {
```
jaclarke, line 72:
```suggestion
        if (pathname.startsWith(`/${config.authRoutesPath}/`)) {
```
Just in case the user has some other route that starts with the word 'auth' (or whatever they've set the auth route as).

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+  private readonly client: Client;
+  private readonly core: Promise<Auth>;
+  private readonly cookies: Cookies;
+
+  public get session() {
+    return new AuthSession(
+      this.client,
+      this.cookies.get(this.config.authCookieName)
+    );
+  }
```
jaclarke, line 91: Now this class is initialised per request, this can just be a property created in the constructor.

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+    return (await this.core).getProvidersInfo();
+  }
+
+  async emailPasswordSignUp({
+    email,
+    password,
+  }: {
+    email: string;
+    password: string;
+  }): Promise<{ tokenData?: TokenData | null }> {
```
jaclarke, line 121: Did we intentionally drop the `{email: string, password: string} | FormData` thing, where you can just pass the formData object directly to these methods?

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+  return {
+    createServerRequestAuth: ({ event }: { event: RequestEvent }) =>
+      new ServerRequestAuth(client, core, event, options),
+    createAuthRouteHook:
+      (handlers: AuthRouteHandlers): Handle =>
+      ({ event, resolve }) => {
+        const pathname = new URL(event.request.url).pathname;
+
+        if (pathname.startsWith(`/${config.authRoutesPath}`)) {
+          return createAuthRouteHandlers(handlers, event, core, config);
```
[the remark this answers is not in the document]
jaclarke: Actually, the handlers should always call `redirect`, so this should actually be something like this I think?
```ts
const noMatchingRoute = Symbol();

// ...

if (await createAuthRouteHandlers(handlers, event, core, config) !== noMatchingRoute) {
  throw new Error('Auth route handler should always call redirect()');
}

// ...

function createAuthRouteHandlers() {
  switch (path) {
    // ...
    case default:
      return noMatchingRoute;
  }
}
```

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+  | ({ error: Error } & ErrorDetails & { [Key in keyof Result]?: undefined });
+
+export interface AuthRouteHandlers {
+  onOAuthCallback(
+    params: ParamsOrError<{
+      tokenData: TokenData;
+      provider: BuiltinOAuthProviderNames;
+      isSignUp: boolean;
+    }>
+  ): Promise<Response>;
```
jaclarke, line 39:
```suggestion
  ): Promise<never>;
```
I think these handlers all should throw a redirect(), so the return type should be `never`.

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+      return await this.client.querySingle<boolean>(
+        `select exists global ext::auth::ClientTokenIdentity`
+      );
+    } catch {
+      return false;
+    }
+  }
+}
+
+async function createAuthRouteHandlers(
```
jaclarke, line 258: While we're being pedantic with naming, this should probably be something like `handleAuthRoutes`.

packages/auth-sveltekit/src/server.ts
```diff
@@ -0,0 +1,489 @@
...
+    onOAuthCallback,
+    onBuiltinUICallback,
+    onEmailVerify,
+    onSignout,
+  }: Partial<AuthRouteHandlers>,
+  event: RequestEvent,
+  core: Promise<Auth>,
+  config: AuthConfig
+) {
+  const url = new URL(event.request.url);
```
scotttrinh, line 269: I guess we can check if this is also already a `URL` object?

```suggestion
  const { url } = event.request;
```

scotttrinh (commented):
Everything is looking good! Let's update the README and this should be ready to merge.

packages/auth-sveltekit/src/server.ts
scotttrinh: This does make it lazy though, right? If you don't call `auth.session` it wouldn't create the class?

jaclarke: Yes, but then you create it everytime you access session though. Ok, maybe it's fine as it is.

scotttrinh: Oh yeah I guess we can set it as an optional private property and return it if it's defined.

jaclarke: Or, if we want to slightly optimise it, we could memoise it. I think I did that for auth-nextjs.

packages/auth-sveltekit/src/server.ts
diksipav:
```
async function handleAuthRoutes(
  ...
  { url, cookies }: RequestEvent,
  ...
) {
  const searchParams = url.searchParams;
  const path = url.pathname.split("/").slice(2).join("/");
  ...
```

packages/auth-sveltekit/src/server.ts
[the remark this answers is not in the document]
diksipav: Yes, this is a great catch! Tnx.

packages/auth-sveltekit/src/server.ts
diksipav: I guess I removed FormData support because supabase and lucia didnt have it too, added it back.

packages/auth-sveltekit/src/server.ts
diksipav: hm is this ok?

```
const sessionCache: Record<string, AuthSession> = {};

export class ServerRequestAuth extends ClientAuth {
  private readonly client: Client;
  private readonly core: Promise<Auth>;
  private readonly cookies: Cookies;

  public get session() {
    const authCookie = this.cookies.get(this.config.authCookieName);

    if (authCookie && sessionCache[authCookie]) {
      return sessionCache[authCookie];
    }

    const authSession = new AuthSession(this.client, authCookie);

    if (authCookie) {
      sessionCache[authCookie] = authSession;
    }

    return authSession;
  }
```

packages/auth-sveltekit/readme.md
```diff
@@ -0,0 +1,161 @@
...
+The currently available auth route handlers are:
+
+- `onOAuthCallback`
+- `onBuiltinUICallback`
+- `onEmailVerify`
+- `onSignout`
+
+In any of them you can define what to do in case of success or error. Every handler should return a redirect call.
+
+5. Now we just need to setup the UI to allow your users to sign in/up, etc. The easiest way to get started is to use the EdgeDB Auth's builtin UI. Or alternatively you can implement your own custom UI.
```
jaclarke, line 106:
```suggestion
4. Now we just need to setup the UI to allow your users to sign in/up, etc. The easiest way to get started is to use the EdgeDB Auth's builtin UI. Or alternatively you can implement your own custom UI.
```

packages/auth-sveltekit/readme.md
```diff
@@ -0,0 +1,161 @@
...
+// src/lib/server/auth.ts
+import { createClient } from "edgedb";
+
+export const client = createClient({
+  //Note: when developing locally you will need to set tls  security to insecure, because the dev server uses  self-signed certificates which will cause api calls with the fetch api to fail.
+  tlsSecurity: "insecure",
+});
+```
+
+3. Create the server auth client in a handle hook. Firstly call `serverAuth` passing to it EdgeDB client you created in the previous step, along with configuration options from step 1. This will give you back the `createServerRequestAuth` and `createAuthRouteHook`. You should call `createServerRequestAuth` inside a handle to attach the server client to `event.locals`. Calling `createAuthRouteHook` will give you back a hook so you should just invoke it inside `sequence` and pass your auth route handlers to it.
```
jaclarke, line 57: This step also needs the user to add a `.d.ts` file with the type of `locals.auth` (if they're using typescript). https://github.com/lucia-auth/examples/blob/main/sveltekit/username-and-password/src/app.d.ts#L5

Issue #845, opened 2024-02-02
The `resendVerificationEmail` server action in `auth-nextjs` should accept `email` param.


MiroslavPetrik:
**Describe the bug**
The [emailPasswordResendVerificationEmail() server action](https://github.com/edgedb/edgedb-js/blob/fd0be5f1d31afecfe442e1083d52598c92232677/packages/auth-nextjs/src/app/index.ts#L137) should work with `email` param which is accepted on [the backend side.](https://github.com/edgedb/edgedb/blob/639c91e9207275c114828556a1b00c4a3029d8c1/edb/server/protocol/auth_ext/http.py#L624).

Currently only the `verification_token` is accepted. 

A case for the email param is, when the user is already signed in, and the verification is optional, I still want to display a button in user's profile to resend the verification email. In that case I don't have the `verification_token`, maybe the user also removed the original email where the old token was present.

**Additional context**

scotttrinh:
Thanks for opening this issue it really is the completion of the job we did in https://github.com/edgedb/edgedb/pull/6723.

I'll stick this into the todo list for now, but PRs welcome if you want to take a swing at it!

MiroslavPetrik:
@scotttrinh I've opened a PR. Would appreciate a review, as the params extraction for OR case is not possible, so I had to nest try catch, but its ordered, but maybe it could be split into 2 separate actions instead. :)

PR #846, opened 2024-02-02
Enable the resendVerificationEmail Next.js action to work with email …

MiroslavPetrik:
…param. (#845)

I've tested locally that the server endpoint work just with the `email` param:

```ts
    const url = new URL(
      "resend-verification-email",
      "http://localhost:10705/db/edgedb/ext/auth/",
    ).toString();

    const res = await fetch(url, {
      method: "post",
      body: JSON.stringify({
        provider: "builtin::local_emailpassword",
        email: formData.get("email"),
      }),
      headers: { "Content-Type": "application/json" },
    });
```

### DIFF
--- a/packages/auth-core/src/core.ts
+++ b/packages/auth-core/src/core.ts
@@ -145,6 +145,13 @@ export class Auth {
     });
   }
 
+  async resendVerificationEmailForEmail(email: string) {
+    await this._post("resend-verification-email", {
+      provider: emailPasswordProviderName,
+      email,
+    });
+  }
+
   async sendPasswordResetEmail(email: string, resetUrl: string) {
     const { challenge, verifier } = await pkce.createVerifierChallengePair();
     return {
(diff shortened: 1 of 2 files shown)


packages/auth-core/src/core.ts
```diff
@@ -145,6 +145,13 @@ export class Auth {
     });
   }
 
+  async resendVerificationEmailForEmail(email: string) {
```
MiroslavPetrik, line 148: I think have this separate makes for cleaner code.

packages/auth-nextjs/src/app/index.ts
```diff
@@ -135,14 +135,33 @@ export class NextAppAuth extends NextAuth {
         return tokenData;
       },
       emailPasswordResendVerificationEmail: async (
-        data: FormData | { verification_token: string }
+        data: FormData | { verification_token: string } | { email: string }
```
MiroslavPetrik, line 138: I kept this in single action, as adding another action would require more docs changes, and potentially would be confusing to have 2 calls for 1 endpoint

packages/auth-core/src/core.ts
scotttrinh, line 148: I thought a bit about how we might be able to support this format directly on `resendVerificationEmail`, but I think I prefer the direction you've taken this here by making it just a separate function. Moves the complexity into the consumer.

scotttrinh (approved):
Thanks for the quick turnaround here 🙏

PR #847, opened 2024-02-02
Pass `verify_url` in the `resendVerificationEmail` Next.js server action

MiroslavPetrik:
https://github.com/edgedb/edgedb/blob/639c91e9207275c114828556a1b00c4a3029d8c1/edb/server/protocol/auth_ext/http.py#L612C9-L612C20

### DIFF
--- a/packages/auth-core/src/core.ts
+++ b/packages/auth-core/src/core.ts
@@ -145,10 +145,11 @@ export class Auth {
     });
   }
 
-  async resendVerificationEmailForEmail(email: string) {
+  async resendVerificationEmailForEmail(email: string, verifyUrl: string) {
     await this._post("resend-verification-email", {
       provider: emailPasswordProviderName,
       email,
+      verify_url: verifyUrl,
     });
   }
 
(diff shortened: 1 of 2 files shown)


MiroslavPetrik:
@scotttrinh I forgot this 😹

packages/auth-core/src/core.ts
```diff
@@ -138,17 +138,19 @@ export class Auth {
     return this.getToken(code, verifier);
   }
 
-  async resendVerificationEmail(verificationToken: string) {
+  async resendVerificationEmail(verificationToken: string, verifyUrl: string) {
```
MiroslavPetrik, line 141: This is shared also in express.
Now I wonder if the verifyUrl should be optional? @scotttrinh 

Or add it everywhere where it fails?

scotttrinh: We don't pass `verifyUrl` when you have a `verificationToken` since the URL is contained within that signed token, so I think you can just remove it for this function.

scotttrinh (approved):

PR #853, opened 2024-02-06
Add README.md files for @edgedb/create recipes

beerose:


packages/create/src/recipes/express/README.md
```diff
@@ -0,0 +1,51 @@
+# Express Integration Recipe for EdgeDB
+
+This recipe provides a starting point for building an [Express](https://expressjs.com/) application with EdgeDB as the database.
+
+We try to actively monitor and incorporate significant changes from the original "create-app" templates to to ensure developers have access to the latest features and best practices.
```
scotttrinh, line 5: FWIW, in this specific case, we aren't using any existing "create-app" as the starting point since I was unable to find anything that felt like a canonical starter for express.

packages/create/src/recipes/nextjs/README.md
```diff
@@ -0,0 +1,51 @@
+# NextJS Integration Recipe for EdgeDB
```
scotttrinh, line 1: I believe they use "Next.js" as their official name formatting

scotttrinh (changes requested):

packages/create/src/recipes/express/README.md
beerose: Ok, makes sense to remove it then

scotttrinh (approved):

PR #857, opened 2024-02-08
Add SvelteKit create tool

diksipav:


packages/create/src/recipes/sveltekit/template/ts/src/lib/server/auth.__auth.ts
```diff
@@ -0,0 +1,6 @@
+import { createClient } from "edgedb";
+
+export const client = createClient({
+  //Note: when developing locally you will need to set tls  security to insecure, because the dev server uses  self-signed certificates which will cause api calls with the fetch api to fail.
```
jaclarke, line 4:
```suggestion
  // Note: when developing locally you will need to set tls security to insecure, because the dev server uses self-signed certificates which will cause api calls with the fetch api to fail.
```

packages/create/src/recipes/sveltekit/template/ts/src/lib/server/auth.__auth.ts
```diff
@@ -0,0 +1,6 @@
+import { createClient } from "edgedb";
```
jaclarke, line 1: Maybe this file should be named something else? It only creates the edgedb client, so it's not specific to using auth, and is useful to include in the base svelte template. So maybe `lib/server/edgedb.ts`?

packages/create/src/recipes/sveltekit/template/ts/src/hooks.server.__auth.ts
```diff
@@ -0,0 +1,45 @@
...
+    redirect(303, "/");
+  },
+};
+
+const { createServerRequestAuth, createAuthRouteHook } = serverAuth(
+  client,
+  options
+);
+
+const createServerAuthClient: Handle = ({ event, resolve }) => {
```
jaclarke, line 36: Can we inline this function in the `sequence` call? It avoids needing to come up with a name for this function. 😄

jaclarke (commented):
I think it would also be useful to include a simple example of using the auth locals stuff, so something like adding a `The user is/isn't logged in` paragraph on the index page, maybe with a link to the builtin ui login page. Just so you can have an immediately working auth example when you create the project.

packages/create/src/recipes/sveltekit/template/ts/src/hooks.server.__auth.ts
diksipav: I think it can be cleaner if function is separate but if u think name is not the best I can inline it.

packages/create/src/recipes/sveltekit/template/ts/src/lib/server/auth.__auth.ts
diksipav: yes, we used to create auth client here, will rename it

diksipav:
> I think it would also be useful to include a simple example of using the auth locals stuff, so something like adding a 'The user is/isn't logged in' paragraph on the index page, maybe with a link to the builtin ui login page. Just so you can have an immediately working auth example when you create the project.

How is the builting ui login page possible if we don't enable and configure it firstly?

jaclarke (approved):

PR #858, opened 2024-02-09
Add optional param example

corbinday:


docs/parameters.rst
```diff
@@ -78,6 +78,20 @@ A type can be made optional with the ``e.optional`` function.
   // or without duration
   const result = await query.run(client, {title: 'The Eternals'});
 
+If you are using the queries generator, add the ``optional`` keyword within a type cast:
```
scotttrinh, line 81: These docs are for the query builder, so I think a better place for this would be in the queries generator docs: https://github.com/edgedb/edgedb-js/blob/master/docs/queries.rst

Feel free to update the example there to use a `with` block and show an optional parameter.

scotttrinh (changes requested):

docs/parameters.rst
corbinday: How's this? It didn't make as much sense to me to put the same content I suggested in `parameters.rst` in `queries.rst`. So instead I just opted to point to the links you mentioned earlier, since a reference to them here would have been useful to me when I was trying to figure out optional parameters.

I also added a caption and changed the type of the code block for the edgeql example file that was already mentioned to make it stand out a little more.

corbinday:
Depends on https://github.com/edgedb/edgedb/pull/6808

docs/queries.rst
```diff
@@ -262,3 +263,14 @@ To exclude the generated files, add the following lines to your ``.gitignore`` f
...
   dbschema/queries.*
+
+Writing Queries with Parameters
+----------------
+To inject external values into your EdgeQL queries, you can use :ref:`parameters <ref_eql_params>`. 
+
+When using the queries generator, you may be tempted to declare the same parameter in multiple places. 
+However, it's better practice to declare it once by assigning it to a variable in a :ref:`with block <ref_eql_with_params>`
+and reference that variable in the rest of your query. This way you avoid mismatched types in your declarations, 
+such as forgetting to mark them all as :ref:`optional<ref_eql_optional_params>`.
```
raddevon, line 274: Relative linking like this will work until https://github.com/edgedb/edgedb/issues/6803 is fixed.

```suggestion
such as forgetting to mark them all as `optional </docs/edgeql/parameters#optional-parameters>`__.
```

docs/queries.rst
```diff
@@ -262,3 +263,14 @@ To exclude the generated files, add the following lines to your ``.gitignore`` f
 
   **/*.edgeql.ts
   dbschema/queries.*
+
+Writing Queries with Parameters
+----------------
```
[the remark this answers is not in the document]
corbinday: Oh good to know. I had a different title at once point and then I forgot to update the underline.

corbinday:
@raddevon Ok, I updated the other link as well.

raddevon:
Great! Thank you! I think your other two links may also have to use relative paths since they are across repos as well.

corbinday:
Oh whoops. You're right. I'll fix that

raddevon (approved):
This looks great! Thank you!

Issue #866, opened 2024-02-15
Error visibility in query file generation


jackfischer:
**Describe the proposed feature**
A clear and concise description of the syntax, what you want to happen, and why it's needed.

Query files generated from `@edgedb/generate queries` seem to generate in alphabetical order. With several hundred queries, any errors from the particular file you're iterating on are buried somewhere in pages of output. It would be helpful if errors could somehow be surfaced at the bottom of the output. Maybe in conjunction with https://github.com/edgedb/edgedb-js/issues/865, write all the output for successful queries first and then all errors at the end.

scotttrinh:
I have a crazy idea: what if the output was [TAP](https://testanything.org/)? That way users would be able to use [one of the many TAP reporters](https://www.npmjs.com/search?ranking=popularity&q=tap%20reporter) to format the output here? Then you could also have a different output format for CI, etc.

Feels like a bit of a stretch to use a testing-specific format for this, but when I was thinking about this problem from a data structure perspective (`({ path } | { path, error })[]`) it started to look suspiciously like TAP.

jackfischer:
Wasn't familiar with TAP but indeed that data structure sounds exactly right!

PR #874, opened 2024-02-21
Facelift for Queries Generator error logging (#866)

themajashurka:
Just a little proposal to improve visibility on the outcome of the generator.

### DIFF
--- a/packages/generate/src/queries.ts
+++ b/packages/generate/src/queries.ts
@@ -4,6 +4,12 @@ import { type CommandOptions } from "./commandutil";
 import { headerComment } from "./genutil";
 import { type Target, camelify } from "./genutil";
 
+function collectAndPrintError(errs:string[], path:string, err:any){
+  const message = `Error in file '${path}': ${err.toString()}`
+  errs.push(message)
+  console.log(`\x1b[31m${message}\x1b[0m`);
+}
+
 // generate per-file queries
 // generate queries in a single file
 // generate per-file queries, then listen for changes and update
@@ -13,6 +19,7 @@ export async function generateQueryFiles(params: {
   client: Client;
   schemaDir: string;
 }) {
+  const t0 = performance.now()
   if (params.options.file && params.options.watch) {
     throw new Error(`Using --watch and --file mode simultaneously is not
 currently supported.`);
@@ -38,6 +45,21 @@ currently supported.`);
 
   console.log(`Detected schema directory: ${params.schemaDir}`);
   const matches = await getMatches(root, params.schemaDir);
+
+  const errs:string[] = [];
+  function printSummary(){
+    const passedCount = matches.length - errs.length
+    const summary = `\n\x1b[${errs.length > 0 ? '1;31m' : '1;32m'}${passedCount}/${matches.length} queries passed${errs.length ? ', '+errs.length+' failed'  : '!'}`
+    console.log(summary)
+    for (const err of errs) {
+        console.log('\x1b[22m')
+        console.error(err)
+    }
+    if(errs.length > 1) console.log(summary)
+    const t1Pretty = ((performance.now()-t0)/1000).toFixed(2)
+    console.log(`\x1b[1;33mCompleted in: ${t1Pretty}s\x1b[0m\n`)
+  }
+
   if (matches.length === 0) {
     console.log(`No .edgeql files found outside of ${params.schemaDir}`);
     return;
@@ -80,9 +102,7 @@ currently supported.`);
           }
         } catch (err) {
           wasError = true;
-          console.log(
-            `Error in file '${prettyPath}': ${(err as Error).toString()}`
-          );
+          collectAndPrintError(errs, prettyPath, err)
         }
       })
     );
@@ -109,6 +129,7 @@ currently supported.`);
         );
       }
     }
+    printSummary();
     return;
   }
 
@@ -131,11 +152,7 @@ currently supported.`);
         );
       }
     } catch (err) {
-      console.log(
-        `Error in file './${adapter.path.posix.relative(root, path)}': ${(
-          err as any
-        ).toString()}`
-      );
+      collectAndPrintError(errs, './'+adapter.path.posix.relative(root, path), err)
     }
   }
 
@@ -146,6 +163,8 @@ currently supported.`);
   // find all *.edgeql files
   // for query in queries:
   //   generate output file
+
+  printSummary();
 }
 
 export function stringifyImports(imports: { [k: string]: boolean }) {


packages/generate/src/queries.ts
```diff
@@ -38,6 +45,21 @@ currently supported.`);
 
   console.log(`Detected schema directory: ${params.schemaDir}`);
   const matches = await getMatches(root, params.schemaDir);
+
+  const errs:string[] = [];
```
scotttrinh, line 49: I won't point out each formatting issue, but if you don't mind running `yarn run format`, we use the default prettier rules for formatting.

packages/generate/src/queries.ts
```diff
@@ -38,6 +45,21 @@ currently supported.`);
...
+  function printSummary(){
+    const passedCount = matches.length - errs.length
+    const summary = `\n\x1b[${errs.length > 0 ? '1;31m' : '1;32m'}${passedCount}/${matches.length} queries passed${errs.length ? ', '+errs.length+' failed'  : '!'}`
+    console.log(summary)
+    for (const err of errs) {
+        console.log('\x1b[22m')
+        console.error(err)
```
scotttrinh, line 56: Does this work as expected? `console.log` outputs to `stdout` and `console.error` outputs to `stderr`.

packages/generate/src/queries.ts
```diff
@@ -38,6 +45,21 @@ currently supported.`);
...
+    }
+    if(errs.length > 1) console.log(summary)
```
scotttrinh, line 58: Did you mean to log the summary twice? Seems like unconditionally logging here at the end is just as good?

packages/generate/src/queries.ts
```diff
@@ -38,6 +45,21 @@ currently supported.`);
 
   console.log(`Detected schema directory: ${params.schemaDir}`);
   const matches = await getMatches(root, params.schemaDir);
+
+  const errs:string[] = [];
+  function printSummary(){
+    const passedCount = matches.length - errs.length
```
scotttrinh, line 51: We prefer a more functional style in general. So instead of this function being a closure within this other function, it could be a top level function that receives `matches` and `errs` as arguments and either prints directly or returns a string that we `console.log` at the end.

packages/generate/src/queries.ts
```diff
@@ -4,6 +4,12 @@ import { type CommandOptions } from "./commandutil";
 import { headerComment } from "./genutil";
 import { type Target, camelify } from "./genutil";
 
+function collectAndPrintError(errs:string[], path:string, err:any){
+  const message = `Error in file '${path}': ${err.toString()}`
+  errs.push(message)
+  console.log(`\x1b[31m${message}\x1b[0m`);
+}
```
scotttrinh, line 11: Instead of mutating the input array, I'd prefer this function just formats and returns a string, and leave it up to the caller to figure out what to do with the string.

Something like:

```suggestion
function formatError(path: string, err: any){
  const message = `Error in file '${path}': ${err.toString()}`
  return `\x1b[31m${message}\x1b[0m`;
}
```

packages/generate/src/queries.ts
```diff
@@ -80,9 +102,7 @@ currently supported.`);
           }
         } catch (err) {
           wasError = true;
-          console.log(
-            `Error in file '${prettyPath}': ${(err as Error).toString()}`
-          );
+          collectAndPrintError(errs, prettyPath, err)
```
scotttrinh, line 105: Given the updates here, we'd change this to:

```suggestion
          errs.push(formatError(prettyPath, err));
```

packages/generate/src/queries.ts
```diff
@@ -131,11 +152,7 @@ currently supported.`);
         );
       }
     } catch (err) {
-      console.log(
-        `Error in file './${adapter.path.posix.relative(root, path)}': ${(
-          err as any
-        ).toString()}`
-      );
+      collectAndPrintError(errs, './'+adapter.path.posix.relative(root, path), err)
```
scotttrinh, line 155: And similarly:

```suggestion
      errs.push(
        formatErrors(
          './' + adapter.path.posix.relative(root, path),
          err
        )
      );
```

scotttrinh (changes requested):
Thanks for jumping in here! Most of my feedback is just formatting and idiom/style things, I like the general shape.

Do you mind also including a screenshot for some review and feedback?

packages/generate/src/queries.ts
themajashurka: My bad. I was literally just throwing in lines to make things work. Next time (if that will be a thing) I pay attention to that!

packages/generate/src/queries.ts
themajashurka: I was literally doing that the first time, closures are embedded in my muscle memory, sorry :)

packages/generate/src/queries.ts
themajashurka: I've fixed this with the latest commit.

themajashurka:
Three screenshots for indicating the feature:

Summary of unsuccessful generation at the bottom
edgedb_js_#866_1

Summary of successful generation at the bottom
edgedb_js_#866_2

"Inline error" coloring for unsuccessful generation
edgedb_js_#866_3

packages/generate/src/queries.ts
scotttrinh: I see your update, but what I had in mind was always logging the summary at the end and never logging it at the beginning.

So:

```ts
console.log(summary);
```

In other words, what I had in mind for the final output here is:

1. Printing each file as their being evaluated, capturing any errors.
2. At the end, printing all of the errors.
3. Print the summary.

What do you think?

packages/generate/src/queries.ts
scotttrinh: I'd still like to see this moved out into the top scope of this module (right after `formatError`) and have it take the `matches` and `errs` as arguments.

packages/generate/src/queries.ts
scotttrinh: Still needs a quick pass with the formatter.

scotttrinh (changes requested):

themajashurka:
My latest experience with this change, is that now it feels kind of redundant to print error messages twice: "inline" and aggregated at the bottom. Maybe it's enough to indicate "inline" errors simply with red color (and giving the detailed error in summary)?

_Before_
before

_After_
after

scotttrinh:
@themajashurka 
> Maybe it's enough to indicate "inline" errors simply with red color (and giving the detailed error in summary)?

Yeah, that looks great!

scotttrinh:
@themajashurka 

> And one more thing: to reduce messiness, put a separator in between error messages?

Yeah, either a separator or just a new line works for me, I trust your judgement here.

themajashurka:
There are still a couple of changes needed to be make in order to make `formatError` and stuff work. Plus I found an unrelated&small bug.
I'll have more time on the weekend to make these changes!

themajashurka:
Sorry to ask it here, but is there an industry standard way of keeping a link to my forked `edgedb-js` in my edgedb projects? I keep jumping back and forth between my forked repo and my local `node_modules` to see the effect of changes.

scotttrinh:
> Sorry to ask it here, but is there an industry standard way of keeping a link to my forked edgedb-js in my edgedb projects?

You can use [`yarn link`](https://classic.yarnpkg.com/en/docs/cli/link#search) to link a dependency to a local build.

scotttrinh:
Going to close this for now just the clean out my review queue. feel free to keep working on it if you have time, or let me know if you won't be able to and I can stick it back in my todo queue.

PR #875, opened 2024-02-22
feat(nuxt): add nuxt template settings

Tahul:
Adds a base of recipe for Nuxt that implements auth and basic blogpost features.

Can be tested by running the create CLI locally using `Nuxt` template.

I'm not sure about what is expected for such create template, feel free to trim anything extraneous.


packages/create/src/recipes/nextjs/index.ts
```diff
@@ -102,7 +102,7 @@ const recipe: Recipe<NextjsOptions> = {
         lint: "next lint",
       },
       dependencies: {
-        ...(useEdgeDBAuth ? { "@edgedb/auth-nextjs": "^0.1.0" } : {}),
+        ...(useEdgeDBAuth ? { "nuxt-edgedb-module": "^0.1.0" } : {}),
```
scotttrinh, line 105: I think this is an accidental change, right?

packages/create/src/recipes/nuxt/index.ts
```diff
@@ -0,0 +1,50 @@
...
+        build: "nuxi build",
+        start: "nuxi start",
+        generate: "nuxi generate"
+      },
+      dependencies: {
+        "@iconify-json/heroicons": "1.1.19",
+        "@nuxt/ui": "^2.13.0"
+      },
+      devDependencies: {
+        "nuxt-edgedb-module": "latest",
```
scotttrinh, line 41: I imagine this is a runtime dependency not just a dev dependency, right? Also, can we use a more strict version range? Maybe `^0.0.42`?

packages/create/src/recipes/nuxt/template/dbschema/default.esdl
```diff
@@ -0,0 +1,38 @@
+using extension auth;
+
+module default {
```
scotttrinh, line 3: In general, I think we should have these recipes be starting point for any kind of app, not really an example app, if that makes sense. I'd expect that the schema here is empty, and that we add the auth extension here conditionally based on options set by the user.

To that end, if you want to design a small single page that just scaffolds some _very_ basic layout and styles, definitely do that. Probably something closer to https://github.com/nuxt/starter/tree/v3 .

scotttrinh (changes requested):

packages/create/src/recipes/nextjs/index.ts
Tahul: Oh yeah it is; 🤦‍♂️

packages/create/src/recipes/nuxt/template/dbschema/default.esdl
Tahul: Sure, sounds great.

I started doing the "most minimal" Nuxt starter, but for Nuxt it just looks like adding `nuxt-edgedb-module` to their modules array.

The rest of it is just described in [`nuxt-edgedb-module`](https://github.com/Tahul/nuxt-edgedb-module).

Should I push that instead?

packages/create/src/recipes/nuxt/index.ts
Tahul: Sure, it has runtime as well.

packages/create/src/recipes/nuxt/template/dbschema/default.esdl
scotttrinh: Sorry for the delay in getting back to you here.

I'm not sure I understand the question but what I expect to see in these recipes is a totally empty `default.esdl` file, not even including the `using extension auth;` expression even. No migrations, no query builder code generator output, etc. These recipes should represent how you'd start day 0 on a new project, not as an example project, if that makes sense. The UI/app itself should also be minimal so that developers can quickly check that everything is working right, but nothing else.

Does that make sense?

PR #883, opened 2024-03-01
Fix order by types to allow enums

CarsonF:


scotttrinh (changes requested):
Do you mind adding a simple test for this? I think you can add an `order_by` to https://github.com/edgedb/edgedb-js/blob/18f0351c6dc9374432db9bc005b9f43f1fb7ca3d/integration-tests/lts/select.test.ts#L1049-L1057 and switch it to `movie.genre`.

CarsonF:
That ok?

scotttrinh (approved):

PR #884, opened 2024-03-01
Accept readonly arrays on assignment

CarsonF:
This is safe because it's saying we expect less.
Rather than giving less on output, which could be a breaking change.

### DIFF
--- a/packages/generate/src/syntax/casting.ts
+++ b/packages/generate/src/syntax/casting.ts
@@ -109,7 +109,7 @@ export type setToAssignmentExpression<
 type getAssignmentLiteral<
   Set extends PrimitiveTypeSet,
   IsSetModifier extends boolean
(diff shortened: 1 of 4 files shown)


packages/generate/src/syntax/casting.ts
```diff
@@ -113,11 +113,11 @@ type getAssignmentLiteral<
   ?
       | TsType
       | (Set["__cardinality__"] extends Cardinality.Many
-          ? TsType[]
+          ? readonly TsType[]
```
scotttrinh, line 116: Can you add a simple test to https://github.com/edgedb/edgedb-js/blob/master/integration-tests/lts/casts.test.ts to cover both `readonly` and mutable arrays?

scotttrinh (changes requested):

scotttrinh (approved):
🙏 Thanks for the tests!

PR #885, opened 2024-03-01
Break down analyzeQuery for more generation flexibility

CarsonF:
I'm trying to get immutable output types, without duplicating a bunch of this code.

### DIFF
--- a/packages/driver/src/reflection/index.ts
+++ b/packages/driver/src/reflection/index.ts
@@ -30,4 +30,4 @@ export * from "./reservedKeywords";
 // export * from "../syntax/funcops";
 // export * from "./syntax";
 export * as introspect from "./queries";
-export { analyzeQuery } from "./analyzeQuery";
+export * from "./analyzeQuery";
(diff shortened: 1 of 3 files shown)


packages/driver/src/reflection/index.ts
```diff
@@ -30,4 +30,4 @@ export * from "./reservedKeywords";
 // export * from "../syntax/funcops";
 // export * from "./syntax";
 export * as introspect from "./queries";
-export * from "./analyzeQuery";
+export { analyzeQuery } from "./analyzeQuery";
```
scotttrinh, line 33: Was this change necessary or are you being defensive here to avoid leaking the new exports?

scotttrinh (approved):
Looks good!

packages/driver/src/reflection/index.ts
CarsonF: Defensive. But I forgot until now that reflection/index isn't exported at top level. So many exporting these functions just to reflection is ok. Thoughts?

scotttrinh: I don't have a strong opinion on this, but I guess it's safest to be restrictive now and add more to the export if we want them in the future, so good to keep what you have here.

CarsonF: Ok I adjusted names to better fit a more generalized export/import.

The more I look at this the more I wonder why this is in the driver package and not the generation one...

scotttrinh (approved):
Dig it. 👍

PR #888, opened 2024-03-04
Wrap types in default module in `$default` namespace and re-export them at top level

### LINKED ISSUE #882: If modules have types that have the same name, the interfaces generator creates invalid references to the default module.
Given the SDL:

```esdl
module default {
  type Foo {}
}

module Bar {
  type Foo extends default::Foo {}
}
```

The interfaces generator will try to reference the `default::Foo` type without any sort of qualifier, so it ends up with something like:

```ts
namespace Bar {
  export interface Foo extends Foo {}
}
```

jaclarke:
Fixes #882

### DIFF
--- a/packages/generate/src/edgeql-js/generateInterfaces.ts
+++ b/packages/generate/src/edgeql-js/generateInterfaces.ts
@@ -1,4 +1,4 @@
-import { CodeBuffer, js, t } from "../builders";
+import { CodeBuffer, t } from "../builders";
 import type { GeneratorParams } from "../genutil";
 import { $ } from "../genutil";
 import { makePlainIdent, quote, splitName, toTSScalarType } from "../genutil";
@@ -32,8 +32,7 @@ export const generateInterfaces = (params: GenerateInterfacesParams) => {
       const modName = modParts[modParts.length - 1];
       const parentModName = modParts.slice(0, -1).join("::");
       const parent = parentModName ? getModule(parentModName) : null;
-      const internalName =
-        modName === "default" && !parentModName ? "" : makePlainIdent(modName);
+      const internalName = makePlainIdent(modName);
 
       module = {
         name: modName,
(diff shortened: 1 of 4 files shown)


packages/generate/src/edgeql-js/generateInterfaces.ts
```diff
@@ -178,12 +176,8 @@ export const generateInterfaces = (params: GenerateInterfacesParams) => {
   const plainTypesExportBuf = new CodeBuffer();
 
   const writeModuleExports = (module: ModuleData) => {
-    const wrapInNamespace = !(module.isRoot && module.name === "default");
-    if (wrapInNamespace) {
-      plainTypesCode.writeln([t`export namespace ${module.internalName} {`]);
-      plainTypesCode.writeln([js`const ${module.internalName} = {`]);
```
scotttrinh, line 184: Why are we dropping the `js` writing here? Looking at the changes here seems like it was only ever writing an empty object with the module name so seems safe, just curious really.

packages/generate/src/edgeql-js/generateInterfaces.ts
```diff
@@ -200,11 +194,17 @@ export const generateInterfaces = (params: GenerateInterfacesParams) => {
...
+    plainTypesCode.writeln([t`}`]);
+    plainTypesCode.addExport(module.internalName, { modes: ["js"] });
+
+    if (module.isRoot && module.name === "default") {
+      for (const [typeName, typeRef] of module.types) {
+        plainTypesCode.writeln([`import ${typeName} = ${typeRef};`]);
+      }
+      plainTypesCode.writeln([
+        `export {${[...module.types.keys()].join(", ")}};`,
+      ]);
```
scotttrinh, line 207: Seems like this is already covered by the existing tests in `interfaces.test.ts`, but do you think it'd be easy to write a quick check for the pathological case we're fixing here?

packages/generate/src/edgeql-js/generateInterfaces.ts
jaclarke: I was just removing it as it seems to be unused, since the interfaces file only contains types, and there's no runtime code. And it looks like the ouput mode is hardcoded to only ever output the `ts` tagged code anyway: https://github.com/edgedb/edgedb-js/blob/master/packages/generate/src/interfaces.ts#L58-L62.

scotttrinh (approved):
👏 

Thanks for taking this on!

PR #889, opened 2024-03-04
Refactor codec to TS generation for fine grained customization

CarsonF:
- Moved `$.parseQuery` -> `Client.parse`. This feels better as it doesn't have to assume as much regarding client pool, options, or session.
  Perhaps a name like `$parse` or `parse$` would be desired though?
- Added `$.util.parseCardinality` as a standard way to convert from the protocol number cardinality to the string enum used by the query builder.
  Happy to move it where ever.
- `parse` returns a string enum `Cardinality` now rather than the protocol one.
  This was the only spot the protocol cardinality leaked into type gen and it felt inconsistent in my (advanced) usage.
- Along those lines I removed the hardcoded protocol cardinality constants the codecs were using in favor of the cardinality enum.

And then the big refactor of `walkCodec`, now `generateTSTypeFromCodec`.
I broke down the logic for how to generate each codec into their own functions.
Now `generateTSTypeFromCodec` just orchestrates calling these methods (optionally customized via input map).

### DIFF
--- a/integration-tests/lts/dbschema/queries/deepArrayInput.edgeql
+++ b/integration-tests/lts/dbschema/queries/deepArrayInput.edgeql
@@ -1,0 +1,1 @@
+select <array<tuple<str, str>>>$deep;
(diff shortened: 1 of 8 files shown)


scotttrinh (approved):
Love the refactor of `walkCodec` here, and moving the Cardinality parsing to the edge to avoid leaking the number constants is a nice touch, too. 💯

PR #891, opened 2024-03-06
Add Auth to Remix create & small fixes

diksipav:
- I updated Readmes to use `yarn run create` instead of `yarn create` when running locally because yarn create is used for creating new projects from create starter kits, it will not run our create scripts but it expects create-starter-kit as an argument. 

Maybe I used something wrong, if this works on your side @jaclarke @scotttrinh I can revert this changes.

- Added auth files to Remix template since it was missing it.

- Updated Remix and Sveltekit index pages to show link to builtin UI if enabled.

### DIFF
--- a/packages/create/src/recipes/remix/index.ts
+++ b/packages/create/src/recipes/remix/index.ts
@@ -15,7 +15,16 @@ const recipe: Recipe = {
     logger("Running remix recipe");
 
     const dirname = path.dirname(new URL(import.meta.url).pathname);
-    await copyTemplateFiles(path.resolve(dirname, "./template"), projectDir);
+
+    let tags;
+
+    if (useEdgeDBAuth) {
+      tags = new Set<string>(["auth"]);
+    }
+
+    await copyTemplateFiles(path.resolve(dirname, "./template"), projectDir, {
+      tags,
+    });
 
     await updatePackage(projectDir, {
       sideEffects: false,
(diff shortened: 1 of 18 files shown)


packages/create/src/recipes/remix/index.ts
```diff
@@ -41,10 +50,15 @@ const recipe: Recipe = {
...
         "@types/react-dom": "^18.2.7",
+        "@typescript-eslint/eslint-plugin": "^6.7.4",
+        "@typescript-eslint/parser": "^6.7.4",
+        eslint: "^8.38.0",
+        "eslint-import-resolver-typescript": "^3.6.1",
         "eslint-plugin-import": "^2.28.1",
         "eslint-plugin-jsx-a11y": "^6.7.1",
         "eslint-plugin-react": "^7.33.2",
         "eslint-plugin-react-hooks": "^4.6.0",
+        typescript: "^5.1.6",
```
scotttrinh, line 61: Didn't check the other new dependencies, but is there a reason why we're referring to an older version of `typescript` here?

packages/create/src/recipes/remix/template/.eslintrc.cjs
```diff
@@ -43,10 +43,13 @@ module.exports = {
...
           { name: "NavLink", linkAttribute: "to" },
         ],
+        "import/resolver": {
+          typescript: {},
+        },
       },
     },
 
-    // TypeScript
+    // Typescript
```
scotttrinh, line 52: TypeScript is correct. (see https://www.typescriptlang.org/)

scotttrinh (approved):

packages/create/src/recipes/remix/template/.eslintrc.cjs
diksipav: ah ok, this is changed because newest Remix create will create this, so I copied their newest  .eslintrc.cjs

scotttrinh: 😅 Fine to keep it then, not a big deal.

packages/create/src/recipes/remix/index.ts
diksipav: not sure if sth can break in the future if we use "*" in package.json

scotttrinh: Yeah, that's a good question. Let's just follow their version ranges for now.

PR #896, opened 2024-03-07
Expose connection params via Client

CarsonF:
I don't love the method or type alias names. Open to better/different ones.


scotttrinh (approved):
Yeah, feels like this is the right time to expose this config as a readonly public interface. We already depend on it in a [separate first party package](https://github.com/edgedb/edgedb-js/blob/3676ed2fcbc01ca2a277ebf506d3ddccddf3f4a0/packages/auth-core/src/core.ts#L34-L36), so might as well make this official.

PR #899, opened 2024-03-08
Add WebAuthn and Magic Links in auth-remix

scotttrinh:
Similar to #897 

Going to do some testing before opening this up for review, but it's a pretty straightforward port from the nextjs work.

### DIFF
--- a/packages/auth-core/src/errors.ts
+++ b/packages/auth-core/src/errors.ts
@@ -88,6 +88,13 @@ export class OAuthProviderFailureError extends UserError {
   }
 }
 
+/** Magic link flow failed. */
+export class MagicLinkFailureError extends UserError {
+  get type() {
+    return "MagicLinkFailure";
+  }
+}
+
 /** Error with email verification. */
 export class VerificationError extends UserError {
   get type() {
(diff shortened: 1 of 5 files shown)


JesseFarebro:
Tried this out and ran into an issue where calling `signupWithWebAuthn` results in this issue: https://github.com/remix-run/remix/issues/4414. Problem being that two requests are made in succession, 1) `webauthn/register` and if successful 2) `token`. The call to token is what results in a socket hang up.

Perhaps you could allow for additional headers to be passed through to the `fetch` call or keep-alive could be the default?

scotttrinh:
@JesseFarebro 

What version of EdgeDB Server are you running? We had a bug that was fixed in 5.0b2 where we were not setting the `Connection: close` header and just ending the connection that _also_ showed a `socket hang up` error but that was fixed by sending that (ostensibly optional) header in the server. If you're on an older beta, or an older nightly, can you try with the latest beta?

JesseFarebro:
Thanks for the prompt reply @scotttrinh. I'm on `5.0-beta.2+96c7381`, should I move to one of the nightly builds?

scotttrinh:
@JesseFarebro 

Oh, it looks like it hasn't gotten into a beta yet. I'll poke the release team to see when the next beta is released. It is definitely in the latest nightlies, but I don't want you to have to switch the nightly just to get this fix.

PR #900, opened 2024-03-09
Fix collision in QB with nested object & module names

CarsonF:
Probably need tests, but wanted to get this up before the weekend.

## Use full namespace in import

Previously these FQNs
```edgeql
A::X::Y
B::X::Y
```
were emitted as:
```ts
import _X from './A/X';
import _X from './B/X';
```

now they are emitted as:
```ts
import _AX from './A/X';
import _BX from './B/X';
```

I opted to use `indentRef.name` instead of path logic.
It just seemed easier, but I understand this is a deviation.
I'm open to changing.

## Merged nested objects with nested modules (just like top level)

Previously these FQNs
```edgeql
A::B
A::B::C
```
were emitted as:
```ts
type __defaultExports = {
  "B": typeof B;
  "B": typeof _module__B;
};
```
Now these types & values are merged, just as in `generateIndex`.

### DIFF
--- a/packages/generate/src/builders.ts
+++ b/packages/generate/src/builders.ts
@@ -488,12 +488,22 @@ class BuilderImportsExports {
       exports.push(`export type { ${exportTypes.join(", ")} };\n`);
     }
     if (refsDefault.length || forceDefaultExport) {
+      const refsByExportAs = [...groupToMapBy(refsDefault, (x) => x.as)].map(
+        ([as, group]) => ({
+          as,
+          refs: group.map(({ ref }) => ref).reverse(),
+        })
+      );
       if (mode === "ts" || mode === "dts") {
         exports.push(
           `${
             mode === "dts" ? "declare " : ""
-          }type __defaultExports = {\n${refsDefault
-            .map(({ ref, as }) => `  ${genutil.quote(as)}: typeof ${ref}`)
+          }type __defaultExports = {\n${refsByExportAs
+            .map(({ refs, as }) => {
+              const key = genutil.quote(as);
+              const value = refs.map((ref) => `typeof ${ref}`).join(" & ");
+              return `  ${key}: ${value}`;
+            })
             .join(";\n")}\n};`
         );
       }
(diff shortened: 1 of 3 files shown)


packages/generate/src/builders.ts
```diff
@@ -603,15 +622,15 @@ export class CodeBuilder {
...
+      const namespaceParts = identRef.name.split("::").slice(0, -1);
+      prefix = `_${namespaceParts.join("")}`;
 
       let importPath = path.posix.join(
         path.posix.relative(
           path.posix.dirname(this.dir),
           path.posix.dirname(ref.dir)
         ),
-        mod
+        namespaceParts.at(-1)!
```
scotttrinh, line 633: TIL: [`Array.prototype.at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at), nice.

Small nit: since we're referencing this and asserting it, it would be helpful to name this thing.

```ts
const mod = namespaceParts.at(-1) as string;
```

Something like that. wdyt?

scotttrinh (changes requested):
One blocking request: can we add a simple test maybe to https://github.com/edgedb/edgedb-js/blob/master/integration-tests/lts/objectTypes.test.ts to demonstrate the issue and the fix here?

packages/generate/src/builders.ts
CarsonF: > TIL: [`Array.prototype.at`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at), nice.

Had to double check if it was supported by all Node versions supported here lol


> Small nit: since we're referencing this and asserting it, it would be helpful to name this thing.

Sure. I wasn't sure what to call it..maybe that's why I inlined.
I didn't like mod, because is `mod` == `A::X` or `X`. Maybe `leafModule`?

`!` vs `as string`: I prefer `!` because from my experience it's safer. Safer because you are asserting less. It's only asserting that the type X is _not null_. Vs X is _not null and it's a string_. I can't think of a scenario of the top of my head, but I've had times where TS considers types to sufficiently overlap enough for the typecast be allowed and the error suppressed.

scotttrinh: > I prefer ! because from my experience it's safer.

That's fair, not a blocker, it just seemed weird to make a not-null assertion on a variable assignment rather than asserting the type directly. More from the reader's perspective than the type checkers perspective. We could also just assign it and move the assertion to where it's used so:

```ts
const mod = namespaceParts.at(-1);

// ...
  mod!
```

> I didn't like mod, because is mod == A::X or X. Maybe leafModule?

Sure! Or what about `moduleFileName`? Anything somewhat obvious here is fine by me, and the whole thing isn't a blocker, just something that made me pause for a second when reading through the code so thought it was worth bikeshedding for a bit on.

CarsonF: > We could also just assign it and move the assertion to where it's used so:

That's definitely not my preference. We know it's not nullable, so that's should be the type of the variable (even if it's inferred). Changing type in usage makes me think there are other code paths that could be nullable.
Also it's the closest to `at()` which is what introduces the nullability that we are suppressing.



> Or what about `moduleFileName`?

I like it.

CarsonF:
I added schema to produce the generation failures.
https://github.com/edgedb/edgedb-js/actions/runs/8240000012/job/22534457000?pr=900
CI passes in rebased commits.

scotttrinh (approved):

PR #901, opened 2024-03-09
Add req object to Nextjs auth callbacks

corbinday:


corbinday:
@scotttrinh 
My main motivation for this is to have access to the request cookies, but it made sense to me not to arbitrarily limit what gets passed in from the request object, so I just added the whole thing. Maybe other people will want access to other data in the request.

Happy to edit as needed.

packages/auth-nextjs/src/shared.ts
```diff
@@ -33,20 +33,21 @@ export interface CreateAuthRouteHandlers {
       tokenData: TokenData;
       provider: BuiltinOAuthProviderNames;
       isSignUp: boolean;
+      req: NextRequest;
```
scotttrinh, line 36: Hmm.. Putting the `req` object here means it would only be defined if there was not an error, but I could see a use case for wanting to access the request object even in the event of a failure.

What do you think about adding it as a second argument to these functions?

corbinday: I had thought of that, but I didn't want to ruin the simplicity of destructuring only the needed params. How's this?

scotttrinh: Yeah. We could also put it on the `ParamsOrError` type. Open to suggestion here if you think destructuring feels more ergonomic.

corbinday: Higher order functions are a little mind-bending. Does making req optional like this...

```typescript
onBuiltinUICallback(
    params: ParamsOrError<
      (
        | {
            tokenData: TokenData;
            provider: BuiltinProviderNames;
          }
        | {
            tokenData: null;
            provider: null;
          }
      ) & { isSignUp: boolean }
    >,
    req?: NextRequest
  ): Promise<never>;
  ```
  
  ...allow for optionally specifying req like this:
  ```typescript
  const { GET, POST } = auth.createAuthRouteHandlers({
    onBuiltinUICallback({ error }, req) {
      if (!!error) {
        console.log(error);
        redirect('/404');
      } else {
       consumeCookieData(req.cookies.someCookie);
        redirect('/profile');
      }
    }
});
  ```
  
  ...or not, like how it's been? Because if it doesn't have to be listed when defining the callbacks, that's plenty ergonomic for me. I was mostly interested in preserving the optionality of specifying req. It doesn't have to be destructuring.

scotttrinh: Yeah, the way you've defined it keeps it optional, the main drawback is that if we add _more_ arguments after that, they also have to be optional and you get into weird call signatures like `foo(bar, undefined, undefined, { beep: "Boop" });` and callbacks like `(_, __, ___, beep) => {}`. At the same time, the properties of the `ParamsOrError` parameter definitely feels fundamentally different than the original `Request` object, so maybe it still makes sense to put this out as an optional trailing parameter.

Let's stick with what you got here and if we find that we have a lot of other context we want to provide, we can address that as part of work toward a 1.0 breaking change.

corbinday: Sounds good to me!

PR #907, opened 2024-03-13
Create PKCE session for verification with email

scotttrinh:
Blocked by https://github.com/edgedb/edgedb/pull/7037

The current email verification flow always assumes that it ends in a PKCE code exchange. However, currently, if you send just the email, we do not start a PKCE flow, so it just ends in a redirect or `204 No Content` response. The above linked PR to the auth server allows specifying a `challenge` in the body along with the email which will trigger a PKCE flow.


scotttrinh:
Now that EdgeDB 4.7 is out, this is ready for release.

PR #918, opened 2024-03-23
Add basic reference docs for `createHttpClient`

scotttrinh:
We probably want to revisit the options docs, but at the very least, this adds documentation for the http client creation function.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -114,6 +114,101 @@ Client
         main();
 
 
+.. js:function:: createHttpClient(options?: string | ConnectOptions | null): Client
+
+    Creates an HTTP client for interacting with an EdgeDB instance from a browser or edge runtime environment. This function returns an instance of :js:class:`Client` configured to communicate over HTTP using the Fetch API.
+
+    :param options:
+        This is an optional parameter. When it is not specified the client
+        will connect to the current EdgeDB Project instance.
+
+        If this parameter is a string it can represent either a
+        DSN or an instance name:
+
+        * when the string does not start with ``edgedb://`` it is a
+          :ref:`name of an instance <ref_reference_connection_instance_name>`;
+
+        * otherwise it specifies a single string in the connection URI format:
+          ``edgedb://user:password@host:port/database?option=value``.
+
+          See the :ref:`Connection Parameters <ref_reference_connection>`
+          docs for full details.
+
+        Alternatively the parameter can be a ``ConnectOptions`` config;
+        see the documentation of valid options below.
+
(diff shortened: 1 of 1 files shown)


docs/reference.rst
```diff
@@ -114,6 +114,101 @@ Client
...
+
+    The above connection options can also be specified by their corresponding
+    environment variable. If none of ``dsn``, ``credentialsFile``, ``host`` or
+    ``port`` are explicitly specified, the client will connect to your
+    linked project instance, if it exists. For full details, see the
+    :ref:`Connection Parameters <ref_reference_connection>` docs.
+
+
+    :param number options.timeout:
+        Connection timeout in milliseconds.
```
raddevon, line 181: I don't think this got rendered as you intended, but I'm not sure if I'd call this a bug in the docs site, or if we just need to rework the documentation.

CleanShot 2024-03-25 at 11 12 33@2x

We get another "Arguments" heading before `options.timeout`. I think we would just want to continue the arguments list after the paragraph about connection options _without_ another heading. I can't think of a case where we would want the current behavior, but I wanted to check in with you and see what you think. I guess if there's an easy fix we can make here we should do that. I'm not sure there is though.

The easy "fix" would be indenting it, maybe as a note inside the description of one of the params, but it's relevant to multiple params so I don't think that really works here.

docs/reference.rst
```diff
@@ -114,6 +114,101 @@ Client
...
 
+.. js:function:: createHttpClient(options?: string | ConnectOptions | null): Client
+
+    Creates an HTTP client for interacting with an EdgeDB instance from a browser or edge runtime environment. This function returns an instance of :js:class:`Client` configured to communicate over HTTP using the Fetch API.
+
+    :param options:
+        This is an optional parameter. When it is not specified the client
+        will connect to the current EdgeDB Project instance.
+
+        If this parameter is a string it can represent either a
```
raddevon, line 125: The newlines here get collapsed. In the rendered output, "If" is directly against the preceding period with nothing between. I've started a thread about it on Slack. Seems like a bug in the docs.

raddevon (approved):
Looking this over again at the end and comparing to the `createClient` docs above it, I believe there's only one very minor change in here for you, but this may have revealed a couple of bugs in the new docs site.

docs/reference.rst
scotttrinh: FWIW, all of this is copied wholesale from the `createClient` docs, so any issues here are also issues there, I'd imagine. Definitely good with taking the time to update these docs both in layout and content though!

PR #920, opened 2024-03-25
Adjust parentheses on QB cast rendering to fix pathing off of it

### LINKED ISSUE #917: Invalid path from object casts
**Code**
The code causing the error.

```typescript
const lang = e.cast(e.Language, e.uuid(id));
const query = lang.engagements.id;
```
Produces
```edgeql
SELECT <default::Language>(<std::uuid>"...").engagements.id
```
The cast needs to be wrapped in parentheses to be valid before `.engagements`

A workaround is to wrap the cast expression in a `e.select`, and then path off of that.

CarsonF:
Fixes #917.
I'm not sure when the inner exp would need to be wrapped in parentheses before casting, but I removed that.
Now wrapping the whole expression.
Also removing the parentheses wrapping for multiple cast siblings.

### DIFF
--- a/packages/generate/src/syntax/toEdgeQL.ts
+++ b/packages/generate/src/syntax/toEdgeQL.ts
@@ -935,7 +935,12 @@ function renderEdgeQL(
     if (expr.__expr__ === null) {
       return `<${typeName}>{}`;
     }
-    return `<${typeName}>(${renderEdgeQL(expr.__expr__, ctx)})`;
+    const rawInnerExpr = renderEdgeQL(expr.__expr__, ctx);
+    const isCast =
+      (expr.__expr__ as any).__kind__ === ExpressionKind.Cast &&
+      rawInnerExpr[0] === "(";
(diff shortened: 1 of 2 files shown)


packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -932,10 +938,11 @@ function renderEdgeQL(
...
         : expr.__element__.__name__;
-    if (expr.__expr__ === null) {
-      return `<${typeName}>{}`;
-    }
-    return `<${typeName}>(${renderEdgeQL(expr.__expr__, ctx)})`;
+    const inner = expr.__expr__
+      ? `(${renderEdgeQL(expr.__expr__, ctx)})`
+      : `{}`;
+    // TODO OPTION: just always wrap casts
+    return `(<${typeName}>${inner})`;
```
scotttrinh, line 945: Yeah, I think it's fine to unconditionally wrap casts in parens for now. Simplest, least error prone, and I don't find it to be too unreadable. This was going to be my approach!

PR #923, opened 2024-03-25
Bump to latest TypeScript to get bad perf

scotttrinh:
TypeScript 5.4 has introduced a pretty gnarly type instantiation regression. I have made some progress on a separate branch, but I wanted to bump TS and update the benchmarks to reflect the current state to make it easier to measure the improvements in that upcoming PR.


scotttrinh:
Closed as fixed in #924

PR #927, opened 2024-03-29
Cache session serialized state for its lifetime

CarsonF:
Rather than only the most recent one used.

I've got a few states, and cycling between them causes them to be recalculated on every use.


scotttrinh (approved):
Ahhh... Makes sense. Nice find!

PR #938, opened 2024-04-08
Set cookies to `Secure` when base URL is `https`

scotttrinh:
Will follow-up with similar changes to the other frameworks.


packages/auth-nextjs/src/shared.ts
```diff
@@ -209,13 +226,7 @@ export abstract class NextAuth extends NextAuthHelpers {
                 req
               );
             }
-            cookies().set({
-              name: this.options.authCookieName,
-              value: tokenData.auth_token,
-              httpOnly: true,
-              sameSite: "lax",
```
scotttrinh, line 216: @jaclarke do you know/remember why we were setting this to lax explicitly?

PR #964, opened 2024-04-15
Add `@edgedb/ai` package for `ext::ai`

scotttrinh:
Provides a high-level library for use with the `ext::ai` extension.


packages/ai/src/core.ts
```diff
@@ -0,0 +1,114 @@
...
+    public readonly client: Client,
+    options: Partial<AIOptions> = {},
+    context: Partial<QueryContext> = {}
+  ) {
+    this.baseUrl = EdgeDBAI.getBaseUrl(client);
+    this.options = {
+      model: options.model ?? "gpt-3.5-turbo",
+      prompt: options.prompt ?? {
+        name: "builtin::rag-default",
+      },
```
jaclarke, line 27: I'm not sure we need these defaults. The prompt is optional anyway, as the server will pick a default prompt. And the model is dependent on which providers you have configured, so it doesn't seem safe to use an openai model as the default?

packages/ai/debug.mts
jaclarke, line 1: Maybe this should be moved to an example in a readme?

packages/ai/src/core.ts
scotttrinh: Oh, that's a good point. Will fix!

packages/ai/debug.mts
scotttrinh: Yeah, I can do that with this and the stream one. We should have full examples with servers in the example repo.

jaclarke (approved):

PR #965, opened 2024-04-16
Update queries generator to pull import path(s) from codecs

CarsonF:
This allows custom codecs to use imports from non-"edgedb" exports

### DIFF
--- a/packages/driver/src/reflection/analyzeQuery.ts
+++ b/packages/driver/src/reflection/analyzeQuery.ts
@@ -17,7 +17,9 @@ type QueryType = {
   result: string;
   cardinality: Cardinality;
   query: string;
-  imports: ImportMap;
+  importMap: ImportMap;
+  /** @deprecated */
+  imports: Set<string>;
 };
 
 export async function analyzeQuery(
@@ -32,12 +34,14 @@ export async function analyzeQuery(
   });
   const result = generateTSTypeFromCodec(outCodec, cardinality);
 
+  const imports = args.imports.merge(result.imports);
   return {
     result: result.type,
     args: args.type,
     cardinality,
     query,
-    imports: args.imports.merge(result.imports),
+    importMap: imports,
+    imports: imports.get("edgedb") ?? new Set(),
   };
 }
 
--- a/packages/driver/src/reflection/analyzeQuery.ts
+++ b/packages/driver/src/reflection/analyzeQuery.ts
@@ -17,7 +17,7 @@ type QueryType = {
   result: string;
   cardinality: Cardinality;
   query: string;
-  imports: Set<string>;
+  imports: ImportMap;
 };
 
 export async function analyzeQuery(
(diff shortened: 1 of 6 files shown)


packages/driver/src/reflection/analyzeQuery.ts
```diff
@@ -229,3 +229,23 @@ export const defaultApplyCardinalityToTsType =
...
+    }
+    this.get(module)!.add(specifier);
+    return this;
+  }
+
+  merge(map: ImportMap) {
+    const out = new ImportMap();
+    for (const [mod, specifiers] of [...this, ...map]) {
+      for (const specifier of specifiers) {
+        this.add(mod, specifier);
```
scotttrinh, line 246:
```suggestion
        out.add(mod, specifier);
```
—?

packages/generate/src/queries.ts
```diff
@@ -147,9 +144,12 @@ currently supported.`);
...
 }
 
-export function stringifyImports(imports: { [k: string]: boolean }) {
-  if (Object.keys(imports).length === 0) return "";
-  return `import type {${Object.keys(imports).join(", ")}} from "edgedb";`;
+export function stringifyImports(imports: $.ImportMap) {
+  return [...imports]
+    .map(([module, specifiers]) => {
+      return `import type {${[...specifiers].join(", ")}} from "${module}";`;
+    })
```
scotttrinh, line 151: Oh my, I love this. One tiny non-blocking nit:

```suggestion
    .map(([module, specifiers]) =>
      `import type {${[...specifiers].join(", ")}} from "${module}";`
    )
```

This line is kinda noisy already, so this feels a little easier on my eyes/brain now that we don't have any branching logic.

packages/generate/src/queries.ts
```diff
@@ -313,3 +304,11 @@ export function ${functionName}(client: Executor${
...
   }
 }
+
+export const cardinalityToExecutorMethod = {
+  One: "queryRequiredSingle",
+  AtMostOne: "querySingle",
+  Many: "query",
+  AtLeastOne: "query",
+  Empty: "query",
+} satisfies Record<`${$.Cardinality}`, keyof Executor>;
```
scotttrinh, line 314: 👏

packages/driver/src/reflection/analyzeQuery.ts
CarsonF: 🤦 yes

packages/generate/src/queries.ts
CarsonF: This is my own nit as well. I think it was late.

CarsonF:
Kinda a breaking change, which is why tests are failing. generator doesn't know to use updated driver types. Thoughts?

packages/generate/src/queries.ts
```diff
@@ -313,3 +311,31 @@ export function ${functionName}(client: Executor${
...
+
+export class ImportMap extends Map<string, Set<string>> {
```
CarsonF, line 323: I don't love that this is duplicated. It seemed like the easiest way though to handle an old driver version.

scotttrinh (approved):
Thanks for working through the backward compatibility dance here!

PR #967, opened 2024-04-17
Add a streaming RAG method

scotttrinh:
Adds an SSE-style streaming response method along with a lower-level async generator.

---

Example output from httpie:

```
http http://localhost:3004 message=="Tell me something about pluto"
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/event-stream
Date: Wed, 17 Apr 2024 02:14:59 GMT
Keep-Alive: timeout=5
Transfer-Encoding: chunked

event: message_start
data: {
    "message": {
        "id": "chatcmpl-9Ep76xGnUpwaq8Xz8bSq8DAVRbFQ2",
        "model": "gpt-4-0125-preview",
        "role": "assistant"
    },
    "type": "message_start"
}

event: content_block_start
data: {
    "content_block": {
        "text": "",
        "type": "text"
    },
    "index": 0,
    "type": "content_block_start"
}

event: content_block_delta
data: {
    "delta": {
        "text": "Pl",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": "uto",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": "'s",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " surface",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " is",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " primarily",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " made",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " up",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " of",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " nitrogen",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " ice",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": ",",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " methane",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": ",",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " and",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " carbon",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": " mon",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": "oxide",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_delta
data: {
    "delta": {
        "text": ".",
        "type": "text_delta"
    },
    "index": 0,
    "type": "content_block_delta"
}

event: content_block_stop
data: {
    "index": 0,
    "type": "content_block_stop"
}

event: message_delta
data: {
    "delta": {
        "stop_reason": "stop"
    },
    "type": "message_delta"
}

event: message_stop
data: {
    "type": "message_stop"
}


```


packages/ai/src/core.ts
```diff
@@ -111,4 +111,52 @@ export class EdgeDBAI {
 
     return data.response;
   }
+
+  async *getRagAsyncGenerator(
```
jaclarke, line 115: I don't understand what this method is meant to do? Shouldn't it parse the events out of the sse response stream and have the iterator return typed event objects? Currently it seems like it just outputs whatever random chunks of data we get from the server.

packages/ai/src/core.ts
```diff
@@ -111,4 +111,52 @@ export class EdgeDBAI {
...
+          for await (const chunk of generator) {
+            controller.enqueue(chunk);
+          }
+        } catch (error) {
+          controller.error(error);
+        } finally {
+          controller.close();
+        }
+      },
+    });
```
jaclarke, line 156: Why do we take the `ReadableStream` from the fetch response body, convert it into an async iterator, then wrap it into a `ReadableStream` again for the output response body? If the idea is to pass the streaming response on to the client, couldn't we just use the fetch response body in the output response body?

packages/ai/src/core.ts
1st1: > Shouldn't it parse the events out of the sse response stream and have the iterator return typed event objects?

Yeah, that's what we need

CarsonF:
@scotttrinh what do you think about an intermediate object to decouple input/output?
```ts
await queryRag(message, context).text()
await queryRag(message, context).response()
for await (part of queryRag(message, context)) { ... }
```

CarsonF:
> @CarsonF 
> 
> > what do you think about an intermediate object to decouple input/output?
> 
> I'm not sure I understand the suggestion here 🤔 

Instead of 3 top level query functions to vary the output shape, just have 1. Then the output shape is picked independently.
Also the async iterable symbol can be used instead of a 3 dev name.
The snippet is what the user call sites would look like.

scotttrinh:
Ahh, I understand now. Seems reasonable so we'll have to answer the question of which is better on some other axes:

1. What is more intuitive for the end user?
2. What do other SDKs (like OpenAI and Anthropic) do?
3. What has the lowest maintenance overhead?
4. What is easiest to evolve in the future?
5. What is easiest to document?
6. What does the Python version of this same functionality do?

Some of those questions are research for me, others will require a bit of thinking and trying to intuit an answer. Feel free to state the case for a single function that returns an object that exposes the different interfaces vs. separate functions.

CarsonF:
As far as # 1 goes, fetch has the same API
```ts
await fetch(...).json()
await fetch(...).arrayBuffer()
```
I think don't see any problems here.

I can't speak to 2 & 6.

As far as 3 & 4 both the inputs & outputs have to work together in implementation obviously. But this decoupling allows modifying args or output methods without needing to adjust multiple signatures. Giving the `EdgeDBAI` class a single entrypoint for RAG would play nice for other AI strategies in the future.

I did a first pass here https://github.com/edgedb/edgedb-js/compare/stream-ai...CarsonF:edgedb-js:stream-ai

scotttrinh:
> As far as # 1 goes, fetch has the same API

Yeah, and while I agree that works as a primitive since it allows you to have one function with lots of different behavior, I think some people find that API annoying as an end-user API. Great for a primitive, though, and maybe that's really what we're building here.

> As far as 3 & 4 both the inputs & outputs have to work together in implementation obviously. But this decoupling allows modifying args or output methods without needing to adjust multiple signatures. Giving the EdgeDBAI class a single entrypoint for RAG would play nice for other AI strategies in the future.

Yeah, I think that's a compelling argument! Somehow that makes me feel even more like this is a primitive on which higher-level functionality should be built (like our soon-to-come Vercel AI SDK integration) and therefore we should optimize for a flexible low-level primitive.

Lemme shop the idea around a bit more, thanks for taking the time to make an example implementation, that makes it nice and concrete for discussion. 🙏

packages/ai/src/core.ts
```diff
@@ -76,35 +116,78 @@ export class EdgeDBAI {
...
+  }
+
+  async stream(): Promise<any> {
+    const res = await this.fetchRag({
+      model: this.model,
+      prompt: this.prompt,
+      context: this.context,
+      query: this.query,
+      stream: true,
+    });
```
scotttrinh, line 139: Since we always pass `this.model`, etc, I think we can simplify `fetchRag` to just take `stream: boolean`.

scotttrinh (commented):
Yeah! Think this looks pretty good. One small comment.

scotttrinh:
OK, so @diksipav and I have been going back and forth a bunch about the API here, and I think I have some more concrete thoughts on `await queryRag(message, context).text()` vs `await queryRag(message, context)`/`await streamRag(message, context)`:

### `fetch`-like API
One major difference here is that, for consumers of `fetch` you need to inspect the `Response` object to even know which method is the appropriate one based on headers like `content-type` and `transfer-encoding`, etc. We don't have that restriction: the caller gets to _tell_ the producer how it wants the data back: as a string, or as a stream.

### Intermediate object
One possible advantage of having an API with an intermediate object you get back from `queryRag` is that you can refetch the same query without having to redefine it. I haven't yet thought of a good reason or use case for this though. Perhaps if it made sense like it does for the query builder where you have fixed messages and context, but might have different providers or config, but I don't think that's common enough to make this a first-class design choice given that the "query" we're talking about here is a pretty simple data structure: `string`.

### Consistency
This is not a deal breaker, but in my mind it might be a good tie-breaker: consistency with the Python package. [There we have separate methods and no intermediate object.](https://github.com/edgedb/edgedb-python/blob/9471e33a5d0002939241deab8d4ffe02cf3f099c/edgedb/ai/core.py#L112-L139)

CarsonF:
Maybe the python library should change too 😛

packages/ai/src/core.ts
```diff
@@ -76,35 +93,70 @@ export class EdgeDBAI {
...
-    const data = await response.json();
-    if (
-      !data ||
-      typeof data !== "object" ||
-      typeof data.response !== "string"
-    ) {
-      throw new Error(
-        "expected response to be object with response key of type string",
-      );
-    }
```
scotttrinh, line 106: Can we restore this runtime checking of the response?

packages/ai/src/types.ts
```diff
@@ -24,3 +24,54 @@ export interface RAGRequest {
   query: string;
   stream?: boolean;
 }
+
+export type MessageStart = {
```
scotttrinh, line 28: Let's fix these eslint errors by converting these types into interfaces.

packages/ai/README.md
```diff
@@ -37,10 +37,18 @@ Creates an instance of `EdgeDBAI` with the specified client and options.
...
   Returns a new `EdgeDBAI` instance with an updated query context.
 
-- `async queryRag(message: string, context: QueryContext = this.context): Promise<string>`
+- `async queryRag(message: string, context?: QueryContext): Promise<string>`
 
   Sends a query with context to the configured AI model and returns the response as a string.
 
+- `streamRag(message: string, context?: QueryContext): AsyncIterable<StreamingMessage> & PromiseLike<Response>`
+
+  Used both when you want to handle streaming data as it arrives or when you need a `Response` object that streams the results.
```
scotttrinh, line 46: Might be worth documenting these separately, almost like two overloads?

packages/ai/src/core.ts
```diff
@@ -76,35 +93,109 @@ export class EdgeDBAI {
...
-      query: message,
+      query,
       stream: false,
     });
 
-    if (!response.headers.get("content-type")?.includes("application/json")) {
-      throw new Error(
-        "expected response to have content-type: application/json",
-      );
-    }
```
scotttrinh, line 95: Did we take this out because we just want to throw the existing exceptions that the `json` method returns?

scotttrinh (commented):
Great job, and thanks for throwing in the embeddings method as a bonus here.

packages/ai/README.md
diksipav: Hey I updated Readme and merged this, if you have time u can check the updated Readme.

PR #973, opened 2024-04-18
Add webauth & magic links to sveltekit

diksipav:
I tested this with example sveltekit and it works. 

I retested and it works, but not sure is the setCookie function definition the best, especially the params it needs, doesn't look very pretty.

### DIFF
--- a/packages/auth-sveltekit/src/server.ts
+++ b/packages/auth-sveltekit/src/server.ts
@@ -17,6 +17,9 @@ import {
   InvalidDataError,
   OAuthProviderFailureError,
   EdgeDBAuthError,
+  MagicLinkFailureError,
+  type AuthenticationResponseJSON,
+  type RegistrationResponseJSON,
 } from "@edgedb/auth-core";
 import {
   ClientAuth,
@@ -64,6 +67,12 @@ export interface AuthRouteHandlers {
       { verificationToken?: string }
     >
   ) => Promise<never>;
+  onMagicLinkCallback(
+    params: ParamsOrError<{
+      tokenData: TokenData;
+      isSignUp: boolean;
+    }>
+  ): Promise<Response>;
   onSignout?: () => Promise<never>;
 }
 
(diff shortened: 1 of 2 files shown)


packages/auth-sveltekit/src/server.ts
```diff
@@ -290,6 +299,104 @@ export class ServerRequestAuth extends ClientAuth {
...
+
+    this.cookies.set(this.config.pkceVerifierCookieName, verifier, {
+      httpOnly: true,
+      sameSite: "strict",
+      path: "/",
+    });
+  }
+
+  async webAuthnSignIn(data: {
+    email: string;
```
diksipav, line 349: here the emaol can come from FormData, but the assertion is separate

packages/auth-sveltekit/src/server.ts
```diff
@@ -290,6 +299,104 @@ export class ServerRequestAuth extends ClientAuth {
...
+      email,
+      `${this.config.authRoute}/magiclink/callback?isSignUp=true`,
+      new URL(this.config.magicLinkFailurePath, this.config.baseUrl).toString()
+    );
+
+    this.cookies.set(this.config.pkceVerifierCookieName, verifier, {
+      httpOnly: true,
+      sameSite: "strict",
+      path: "/",
+    });
```
scotttrinh, line 322: We recently introduced a method to set the two different cookies, so this will be:

```suggestion
    this.setVerifierCookie(verifier);
```

packages/auth-sveltekit/src/server.ts
scotttrinh: I don't think we should use FormData here at all since the assertion _must_ be JSON: in the form, we should get the email value and pass it as JSON with the rest of the body here.

packages/auth-sveltekit/src/server.ts
```diff
@@ -577,6 +685,124 @@ async function handleAuthRoutes(
...
+      });
+    }
+
+    case "webauthn/signup/options": {
+      const email = searchParams.get("email");
+      if (!email) {
+        throw new InvalidDataError("email missing");
+      }
+
+      return redirect(303, (await core).getWebAuthnSignupOptionsUrl(email));
```
scotttrinh, line 751: I think the other frameworks use 302 Found here?

scotttrinh (changes requested):
Looks great! A couple of small things once you rebase this on the latest `master`.

packages/auth-sveltekit/src/server.ts
diksipav: @scotttrinh  is 302 default if u don't provide anything, because svelte enforces you to provide the status... I was researching this while ago and I used 303 in all svelte examples, https://github.com/sveltejs/kit/issues/5976, https://www.drlinkcheck.com/blog/http-redirects-301-302-303-307-308
Now I found this links and I would still go with 303 hm?

scotttrinh: Hmm, in that case I think `307` is more accurate here. I know these are already GET requests, but I don't think we are intentionally trying to imply that you should always make a GET request here. I think the intention here is "You made some HTTP request to this address, but make the same request again to this new address."

diksipav: hmm makes sense

packages/auth-sveltekit/src/server.ts
```diff
@@ -575,6 +663,124 @@ async function handleAuthRoutes(
...
+      let tokenData: TokenData;
+      try {
+        tokenData = await (await core).getToken(code, verifier);
+      } catch (err) {
+        return onMagicLinkCallback({
+          error: err instanceof Error ? err : new Error(String(err)),
+        });
+      }
+
+      cookies.set(config.authCookieName, tokenData.auth_token, {
```
scotttrinh, line 705: a few more cookies to switch to `this.setAuthCookie`

packages/auth-sveltekit/src/server.ts
```diff
@@ -575,6 +663,124 @@ async function handleAuthRoutes(
...
+          await core
+        ).verifyWebAuthnSignup(verificationToken, verifier);
+      } catch (err) {
+        return onEmailVerify({
+          error: err instanceof Error ? err : new Error(String(err)),
+          verificationToken,
+        });
+      }
+
+      cookies.set(config.authCookieName, tokenData.auth_token, {
```
scotttrinh, line 772: a few more cookies to switch to `this.setAuthCookie`

diksipav: hm, these are inside separate function handleAuthRoutes, its not part of the class... cookies are passed as a param

scotttrinh: Ahh. Maybe we need a function then that does the same thing as the class method does just so we don't have to work so hard to have consistent definitions for the cookie parameters themselves.

Even just a constant for the options would be fine if a function feels too heavy.

packages/auth-sveltekit/src/server.ts
```diff
@@ -106,6 +106,34 @@ export default function serverAuth(client: Client, options: AuthOptions) {
   };
 }
 
+function setCookie(
+  cookies: Cookies,
+  value: string,
+  config: AuthConfig,
+  isAuth = false
```
scotttrinh, line 113: I think I prefer two separate functions here than a boolean with a default value. If you want to share the base configuration object between both functions, that's fine! Something like:

```ts
const BASE_COOKIE_CONFIG = {
  httpOnly: true,
  sameSite: "lax",
  path: "/",
};
```

then in the functions:

```ts
function setVerifierCookie(cookies: Cookies, config: AuthConfig, value: string): void {
  const expires = new Date(Date.now() + 1000 * 60 * 24 * 7); // in 7 days
  cookies.set(config.pkceVerifierCookieName, value, {
    ...BASE_COOKIE_CONFIG,
    expires,
    secure: config.baseUrl.startsWith("https"),
  });
}
```

Then in the class you can:

```ts
private setVerifierCookie(verifier: string) {
  setVerifierCookie(this.cookies, this.config, verifier);
}
```

---

The advantage here is that you can look at the call site and know which cookie it's setting without knowing if the optional `true` means verifier or auth token, etc.

scotttrinh (changes requested):

packages/auth-sveltekit/src/server.ts
diksipav: yes, I agree with all of these, will update : )

packages/auth-sveltekit/src/server.ts
```diff
@@ -132,28 +182,6 @@ export class ServerRequestAuth extends ClientAuth {
...
-  private setAuthCookie(authToken: string) {
-    const expires = Auth.getTokenExpiration(authToken);
-    this.cookies.set(this.config.authCookieName, authToken, {
-      httpOnly: true,
-      sameSite: "lax",
-      path: "/",
-      expires: expires ?? undefined,
-      secure: this.isSecure,
-    });
-  }
```
scotttrinh, line 155: Not a blocker, but the change here would be smaller and easier to follow for future endpoints if we _kept_ these class methods and just had them call the new function, so that other methods could continue to just call `this.setAuthCookie(tokenData.auth_token)`, etc.

```ts
  private setVerifierCookie(verifier: string) {
    setVerifierCookie(this.cookies, this.config, verifier);
  }

  private setAuthCookie(authToken: string) {
    setAuthCookie(this.cookies, this.config, authToken);
  }
```

And for bonus points doing a similar thing for the delete cookie functionality:

```ts
  private deleteVerifierCookie() {
    deleteCookie(this.cookies, this.config.pkceVerifierCookieName);
  }

  private deleteAuthCookie() {
    deleteCookie(this.cookies, this.config.authCookieName);
  }
```

scotttrinh (approved):
🏆

PR #982, opened 2024-04-27
Fix Docs Interface Generator runtime command name

ggallon:
Hi,

Corrected the command name `... @edgedb/generate queries` to  `.. @edgedb/generate interfaces`


scotttrinh (approved):
Ahh, nice catch thanks!

PR #983, opened 2024-04-27
Fix Bun tab in Docs Generators

ggallon:
Hi,

Corrects tab display, Bun tab is wrong, see ([link](https://docs.edgedb.com/libraries/js/generation#generators)) :

Capture d’écran 2024-04-27 à 16 28 35


scotttrinh (approved):
🏆

PR #986, opened 2024-04-29
Extract driver fetch configuration for reuse

diksipav:
AI ext and all other exts except the 'auth' one go through authentication.
We extract driver's authenticated fetch into a reusable util.     
The util creates db path that it uses for the url, and adds headers to the fetch req.

### DIFF
--- a/packages/driver/src/utils.ts
+++ b/packages/driver/src/utils.ts
@@ -88,8 +88,8 @@ export async function makeFetch(
   const baseUrl = `${protocol}://${address[0]}:${address[1]}`;
   const databaseUrl = `${baseUrl}/db/${database}/${basePath ?? ""}`;
 
-  if (!token) {
-    token = await httpSCRAMAuth(baseUrl, config.user, config.password ?? "");
+  if (!token && config.password != null) {
+    token = await httpSCRAMAuth(baseUrl, config.user, config.password);
     _tokens.set(config, token);
   }
 
--- a/packages/driver/src/utils.ts
+++ b/packages/driver/src/utils.ts
@@ -16,11 +16,13 @@
  * limitations under the License.
  */
 
-import { ProtocolVersion } from "./ifaces";
+import type { ResolvedConnectConfig } from "./conUtils";
+import type { HttpSCRAMAuth } from "./httpScram";
+import type { ProtocolVersion } from "./ifaces";
 
 const idCounter: { [key: string]: number } = {};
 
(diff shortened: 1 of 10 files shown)


packages/driver/src/utils.ts
```diff
@@ -65,3 +67,48 @@ export interface CryptoUtils {
...
+  let token = config.secretKey ?? _tokens.get(config);
+
+  const { address, tlsSecurity, database } = config;
+
+  const protocol = tlsSecurity === "insecure" ? "http" : "https";
+  const baseUrl = `${protocol}://${address[0]}:${address[1]}`;
+  const databaseUrl = `${baseUrl}/db/${database}/${basePath ?? ""}`;
+
+  if (!token) {
+    token = await httpSCRAMAuth(baseUrl, config.user, config.password ?? "");
```
jaclarke, line 92:
```suggestion
  if (!token && config.password != null) {
    token = await httpSCRAMAuth(baseUrl, config.user, config.password);
```

packages/driver/src/utils.ts
```diff
@@ -65,3 +67,48 @@ export interface CryptoUtils {
...
 }
+
+const _tokens = new WeakMap<ResolvedConnectConfig, string>();
+
+export type FetchWrapper = (
+  path: string,
+  init: RequestInit
+) => Promise<Response>;
+
+export async function makeFetch(
```
jaclarke, line 78: Nitpick: Should we name this something like `AuthenticatedFetch`/`AuthedFetch` and `makeAuthFetch`, so it's more clear what it does?

scotttrinh: Yeah, this is a good question. Furthermore, I wonder if `fetch` is even really the best name for this since it ends up encouraging you to shadow the `fetch` global which can be confusing.

Also, it really does slightly more than authentication, it also sets up the base URL, and maybe more things in the future. The thing it does is roughly: "Make an authenticated HTTP request a specific EdgeDB branch", so maybe there is something there? I'll think of a few names as well.

diksipav: Hm, I am not completely sure, I renamed the util func to getAuthenticatedFetch and it returns AuthenticatedFetch... it can maybe also be AuthenticatedDBFetch... but yes, not 100% what is the most future proof

scotttrinh: I was thinking something like `ServerRequest` or something like that? Still not sure 😅

PR #988, opened 2024-05-01
Use browser crypto for SCRAM

scotttrinh:
Since all supported platforms support browser-compatible crypto, remove
the extraneous adapters and standardize on the browser API. This
requires a flag for Node v18 to enable the browser global
`--experimental-global-webcrypto`, but this is already required by a few
other libraries in our ecosystem.


README.md
```diff
@@ -39,6 +39,9 @@ writing some simple queries.
 ### Requirements
 
 - Node.js 18+
+  - The driver requires the Browser API `crypto` to be available as a global.
+    Therefore, for Node 18, you _must_ run `node` with the special command line
```
elprans, line 43: Since you mention 18+ above, I would rephrase this to clarify that _only_ 18 needs the flag, as the feature stabilized in 20+

scotttrinh: Instead of "Therefore, for Node 18," maybe "Therefore, for Node 18 only,"?

scotttrinh: Or maybe "Therefore, for Node versions before 19,"?

scotttrinh:
I'm going to sit on this for a while. Feels like something that we'd want to require a major version bump 😰

PR #1007, opened 2024-05-06
Replace `Error.source` with standardized `cause`

scotttrinh:
`Error.cause` is now supported across all of our supported platforms. Use that instead of the custom `source` property.

See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)

### DIFF
--- a/packages/driver/src/rawConn.ts
+++ b/packages/driver/src/rawConn.ts
@@ -152,7 +152,7 @@ export class RawConnection extends BaseRawConnection {
       this.buffer.takeMessage() &&
       this.buffer.getMessageType() === chars.$E
     ) {
-      newErr.source = this._parseErrorMessage();
+      newErr.cause = this._parseErrorMessage();
     }
 
     this._abortWithError(newErr);
@@ -162,7 +162,7 @@ export class RawConnection extends BaseRawConnection {
     const newErr = new errors.ClientConnectionClosedError(
       `network error: ${err}`
     );
-    newErr.source = err;
+    newErr.cause = err;
 
     try {
       this._abortWaiters(newErr);
@@ -335,7 +335,7 @@ export class RawConnection extends BaseRawConnection {
             );
             break;
         }
-        err.source = e;
+        err.cause = e;
         throw err;
       }
     } finally {
--- a/packages/driver/src/rawConn.ts
+++ b/packages/driver/src/rawConn.ts
@@ -152,7 +152,10 @@ export class RawConnection extends BaseRawConnection {
       this.buffer.takeMessage() &&
       this.buffer.getMessageType() === chars.$E
     ) {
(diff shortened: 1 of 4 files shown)


packages/driver/src/rawConn.ts
CarsonF, line 1: These assignments cause the `cause` to become enumerable.

Could be changed to
```ts
Object.defineProperty()
```
Or maybe a helper function?

scotttrinh: Hmm, that's the default behavior of `Error.cause`, even when you add it in the error constructor. Would be weird if it was different on direct assignment here. Is the enumerable thing a big deal breaker here?

scotttrinh: oh, ha, when I was doing the testing just in the browser console, I used the wrong variable in my `for..in` check. I agree that if direct assignment makes it non-enumerable, that our usage should keep that. Will fix!

packages/driver/src/rawConn.ts
```diff
@@ -162,7 +162,7 @@ export class RawConnection extends BaseRawConnection {
     const newErr = new errors.ClientConnectionClosedError(
       `network error: ${err}`
     );
-    newErr.source = err;
+    newErr.cause = err;
```
CarsonF: This could be moved into constructor option

packages/driver/src/rawConn.ts
```diff
@@ -335,7 +335,7 @@ export class RawConnection extends BaseRawConnection {
             );
             break;
         }
-        err.source = e;
+        err.cause = e;
```
CarsonF: I think this duplicates the logic above?

packages/driver/src/rawConn.ts
```diff
@@ -152,7 +152,7 @@ export class RawConnection extends BaseRawConnection {
       this.buffer.takeMessage() &&
       this.buffer.getMessageType() === chars.$E
     ) {
-      newErr.source = this._parseErrorMessage();
+      newErr.cause = this._parseErrorMessage();
```
CarsonF: Does this logic truly has to be done after the waiter check above? If not it could be merged into error constructor, otherwise I think it's the only remaining assignment.

packages/driver/src/rawConn.ts
scotttrinh: lol, I added that for debugging and it made it into `master` and I totally forgot about it 🤦 will fix.

packages/driver/src/rawConn.ts
scotttrinh: I don't think it makes sense to refactor this logic in the context of this PR, so I think it's safest to leave this one as a conditional assignment.

CarsonF (approved):
Nice LGTM!

PR #1008, opened 2024-05-06
Mark generateFiles imports as optional

CarsonF:
A small tweak easing migration to `importMap`

Mark both as optional so both old & new `analyzeQuery` can be used as input.


scotttrinh (approved):
Make sense!

PR #1014, opened 2024-05-09
Check globalThis.Buffer to avoid runtime error

### LINKED ISSUE #1003: Uncaught ReferenceError: Buffer is not defined in Partykit (Cloudflare Worker)
**Describe the bug**

I am trying to use EdgeDB in Partykit, which runs on Cloudflare Workers. When starting the Partykit dev server locally, it throws the error: `Uncaught ReferenceError: Buffer is not defined`

I am not sure how far the local dev server goes to simulate a Cloudflare Worker environment, but CF Workers do support Buffer: https://developers.cloudflare.com/workers/runtime-apis/nodejs

It has to be imported, it can't be used as a global.

The error happens immediately when the dev server is started.

**Reproduction**

Any use of the edgedb-js client.

**Expected behavior**

I do not expect an error.

- OS:
- EdgeDB version (e.g. `2.0`): 5.2+1e07d04
- EdgeDB CLI version (e.g. `2.0`): EdgeDB CLI 5.1.0+8ff9f85
- `edgedb-js` version (e.g. `0.20.10`;): 1.5.5
- Node/Deno version: 20.12.2

**Additional context**

I am running the Partykit dev server with Bun, but I also tried with Node and it had the same result.

scotttrinh:
Trying to access `Buffer` directly from global rather than from the `globalThis` object results in a runtime error on Cloudflare (and likely other runtimes as well).

Also, it turns out we were never hitting the `Buffer` case because we were comparing the function to the string `"function"` rather than comparing the type! And because we were using `tsignore` we were not getting a helpful type error. Eek.

Closes #1003

### DIFF
--- a/packages/driver/src/primitives/buffer.ts
+++ b/packages/driver/src/primitives/buffer.ts
@@ -21,7 +21,7 @@ import * as chars from "./chars";
 import { LegacyHeaderCodes } from "../ifaces";
 
 /* WriteBuffer over-allocation */
-const BUFFER_INC_SIZE: number = 4096;
+const BUFFER_INC_SIZE = 4096;
 
 const EMPTY_BUFFER = new Uint8Array(0);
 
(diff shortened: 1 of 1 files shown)


scotttrinh:
Actually, instead of doing this and needing a bunch of type manipulation for the Deno case, we're going to just lean on node-compatibility for environments for now and use Deno's `node:buffer` implementation (which is great).

PR #1019, opened 2024-05-13
Escape double-backslashes in EdgeQL

### LINKED ISSUE #878: Bug in backslash escaping of TS query generation
**Describe the bug**
I hit this bug when using regexes in my edgeql query, which require backslash escapes "\\". It looks like the query generator is not adding another pair of escapes "\\" for the TS code. When I add them, everything works as expected.

Example
```
debtor_name := re_match('Debtor [Nn]ame:?\\s*(.*)?\\s+United States', .extracted_text)[0],
```
vs
```
debtor_name := re_match('Debtor [Nn]ame:?\\\\s*(.*)?\\\\s+United States', .extracted_text)[0],
```

**Reproduction**
Run `npx @edgedb/generate queries --target ts` . Call the generated query from a TS app. See error.

**Expected behavior**
Expect query to execute. 

4.2+8778fc1

scotttrinh:
These show up in regular expressions, so do a blanket find-and-replace to a quadruple backslash.

Closes #878

### DIFF
--- a/integration-tests/lts/dbschema/queries/freeShape.edgeql
+++ b/integration-tests/lts/dbschema/queries/freeShape.edgeql
@@ -4,4 +4,5 @@ select {
   data := <str>$data,
   required multi arg := {'asdf'},
   enums := [Genre.Horror, Genre.Action],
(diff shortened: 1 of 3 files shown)


gregdingle:
Nice work!

PR #1020, opened 2024-05-13
Transition to eslint

diksipav:
**NOTE:** I suggest reviewing commit by commit.

I added the config (eslint.config) for eslint v9 with flat config.

I fixed majority of errors throughout the codebase. And hopefully `any` types in Remix auth lib.

But definitely a good review is needed of the changes I made.

And there are some errors I wasn't sure about how to fix and also many other any type warnings.

I left todo comments somewhere where I wasn't sure if my "fix" is good enough.

More info: here are `ts-eslint` 
[recommended](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts) rules,
[recommended-type-checked](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended-type-checked.ts) rules,
[strict](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/strict.ts),
[strict-type-checked](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/strict-type-checked.ts),
[stylistic](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/stylistic.ts),
[stylistic-type-checked](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/stylistic-type-checked.ts).

EDIT:
I removed stylistic "@typescript-eslint/class-literal-property-style" rule because it is enforcing `readonly field` class literal values, and we use getters in many places (which is actually safer).

TODO:
Check `packages/generate/src/syntax/castMaps.ts`. Not sure what's this file and how to fix it.


package.json
```diff
@@ -7,19 +7,18 @@
   ],
   "packageManager": "yarn@1.22.19",
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^5.60.1",
-    "@typescript-eslint/parser": "^5.60.1",
-    "eslint": "^8.43.0",
+    "@eslint/js": "^9.2.0",
+    "@typescript-eslint/eslint-plugin": "^7.8.0",
+    "@typescript-eslint/parser": "^7.8.0",
```
scotttrinh, line 12: I don't think we need these anymore, right? Just `typescript-eslint`?

package.json
```diff
@@ -7,19 +7,18 @@
...
-    "tslint-config-prettier": "^1.18.0",
-    "tslint-plugin-prettier": "^2.3.0",
-    "typescript": "^5.4.5"
+    "typescript": "^5.4.5",
+    "typescript-eslint": "^7.8.0"
   },
   "scripts": {
-    "lint": "tslint 'packages/*/src/**/*.ts'",
-    "eslint": "eslint 'packages/*/src/**/*.ts'",
+    "lint": "eslint 'packages/*/src/**/*.ts'",
```
scotttrinh, line 20: What do you think of maybe passing `--quiet` here? That will hide `warn` level rule violations. That means we'll still see warnings in our editors (like explicit `any`), but when we get a lint failure, we'll just see the errors?

```suggestion
    "lint": "eslint 'packages/*/src/**/*.ts' --quiet",
```

Docs: https://eslint.org/docs/latest/use/command-line-interface#--quiet

packages/driver/src/errors/base.ts
```diff
@@ -65,7 +64,9 @@ function tryParseInt(val: any) {
   if (val instanceof Uint8Array) {
     try {
       return parseInt(utf8Decoder.decode(val), 10);
-    } catch {}
+    } catch (err) {
+      console.error("Failed to parse integer value: ", val, err);
```
scotttrinh, line 68: If eslint is complaining about an empty block here, I'd rather not log an error here just to silence the linter.

Oh, actually, I think the fix here is to actually return `null` inside of this catch and remove it from the outer scope instead of silently swallowing the error, or loudly logging the error.

```suggestion
      return null;
```

packages/generate/src/builders.ts
```diff
@@ -849,7 +851,9 @@ export class DirBuilder {
       let oldContents = "";
       try {
         oldContents = await readFileUtf8(filePath);
-      } catch {}
+      } catch (err) {
+        console.error("Error occurred while reading file", filePath, err);
```
scotttrinh, line 855: Similar to the last comment, let's not add side effects in this linting PR. I think we can maintain the same behavior here by removing the assignment of `oldContents` from outside of the try catch and assign the empty string in the catch block.

```ts
let oldContents;
try {
  oldContents = await readFileUtf8(filepath);
} catch {
  oldContents = "";
}
```

Or even using `then` to avoid the awkward try catch here:

```ts
const oldContents = await readFileUtf8(filepath)
  .then((content) => content)
  .catch(() => "");
```

packages/generate/src/edgeql-js.ts
```diff
@@ -185,7 +185,9 @@ export async function generateQueryBuilder(params: {
     let oldContents = "";
     try {
       oldContents = await readFileUtf8(outputPath);
-    } catch {}
+    } catch (err) {
+      console.error("Error occurred while reading file", outputPath, err);
```
scotttrinh, line 189: Same thing in this file about avoiding side effects in these catch blocks.

packages/generate/src/syntax/insert.ts
```diff
@@ -252,10 +252,12 @@ export function $normaliseInsertShape(
...
-    if (pointer.__kind__ === "property") {
-      if (pointer.target.__name__ === "std::json") {
-      }
-    }
+
+    // Todo delete or fix this
+    // if (pointer.__kind__ === "property") {
+    //   if (pointer.target.__name__ === "std::json") {
+    //   }
+    // }
```
scotttrinh, line 260: ha! Yeah, just delete it, it obviously isn't needed and if it's a bug, having this no-op block of code won't really help us fix the bug.

scotttrinh (changes requested):

scotttrinh:
We also want to:

- [ ] Remove the [extra call to `yarn eslint` from the test workflow](https://github.com/edgedb/edgedb-js/blob/2e247adc7b04197f6c5eebba7767a0cbafe86248/.github/workflows/tests.yml#L76-L78)
- [ ] Fix the failing errors (I think we can just add an ignore rule for now)
- [ ] Remove all of the tslint disable comments throughout

package.json
diksipav: Oh yes, I forgot to delete them, I also removed `eslint-config-prettier` . 
I'll check the default rules that are used but I believe that stylistic rules that are handled by prettier are removed from recommended `eslint` and `ts-eslint` rules.

packages/driver/src/errors/base.ts
diksipav: hmm well in this case null is returned if `if block` is not satisfied, so if we remove outer null then undefined will be returned in this case... I dont know why we explicitly return null but maybe there's a reason

scotttrinh: Well, when the `if` exits without having returned the `parseInt`, it will still return `null` here. So what we have here, logically is:

if "val" is an instance of "Uint8Array" and it we successfully parse an int from the UTF-8 buffer decoder with radix of 10, we return that parsed int.  otherwise, return `null`.

There isn't (currently) a flow where we return `undefined` here, and I'm not suggesting that either. Just:

```ts
function tryParseInt(val: any) {
  if (val instanceof Uint8Array) {
    try {
      return parseInt(utf8Decoder.decode(val), 10);
    } catch {
      return null;
    }
  }
  return null;
}
```

or if we want to do an early return:

```ts
function tryParseInt(val: any) {
  if (!(val instanceof Uint8Array) return null;
  try {
    return parseInt(utf8Decoder.decode(val), 10);
  } catch {
    return null;
  }
}
```

diksipav: ah sorry, I thought u said to delete "return null" from the outer scope (outside try/catch block) and return null only inside catch block...

scotttrinh: oh, I did say that originally! 🤦 glad you caught that, sorry for the confusion.

package.json
```diff
@@ -8,10 +8,7 @@
   "packageManager": "yarn@1.22.19",
   "devDependencies": {
     "@eslint/js": "^9.2.0",
-    "@typescript-eslint/eslint-plugin": "^7.8.0",
-    "@typescript-eslint/parser": "^7.8.0",
     "eslint": "^9.2.0",
-    "eslint-config-prettier": "^8.8.0",
```
scotttrinh, line 13: Actually, I think we should leave this and update it to the version that works with flat config so it turns off all of the styling rules that are covered by Prettier.

packages/generate/src/edgeql-js.ts
```diff
@@ -306,12 +299,9 @@ project to exclude these files.`
...
 
-    let gitIgnoreFile: string | null = null;
-    try {
-      gitIgnoreFile = await readFileUtf8(gitIgnorePath);
-    } catch (err) {
-      console.error("Error occurred while reading file", gitIgnorePath, err);
-    }
+    const gitIgnoreFile = await readFileUtf8(gitIgnorePath)
+      .then((content) => content)
+      .catch(() => "");
```
scotttrinh, line 304: I think this should `catch` to `null` instead of the empty string?

```suggestion
    const gitIgnoreFile = await readFileUtf8(gitIgnorePath)
      .then((content) => content)
      .catch(() => null);
```

package.json
diksipav: Do you think some stylistic rule conflicts with prettier, because `ts-eslint` state on their website: "Neither typescript-eslint nor ESLint core enable any formatting-related rules in any recommended presets."

So that's why I thought its safe to just omit `eslint-config-prettier` completely.

packages/generate/src/edgeql-js.ts
diksipav: oh yes, good catch!

PR #1043, opened 2024-06-17
Use given cardinality in select expression

scotttrinh:
Previously, if you made a select expression with a constant or coalescing expression, we _always_ calculated the cardinality of a object key based on the pointer's cardinality, rather than allowing you to return something _assignable_ to that cardinality. This change will check if the assignment is possible, and override the cardinality when converting the query builder expression to a TS type.


CarsonF (approved):
Sweet! Thanks @scotttrinh!

PR #1051, opened 2024-07-05
Remove from object shape keys with value never

### LINKED ISSUE #1037: Shape with pointer of `false` produces type suggesting the pointer is returned
```typescript
e.select(e.X, {
  foo: false
})
```
Runtime omits `foo` from EdgeQL, but TS types it as `never`.
This allows it to be compatible, somehow, with a type expecting that property.

```typescript
class X {
  foo: string;
}
const withX = (x: X) => {...};

const result = await e.select(e.X, {
  foo: false
}).run();

// This is fine even though `foo` is required, and not in the DB result.
withX(result[0]);
```

I think the type should be changed to `undefined`, or maybe `unknown` instead?

diksipav:
Hey, I am not sure if this is 100% what we need in order to fix #1037 .

It works okay with the test mentioned in the issue.

Screenshot 2024-07-05 at 18 39 56
Screenshot 2024-07-05 at 18 39 33

### DIFF
--- a/integration-tests/lts/dbschema/default.esdl
+++ b/integration-tests/lts/dbschema/default.esdl
@@ -31,6 +31,7 @@ module default {
       constraint exclusive;
     };
     property height -> decimal;
+    property isAdult -> bool;
   }
 
   type Villain extending Person {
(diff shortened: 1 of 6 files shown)


scotttrinh (changes requested):
Let's also add some tests to https://github.com/edgedb/edgedb-js/blob/master/integration-tests/lts/select.test.ts to prevent future regressions and to ensure this change has the desired effect like your screenshots.

diksipav:
I saw the issu

> Let's also add some tests to https://github.com/edgedb/edgedb-js/blob/master/integration-tests/lts/select.test.ts to prevent future regressions and to ensure this change has the desired effect like your screenshots.

Yes, will do. And actually some tests are failing. Have to see what's going on there.

scotttrinh (approved):
Awesome job 🏆 . I'm a little 🤔 about why we were asserting `never` on that computed before, but I can't think of _wny_ we'd want it to work like it was before.

PR #1053, opened 2024-07-16
Add __EDGEDB_DEVMODE=1 when running tests locally

diksipav:
We need this in order to successfully run tests locally. 

Thank you @jaclarke  for helping me solving this problem.

### DIFF
--- a/packages/driver/test/testUtil.ts
+++ b/packages/driver/test/testUtil.ts
@@ -90,6 +90,10 @@ export const getServerCommand = (
     srvcmd = process.env.EDGEDB_SERVER_BIN;
   }
 
+  if (!(process.env.EDGEDB_SERVER_BIN || process.env.GITHUB_ACTIONS)) {
+    process.env.__EDGEDB_DEVMODE = "1";
+  }
+
   let args = [srvcmd];
   if (process.platform === "win32") {
     args = ["wsl", "-u", "edgedb", ...args];
--- a/packages/driver/test/testUtil.ts
+++ b/packages/driver/test/testUtil.ts
@@ -90,7 +90,7 @@ export const getServerCommand = (
     srvcmd = process.env.EDGEDB_SERVER_BIN;
   }
 
-  if (!(process.env.EDGEDB_SERVER_BIN || process.env.GITHUB_ACTIONS)) {
+  if (!(process.env.EDGEDB_SERVER_BIN || process.env.CI)) {
     process.env.__EDGEDB_DEVMODE = "1";
   }
 
--- a/integration-tests/legacy/package.json
+++ b/integration-tests/legacy/package.json
@@ -15,7 +15,7 @@
     "conditional-type-checks": "^1.0.6",
     "jest": "^29.7.0",
     "ts-jest": "^29.1.4",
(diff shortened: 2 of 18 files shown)


packages/driver/test/testUtil.ts
```diff
@@ -90,6 +90,10 @@ export const getServerCommand = (
     srvcmd = process.env.EDGEDB_SERVER_BIN;
   }
 
+  if (!(process.env.EDGEDB_SERVER_BIN || process.env.GITHUB_ACTIONS)) {
```
jaclarke, line 93:
```suggestion
  if (!(process.env.EDGEDB_SERVER_BIN || process.env.CI)) {
```
I think `CI` is the more commonly used env var?

scotttrinh, line 93: How do you end up having no `EDGEDB_SERVER_BIN` when running the driver test? Do you have `edgedb-server` in your `PATH` somehow?

diksipav: We didn't have EDGEDB_SERVER_BIN, but we firstly come up with that solution using edgedb-server portable install and then pointing that install with EDGEDB_SERVER_BIN, and then after that we figured out __EDGEDB_DEVMODE.

scotttrinh: > then pointing that install with EDGEDB_SERVER_BIN, and then after that we figured out __EDGEDB_DEVMODE.

But doesn't this only set `__EDGEDB_DEVMODE` if there is no `EDGEDB_SERVER_BIN` set? When is that ever true?

diksipav: I think that usually locally `EDGEDB_SERVER_BIN` is not set? but if you set it then `__EDGEDB_DEVMODE` is not required (not sure if is an error to set it if `EDGEDB_SERVER_BIN` is also set )...

or you have different idea how to solve this? maybe I can just completely remove `EDGEDB_SERVER_BIN` from here

scotttrinh: > I think that usually locally EDGEDB_SERVER_BIN is not set?

Well, sure, but in order to run the driver tests, you _have_ to specify a path to the server binary or have a server binary in your PATH already, otherwise you cannot start the server for the tests (since it doesn't use an EdgeDB project for testing).

Or maybe I'm misunderstanding and there is a way to run the tests without a EDGEDB_SERVER_BIN?

diksipav: oh but I was running tests inside venv

scotttrinh: Ahh, so in that case `edgedb-server` _is_ in the PATH, right, but then we need to explicitly set devmode, got it.

Hmmm... I'd rather document using `EDGEDB_SERVER_BIN` better rather than requiring that someone also has the edgedb repo locally and knows how to set up a venv to put `edgedb-server` in their PATH.

diksipav: yes, I completely agree we should document it, probably both options... it took me hours to debug this :D 
I'll add some guidelines in the morning.

diksipav:
I moved the PR [here](https://github.com/edgedb/edgedb-js/pull/1056/files)

PR #1056, opened 2024-07-17
Update readme with config needed to run tests locally

diksipav:


packages/driver/test/testUtil.ts
```diff
@@ -90,6 +90,10 @@ export const getServerCommand = (
     srvcmd = process.env.EDGEDB_SERVER_BIN;
   }
 
+  if (!(process.env.EDGEDB_SERVER_BIN || process.env.CI)) {
+    process.env.__EDGEDB_DEVMODE = "1";
+  }
```
scotttrinh, line 95: Alright, one last objection to this very simple change 😅 

It seems a little bit odd to make setting an environment variable on the current process a side-effect of calling the `getServerCommand` function. I think there _might_ be a better place to do this:

https://github.com/edgedb/edgedb-js/blob/455e3480f1c12f8fa5dfc732b7240521cf8689b6/packages/driver/test/testUtil.ts#L150-L152

Basically provide it to the process that gets spawned to run the server directly. Something like:

```ts
  const maybeEnvWithDevMode =
    process.env.EDGEDB_SERVER_BIN || process.env.CI
      ? {}
      : {
          __EDGEDB_DEVMODE: "1",
        };
  const proc = child_process.spawn(cmd[0], cmd.slice(1), {
    env: { ...process.env, ...env, ...maybeEnvWithDevMode },
  });
```

scotttrinh (approved):

PR #1058, opened 2024-07-18
add auth support for solid start

Huliiiiii:
[template](https://github.com/Huliiiiii/solid-start-edgedb-auth-template)
Based on auth-nextjs and auth-remix, replace functions with corresponding alternatives. Tested on the template, capable of normal sign up, sign in, add and delete items, but no further extensive testing has been conducted.


Huliiiiii:
Their are some type issues in CreateAuthRouteHandlers, is replacing return types with unknown a good solution?

scotttrinh:
Amazing, thanks so much for the contribution! I'll take a closer look at it later today, but based on my initial light pass, it seems solid 😉 

> There are some type issues in CreateAuthRouteHandlers (redirect function in solid router return a response, but other router maynot do the same thing), is replacing return types with unknown (or Response | never) a good solution?

It seems idiomatic to `throw` using `redirect` (see https://docs.solidjs.com/solid-router/reference/data-apis/response-helpers#redirect) so I _think_ it's fine to keep `never` here? Can you share some code that you have that shows the type error you're seeing?

Huliiiiii:
https://github.com/solidjs/solid-start/issues/1512#issuecomment-2161045705
Actually, throw redirect is handle by wrappers like:
```typescript
try {
  // do something
} catch (err) {
  if (err instanceof Response) {
    return err
  } else {
    throw err
  }
}
```
If the code is not wrapped by a wrapper, it will cause an error. This may be a solid start issue. Therefore, it is safer to return a redirect if needed.
```typescript
export const { GET, POST } = serverAuthHelper.createAuthRouteHandlers({
  async onBuiltinUICallback({ error, tokenData, isSignUp }) {
    "use server"
    if (error) {
      console.error("Authentication failed: ", error)
      return redirect("/error?error=auth-failed")
    }

    if (!tokenData) {
      console.error("Email verification required.")
      return redirect("/error?error=email-verification-required")
    }

    if (isSignUp) {
      const client = serverAuthHelper.getSession({
        tokenData,
      }).client

      const emailData = await client.querySingle<{ email: string }>(`
        SELECT ext::auth::EmailFactor {
          email
        } FILTER .identity = (global ext::auth::ClientTokenIdentity)
      `)

      console.log("emailData: ", emailData)

      await client.query(`
        INSERT User {
          name := '',
          email := '${emailData?.email}',
          userRole := 'user',
          identity := (global ext::auth::ClientTokenIdentity)
        }
      `)
    }

    return redirect("/")
  },
  async onSignout() {
    "use server"
    throw redirect("/")
  },
})
```
By the way, here's some code from react that needs to be cleaned up (e.g. the handle actions, I didn't figure out how it works).

scotttrinh:
> https://github.com/solidjs/solid-start/issues/1512#issuecomment-2161045705

Ahh, I see. Well, it's fine to return `Response | never` here! I never really liked the "throw redirects" style that other frameworks have, in my opinion it's more clear if functions actually return values rather than opt out of the type system and normal control flow via `throw`.

> By the way, here's some code from react that needs to be cleaned up (e.g. handle actions, I didn't figure out how it works).

Can you point out what needs to be cleaned up? That looks like example code, not library code.

Huliiiiii:
> > [solidjs/solid-start#1512 (comment)](https://github.com/solidjs/solid-start/issues/1512#issuecomment-2161045705)
> 
> Ahh, I see. Well, it's fine to return `Response | never` here! I never really liked the "throw redirects" style that other frameworks have, in my opinion it's more clear if functions actually return values rather than opt out of the type system and normal control flow via `throw`.
> 
> > By the way, here's some code from react that needs to be cleaned up (e.g. handle actions, I didn't figure out how it works).
> 
> Can you point out what needs to be cleaned up? That looks like example code, not library code.

I mean the library (sorry for my English).

The part need to clean is:
https://github.com/edgedb/edgedb-js/blob/f8e5af2adf01ee152e4b8a88c4e8f1eabc2f6d46/packages/auth-solid-start/src/server/index.ts#L957-L959
This may always be false. I think just need to remove them.

packages/auth-solid-start/src/server/index.ts
```diff
@@ -0,0 +1,987 @@
...
+}
+
+export interface CreateAuthRouteHandlers {
+  onOAuthCallback(
+    params: ParamsOrError<{
+      tokenData: TokenData;
+      provider: BuiltinOAuthProviderNames;
+      isSignUp: boolean;
+    }>,
+  ): Promise<never | Response>;
```
scotttrinh, line 67: Oh actually, you can type these return values as `Promise<Response>` since returning `Promise<never>` will always match `Promise<Response>` due to ["`never` is assignable to everything."](https://www.typescriptlang.org/docs/handbook/type-compatibility.html#any-unknown-object-void-undefined-null-and-never-assignability) That means that if your function _does_ return `never`, it'll still type check here.

Huliiiiii:
Trying to use it in my own project with a remote database, but received: 
```
Authentication failed: InvalidDataError: Missing query parameter: code
    at decodeError (D:\...\node_modules\@edgedb\auth-core\dist\utils.js:117:16)
    at requestGET (D:\...\node_modules\@edgedb\auth-core\dist\utils.js:47:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async GET (D:/.../node_modules/edge-auth-solid-start/src/server/index.ts:293:27)    
    at async eval (D:/.../node_modules/@solidjs/start/dist/server/handler.js:38:23)        
    at async _callHandler (file:///D:/.../node_modules/h3/dist/index.mjs:1821:16)
    at async _callHandler (file:///D:/.../node_modules/h3/dist/index.mjs:1821:16)
    at async file:///D:/.../node_modules/h3/dist/index.mjs:1962:19
    at async Object.callAsync (file:///D:/.../node_modules/unctx/dist/index.mjs:72:16)     
    at async Server.toNodeHandle (file:///D:/.../node_modules/h3/dist/index.mjs:2249:7) 
```
Is the problem in auth/core or this package?

Huliiiiii:
QQ_1721620240738


QQ_1721620213733

scotttrinh:
Sorry we've been sitting on this so long! I'll get back to this in a bit, sorry for letting it slip between the cracks.

PR #1062, opened 2024-07-23
Add Turborepo

diksipav:
We added [Turborepo](https://turbo.build/) to the` edgedb-js` monorepo. 

The idea is to simplify development and building/testing, especially when some package depends on another packages. Turborepo knows what depends on what and will build packages in the correct order. Previously we (developers) needed to take care of building things in the correct order after applying changes.


packages/auth-express/tsconfig.json
```diff
@@ -1,13 +1,11 @@
 {
   "compilerOptions": {
-    "outDir": "./dist",
```
scotttrinh, line 3: Should this also extend `@repo/tsconfig/client.json`?

packages/auth-remix/package.json
```diff
@@ -16,18 +16,21 @@
...
+    "typescript": "^5.5.2",
+    "edgedb": "*",
+    "@repo/tsconfig": "*"
   },
   "peerDependencies": {
     "edgedb": "^1.3.6"
   },
   "dependencies": {
-    "@edgedb/auth-core": "^0.2.0",
+    "@edgedb/auth-core": "*",
```
scotttrinh, line 33: Can you remind me how this works when we actually publish this package? This gets replaced by a specific version, right?

packages/tsconfig/client.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Client",
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "target": "ESNext",
```
scotttrinh, line 6: I wonder if we should pick a specific target here since `esnext` is a moving target. Maybe `es2022`?

.github/workflows/tests.yml
```diff
@@ -96,42 +87,39 @@ jobs:
         run: |
           echo ACTIVE_EDGEDB_VERSION=$(edgedb query 'select sys::get_version_as_str()' -I test) >> $GITHUB_ENV
 
-      - name: Run query builder tests
+      - name: Generate package - edgedb project init
+        run: |
+          cd packages/generate
+          npm exec edgedb -- project init --non-interactive
```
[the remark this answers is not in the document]
diksipav: Hmm for some reason it didn't work (I kinda think also locally it didn't work for the same reason but can't remember 100%), there was an error in the CI previously that there's edgedb.toml but that project is not initiated. Hmm, I'll try without this again.

packages/auth-express/tsconfig.json
diksipav: hm, I tried to keep all old rules the same... this package only has these rules, client.json and base.json have more rules... but also, I can go again through all rules and see what are the best for our packages

.github/workflows/tests.yml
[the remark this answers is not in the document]
diksipav: hmm... now it passed, weird

packages/auth-express/tsconfig.json
scotttrinh: Yeah, I don't think I intended them to be wildly different, but maybe it makes sense to use the settings that are most popular with the framework's own ecosystem. I don't think there is probably a good candidate for Express like there is for the more full-featured meta-frameworks like Next and Remix, but if it can use some shared config like this, seems fine. In the end I don't think it matters much since consumers will be consuming the emitted `.js`+`.d.ts` files rather than the `.ts` files directly, so as long as we emit reasonable JS code, everything should be good.

The real reason to prefer extending one of our existing configs is just to make it easier/simpler on us to update these when needed.

packages/auth-remix/package.json
scotttrinh: Ahh, right. I was confusing this with the `workspace:` protocol syntax which does some of this publish-time magical replacement: https://yarnpkg.com/features/workspaces#cross-references

Yeah, let's prefer specific versions until we have something like that setup.

packages/tsconfig/base.json
```diff
@@ -2,21 +2,20 @@
...
-    "strict": true,
     "removeComments": true,
     "stripInternal": true,
-    "isolatedModules": true,
-    "esModuleInterop": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "noUnusedLocals": true,
-    "useDefineForClassFields": true,
-    "resolveJsonModule": true
```
scotttrinh, line 18: We want to keep this, right?

scotttrinh (approved):

PR #1074, opened 2024-08-01
Use `queryRequired` for `AtLeastOne` cardinality

### LINKED ISSUE #1072: Queries that return an `AtLeastOne` scalar are inferred as `T[]` instead of `[T, ...T[]]`
Sample query:

```edgeql
select { "a", "b", "c" };
```

Generates the following TypeScript function:

```ts
export type SetOfStrsReturns = [string, ...string[]];

export function setOfStrs(client: Executor): Promise<SetOfStrsReturns> {
  return client.query(`\
select { "a", "b", "c" }`);

}
```

The `query` function gets its generic inferred as `string`, which makes the return value from `query` become `string[]` given the definition of `query`. We do not currently have a query method like `queryRequired` that correctly returns `[T, ...T[]]` instead of `T[]`.

scotttrinh:
Fixes a TypeScript error where `query` was returning `T[]` but the
return type of the function as `[T, ...T[]]` and TypeScript was rightly
complaining that they are incompatible.

Fixes #1072


CarsonF:
Don't we know the number of items in the tuple? and their types. Why don't we strictly enumerate those in the generated TS?

scotttrinh:
> Don't we know the number of items in the tuple?

We do in a tuple, but this is a set, not a tuple. All we know is the type and cardinality.

Querying for a tuple already does the right thing of returning either a tuple type (for tuples) or an object (for named tuples).

CarsonF:
Ah of course 👍

diksipav (approved):

PR #1079, opened 2024-08-09
1078 use right buffer in browser

### LINKED ISSUE #1078: createHttpClient: crypto lib gets imported in the browser
**Describe the bug**
In edgedb version `1.5.6` I believe an error was introduced in this [PR](https://github.com/edgedb/edgedb-js/pull/1015/files).
When creating `httpClient` it will try to import node `crypto` lib which does not exists in the browser, and this is because `Buffer` is imported from ` ../adapter.node`;

**Reproduction**

```typescript
import * as edgedb from "edgedb";

edgedb.createHttpClient();
```

**Expected behavior**
Client should be created without errors.

- `edgedb-js` version (e.g. `^1.5.6`;):

diksipav:
fix #1078 

This is tricky, I spend some time today to track this down (@jaclarke helped with some debugging): 
we need encode/decode for Node, Deno and browser. 
- [Here](https://github.com/edgedb/edgedb-js/issues/1003) the "Buffer not defined" err was thrown because on line 35 `typeof` was missing:` if (Buffer === "function") {`. So native node Buffer was never used until new PR that fixes that but broke browser functioning.

- Require doesn't work in Deno. Test is failing. I believe err should be caught in catch block but for some reason test fails.
 In browser require can work but requiring `node:Buffer` not. Node prefix is not recognised by Webpack, so building is failing. So I guess going with this try/catch block with require in try block will anyway not be good enough.

- CloudFlare needs [nodejs compat flag](https://developers.cloudflare.com/workers/runtime-apis/nodejs/buffer/) if node comp libs are used. I don't know what we want here to be default behaviour for Cloudflare, do we use browser libs by default or we use node compat libs and require users to set Cloudflare to use node libs.

We can revert everything to that first solution when checking if global Buffer was available and fallback to atob/btoa versions. In Deno global Buffer should exists so it will be used, in CloudFlare browser replacements will be used.
Or investigate further how to use node:buffer in deno and cloudflare by default.

### DIFF
--- a/packages/driver/src/primitives/buffer.ts
+++ b/packages/driver/src/primitives/buffer.ts
@@ -19,7 +19,21 @@
 import type char from "./chars";
 import * as chars from "./chars";
 import { LegacyHeaderCodes } from "../ifaces";
-import { Buffer } from "buffer";
+
+function loadBuffer() {
+  let Buffer;
+
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    Buffer = require("node:buffer").Buffer;
+  } catch (_) {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    Buffer = require("buffer/").Buffer;
+  }
+  return Buffer;
+}
+
+const Buffer = loadBuffer();
 
 /* WriteBuffer over-allocation */
 const BUFFER_INC_SIZE = 4096;
--- a/packages/driver/src/primitives/buffer.ts
+++ b/packages/driver/src/primitives/buffer.ts
@@ -26,7 +26,7 @@ export const utf8Decoder = new TextDecoder("utf8");
 let decodeB64: (_: string) => Uint8Array;
 let encodeB64: (_: Uint8Array) => string;
 
-if (typeof globalThis.Buffer === "function") {
+if (typeof Buffer === "function") {
   decodeB64 = (b64: string): Uint8Array => {
     return Buffer.from(b64, "base64");
   };
--- a/packages/driver/src/primitives/buffer.ts
+++ b/packages/driver/src/primitives/buffer.ts
@@ -20,40 +20,47 @@ import type char from "./chars";
 import * as chars from "./chars";
 import { LegacyHeaderCodes } from "../ifaces";
 
-function loadBuffer() {
-  let Buffer;
+export const utf8Encoder = new TextEncoder();
+export const utf8Decoder = new TextDecoder("utf8");
 
-  try {
-    // eslint-disable-next-line @typescript-eslint/no-require-imports
-    Buffer = require("node:buffer").Buffer;
-  } catch (_) {
-    // eslint-disable-next-line @typescript-eslint/no-require-imports
-    Buffer = require("buffer/").Buffer;
-  }
-  return Buffer;
+let decodeB64: (_: string) => Uint8Array;
+let encodeB64: (_: Uint8Array) => string;
+
+try {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const Buffer = require("node:buffer").Buffer;
+
+  decodeB64 = (b64: string) => {
+    return Buffer.from(b64, "base64");
+  };
+  encodeB64 = (data: Uint8Array) => {
+    const buf = Buffer.isBuffer(data)
+      ? data
+      : Buffer.from(data.buffer, data.byteOffset, data.byteLength);
+    return buf.toString("base64");
+  };
+} catch (_) {
+  decodeB64 = function (b64: string) {
+    const binaryString = atob(b64);
+    const size = binaryString.length;
+    const bytes = new Uint8Array(size);
+    for (let i = 0; i < size; i++) {
+      bytes[i] = binaryString.charCodeAt(i);
+    }
+    return bytes;
+  };
+  encodeB64 = function (data: Uint8Array) {
+    const binaryString = String.fromCharCode(...data);
+    return btoa(binaryString);
+  };
 }
 
-const Buffer = loadBuffer();
+export { decodeB64, encodeB64 };
 
 /* WriteBuffer over-allocation */
 const BUFFER_INC_SIZE = 4096;
 
 const EMPTY_BUFFER = new Uint8Array(0);
-
-export const utf8Encoder = new TextEncoder();
-export const utf8Decoder = new TextDecoder("utf8");
-
-const decodeB64 = (b64: string): Uint8Array => {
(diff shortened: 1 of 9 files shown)


scotttrinh:
> I thought we had updated everything to use Uint8Array instead of Buffer so it could work in the browser

Yeah, for what it's worth, I think we should _try_ to use the platform's faster/better APIs when we can rather than taking a least-common denominator approach. The issue here was that I did not consider the browser platform when I worked on #1015, only the Deno environment and was leaning on both Deno and Bun having `node:buffer` compatibility.

We absolutely can just do the base64 decoding ourselves in pure JS, but the old implementation was also subject to unicode problems ([`atob` and `btoa` don't handle unicode](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem)).

At the moment, I think we should attempt to import `node:buffer` and if we fail, fall back to a pure JS implementation. In the future, I think we should rethink our platform strategy to be able to more easily take advantage of platform-specific APIs and do less runtime/buildtime magic.

Would also be a good idea to add some browser-like smoke tests (maybe JSDom is good enough 🤷 ) to ensure I don't break this again 😅

packages/driver/src/primitives/buffer.ts
```diff
@@ -19,28 +19,49 @@
...
+  };
+  encodeB64 = (data: Uint8Array) => {
+    const buf = Buffer.isBuffer(data)
+      ? data
+      : Buffer.from(data.buffer, data.byteOffset, data.byteLength);
+    return buf.toString("base64");
+  };
+} catch (_) {
+  decodeB64 = function (b64: string) {
+    const binaryString = atob(b64);
```
scotttrinh, line 44: BTW, I think this isn't unicode-safe, see [this note on MDN](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). Their implementation of a unicode safe codec is VERY slow, so one thing we could do would be to use the `base64-js` package that `buffer` uses, which is almost as fast as `Buffer`.

We should probably also add a test case for encoding and decoding unicode buffers.

diksipav: @scotttrinh but isn't here the string always base64 encoded string? so it is actually safe... 
I believe that decodeB64 means: decode base64 encoded string? and encodeB64: encode to base64 string.

scotttrinh: Oh, sorry, I think the issue is on the `encodeB64` side: If the bytes are beyond `0xFF`, it'll throw an error at runtime. Maybe this isn't a problem in practice, but we should maybe add a test using the same test string in the linked MDN article?

packages/driver/src/primitives/buffer.ts
```diff
@@ -19,28 +19,49 @@
...
-    ? data
-    : Buffer.from(data.buffer, data.byteOffset, data.byteLength);
-  return buf.toString("base64");
-};
+let decodeB64: (_: string) => Uint8Array;
+let encodeB64: (_: Uint8Array) => string;
+
+try {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const Buffer = require("node:buffer").Buffer;
```
scotttrinh, line 31: The tests are failing because we cannot use `require` in Deno, and Deno will unconditionally try to run this code. We typically can get around this by putting the non-Deno stuff in the `adapter.node.ts` module, and making a Deno-compatible import available in `adapter.deno.ts`, and at `buildDeno`-time, we swap these out in the source. This is what we did before by exporting `Buffer` from the adapter.

I _think_ we should be able to do that if we move these base64-related functions into the `adapter.*` modules and import them instead of importing `Buffer`. And having the `adapter.node.ts` module do this check instead.

packages/driver/src/primitives/buffer.ts
jaclarke: How can a byte be greater than `0xFF`?

scotttrinh: Yep, you're right, not the bytes, but the code points of the input string.

> Since btoa interprets the code points of its input string as byte values, calling btoa on a string will cause a "Character Out Of Range" exception if a character's code point exceeds 0xff. 

The fact that we are dealing with `Uint8Array`s already means I think we're safe here since the encoding itself will take care of ensuring that the binary data is split into utf-8 bytes correctly. Just did some quick in-browser testing and it _does_ look like this works as expected. I'd still like a test to make sure we don't break it in the future, but we can add that test separately.

packages/driver/src/primitives/buffer.ts
diksipav: I don't think this will work in the browser because then we go back to importing things from adapter.node which will try importing node specific libs that are not valid in the browser.

diksipav: this will probably not work because we will import from adapter.node and that file cant be read in browser

scotttrinh: yeah, that's true. I didn't really consider how this module before was totally runtime-agnostic, where the adapter stuff was always imported into server-only contexts.

> https://github.com/edgedb/edgedb-js/issues/1003 the "Buffer not defined" err was thrown because on line 35 typeof was missing: if (Buffer === "function") {. So native node Buffer was never used until new PR that fixes that but broke browser functioning.

Let's just adopt this where we fix the check, and keep most of the old code. I think a more thorough fix will come when we make platform-specific packages so the modules don't need to be so dynamic, but we are still a ways off from working on that.

diksipav: @scotttrinh @jaclarke I think this should be it. Btw I didnt add any new tests, can do that too.

packages/driver/src/primitives/buffer.ts
```diff
@@ -19,28 +19,50 @@
...
-  const buf = Buffer.isBuffer(data)
-    ? data
-    : Buffer.from(data.buffer, data.byteOffset, data.byteLength);
-  return buf.toString("base64");
-};
+let decodeB64: (_: string) => Uint8Array;
+let encodeB64: (_: Uint8Array) => string;
+
+// @ts-ignore: Buffer is not defined in Deno
+if (typeof Buffer === "function") {
```
scotttrinh, line 30: Can you try this form and see if it allows us to remove the `@ts-ignore` here?

```suggestion
if (typeof globalThis.Buffer === "function") {
```

diksipav: aaa yes, globalThis!! 🎉

scotttrinh (approved):
Thanks for working through all of the various ideas and speed bumps here!

PR #1088, opened 2024-08-29
Add support for SNI when connecting to the instance

diksipav:
I will try to add these tomorrow:

Documentation is missing:
- SNI,
- Cloud Profile,
- tlsCA maybe should be documented too,
- `credentials` is not documented.

**More things to be done**: add SNI to credentials.json in the cli, and later read it from there: `Oh and to address this specifically, although very marginal, we may want to update the CLI to include tls_server_name in credentials.json if --tls-server-name was specified while linking an instance.`

Atm CLOUD_PROFILE is taken from env var when using options. And when using env vars it is not used at all?

### DIFF
--- a/packages/driver/src/conUtils.ts
+++ b/packages/driver/src/conUtils.ts
@@ -68,6 +68,7 @@ export interface NormalizedConnectConfig extends PartiallyNormalizedConfig {
   logging: boolean;
 }
 
+// explicit config
 export interface ConnectConfig {
   dsn?: string;
   instanceName?: string;
@@ -84,6 +85,7 @@ export interface ConnectConfig {
   tlsCA?: string;
   tlsCAFile?: string;
   tlsSecurity?: TlsSecurity;
+  tlsServerName?: string;
 
   timeout?: number;
   waitUntilAvailable?: Duration | number;
@@ -124,6 +126,7 @@ type ConnectConfigParams =
   | "cloudProfile"
   | "tlsCAData"
   | "tlsSecurity"
+  | "tlsServerName"
   | "waitUntilAvailable";
 
 export type ResolvedConnectConfigReadonly = Readonly<
@@ -165,6 +168,9 @@ export class ResolvedConnectConfig {
   _tlsSecurity: TlsSecurity | null = null;
   _tlsSecuritySource: string | null = null;
 
+  _tlsServerName: string | null = null;
+  _tlsServerNameSource: string | null = null;
+
   _waitUntilAvailable: number | null = null;
   _waitUntilAvailableSource: string | null = null;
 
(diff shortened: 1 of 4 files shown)


packages/driver/src/conUtils.ts
```diff
@@ -565,6 +587,7 @@ async function parseConnectDsnAndArgs(
       tlsCA: `'tlsCA' option`,
       tlsCAFile: `'tlsCAFile' option`,
       tlsSecurity: `'tlsSecurity' option`,
+      tlsServerName: `tlsServerName option`,
```
scotttrinh, line 590:
```suggestion
      tlsServerName: `'tlsServerName' option`,
```

packages/driver/src/conUtils.ts
```diff
@@ -619,6 +643,7 @@ async function parseConnectDsnAndArgs(
           secretKey: `'EDGEDB_SECRET_KEY' environment variable`,
           tlsCA: `'EDGEDB_TLS_CA' environment variable`,
           tlsCAFile: `'EDGEDB_TLS_CA_FILE' environment variable`,
+          tlsServerName: `EDGEDB_TLS_SERVER_NAME environment variable`,
```
scotttrinh, line 646:
```suggestion
          tlsServerName: `'EDGEDB_TLS_SERVER_NAME' environment variable`,
```

packages/driver/src/conUtils.ts
```diff
@@ -663,6 +688,10 @@ async function parseConnectDsnAndArgs(
           .readFileUtf8(stashDir, "database")
           .then((name) => name.trim())
           .catch(() => undefined),
+        serverUtils
+          .readFileUtf8(stashDir, "branch")
+          .then((name) => name.trim())
+          .catch(() => undefined),
```
scotttrinh, line 694: We're not assigning this to a variable on line 682. Are we missing a test case for this in the shared tests?

diksipav: oh yes, u are right

packages/driver/src/conUtils.ts
```diff
@@ -694,13 +694,26 @@ async function parseConnectDsnAndArgs(
           .catch(() => undefined),
       ]);
 
+      let database = _database;
+
+      if (database !== undefined && branch !== undefined) {
+        if (database !== branch) {
+          throw new InterfaceError(
+            "Both database and branch exist in the config dir and don't match.",
```
scotttrinh, line 702: Do we have a shared client test case for this?

scotttrinh (approved):

packages/driver/src/conUtils.ts
diksipav: I don't think that we have for this specific case, but possibillity for this is very rare.

scotttrinh (approved):

packages/driver/src/conUtils.ts
scotttrinh: yeah, the test cases cover (currently) rare or impossible cases just to make sure all clients treat different edge cases the same way. probably worth adding a case!

PR #1090, opened 2024-08-31
Polymorphic queries -> discriminated union types

### LINKED ISSUE #454: Better return types for `.__type__.name` and polymorphic queries if possible

CarsonF:
I revived #508

Fixes #454

Probably a breaking change since the types are getting stricter.


CarsonF:
I restored some of the old functionality, and gated the stricter types behind an opt-in via declaration merging. Just looking for ways to get this out without a major. But open to reverting it or other suggestions.

packages/generate/src/edgeql-js/generateObjectTypes.ts
```diff
@@ -175,6 +175,18 @@ export const getStringRepresentation: (
...
+  for (const type of types.values()) {
+    if (type.kind === "object" && !type.is_abstract) {
+      for (const base of type.bases) {
+        if (!descendents.has(base.id)) {
+          descendents.set(base.id, new Set());
+        }
+        descendents.get(base.id)!.add(type.name);
+      }
+    }
+  }
```
CarsonF, line 188: This doesn't appear to handle grand-children/parents.

CarsonF:
Some scenario in my codebase is producing `{ __typename: string, ... } |` which breaks the narrowing with the union-ed concrete types. Digging...

CarsonF:
Even though the implicit computed `__typename` produce the right string literals, these cannot yet be explicitly referenced.
```ts
e.shape(e.X, scope => ({
  __typename: scope.__type__.name
})
```
still produces a `string` not `"default::X"`

packages/generate/src/syntax/path.ts
```diff
@@ -39,7 +41,17 @@ type getChildOfObjectTypeSet<
...
-  Root["__element__"]["__pointers__"][ChildKey]["target"],
+  ChildKey extends "name"
+    ? Root extends { [typenameSymbol]: string }
+      ? ScalarType<
+          "std::str",
+          TypesystemOptions["future"]["strictTypeNames"] extends true
+            ? Root[typeof typenameSymbol]
+            : string,
+          Root[typeof typenameSymbol]
+        >
```
CarsonF, line 52: I'm not sure I understand the difference between Scalar's `TsType` & `TsArgType`. I was thinking that they were the output / input types respectively.

I don't understand why the input type would be called out here explicitly and set to the literal string - where/how would that be used be as an input?

I figured the output type remained as a `string` for BC reasons. So I created another future option to make it strict.
This is for this btw:
```ts
scope => ({ t: scope.__type__.name })
```

But then I was thinking about this case
```ts
scope => ({ __type__: { name: true } })
```
Which had already made the output type a strict string.

So I thought the other case was a miss, but there is some explicit code here.

I guess all that to say I could just do this:
```suggestion
      ? ScalarType<"std::str", Root[typeof typenameSymbol]>
```

integration-tests/lts/select.test.ts
```diff
@@ -1136,6 +1136,35 @@ SELECT __scope_0_defaultPerson {
...
   });
 
+  test.skip("polymorphic from type intersection", async () => {
+    const query = e.select(e.Movie.characters, (person) => ({
+      heroMovieCount: person.is(e.Hero).number_of_movies,
+      heroInfo: e.select(person.is(e.Hero), (hero) => ({
+        number_of_movies: true,
+        numMovies: hero.number_of_movies,
+      })),
+    }));
```
CarsonF, line 1146: This PR doesn't yet create a discriminated union from this 😢

CarsonF: This is result is the same in master. Since it hasn't regressed in this PR, I'll leave it for a separate PR.

packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -1395,6 +1398,9 @@ function shapeToEdgeQL(
   if (lines.length === 0 && injectImplicitId) {
     addLine("id");
   }
+  if (hasPolyEl && !seen.has("__typename")) {
+    addLine("__typename := .__type__.name");
+  }
```
CarsonF, line 1403: The new functionality implies a `__typename` field when needed. If the user specifies this and it doesn't match what is assumed then the types will be wrong. We could check the `__typename` value here if it is already present (from a user) and confirm it is `.__type__.name` or throw.
I couldn't figure out a clean way to do this because the actual generated EdgeQL explicitly references the scope, which I couldn't figure out how to access.
```edgeql
single __typename := __scope_1_defaultMovie.__type__.name,
```

scotttrinh (approved):
Very solid 👏 Thanks for the huge effort here. Love that we now have some feature-flagging/opt-in machinery set up in the typesystem, too, that seems generally very useful.

scotttrinh:
One thing I want to think about is maybe having some tests with the legacy behavior just to catch any regressions we make between now and when we switch the new behavior to the default in a future major release. I can add those.

packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -1299,6 +1301,7 @@ function shapeToEdgeQL(
     if (val.__kind__ === ExpressionKind.PolyShapeElement) {
       polyType = val.__polyType__;
       val = val.__shapeElement__;
+      hasPolyEl = true;
```
scotttrinh, line 1304: Ah, we don't actually have a runtime representation of the type-level "opt in" for this behavior, so this ends up still being an arguably breaking change since we will be returning more. Definitely arguable as to whether or not returning strictly more data is breaking, but anyone who currently has tests for the existing behavior will get failures.

I wonder if we can make a runtime configuration, and actually _infer_ the type-level config from that runtime config?

CarsonF: I probably wouldn't say that adding an additional field is a breaking change. User tests may not break. Only if they are asserting exactly or using snapshots.

But I can see how it's not black/white. IMO none of this stuff is. kinda like TS changes lol

packages/generate/src/syntax/path.ts
CarsonF: @scotttrinh thoughts here? I don't think we need a flag for this & just use my suggestion

packages/generate/src/syntax/toEdgeQL.ts
scotttrinh: > User tests may not break. Only if they are asserting exactly or using snapshots.

I think this is _pretty_ common: make a query (or use a data layer more likely) and assert that the returned data is precisely these objects. I agree in general that if you are introducing new properties/data in your own application, maybe you don't consider that a breaking change, but this might warrant a little more care to avoid tripping up people when upgrading point releases.

CarsonF: Are you looking for a CLI flag for this as well?

scotttrinh: Yeah, I think we should add it, like `--enable-polymorphism-as-discriminated-unions` as silly as that sounds. This is non-trivial, but I think it's worth it to avoid going with a major version bump here. I'm honestly open to the major version bump, but that will absolutely delay getting this released.

If you think you have the bandwidth to add the CLI flag, let me know. Otherwise, I'll stick it in the queue and try to get to it (or maybe have @diksipav pick this up) ASAP.

CarsonF:
```
--future-polymorphism-as-discriminated-unions
```

packages/generate/src/syntax/path.ts
scotttrinh: Good question. I'm not sure, but I'm fine with going with the suggested change for now and looking at what the fallout is.

packages/generate/src/syntax/toEdgeQL.ts
scotttrinh: yeah, that was my first thought, but then I thought maybe in the next major we could allow you to still opt-into the old behavior, but now I'm thinking there is no advantage to keeping the old behavior around in the next major.

CarsonF: Yeah I don't see a reason to keep the old behavior long term.

packages/generate/src/syntax/typesystem.ts
```diff
@@ -326,9 +359,19 @@ type shapeElementToTs<Pointer extends PropertyDesc | LinkDesc, Element> = [
...
+                typeutil.flatten<
+                  ([ParentTypeName] extends [string]
+                    ? Element extends { name: true }
+                      ? { name: ParentTypeName }
+                      : unknown
+                    : unknown) &
+                    computeObjectShape<
+                      Pointer["target"]["__pointers__"] & Pointer["properties"],
+                      Element,
+                      TypesystemOptions["future"]["strictTypeNames"] extends true
```
scotttrinh, line 371: So like,

```suggestion
                      (typeof future["strictTypeNames"]) extends true
```

etc

packages/generate/src/syntax/typesystem.ts
```diff
@@ -12,6 +12,33 @@ import type { cardutil } from "./cardinality";
...
+     * Opt-in to strict __type__.name string literal unions.
+     */
+    strictTypeNames: SetTypesystemOptions extends {
+      future: { strictTypeNames: true };
+    }
+      ? true
+      : false;
+  };
+};
+
```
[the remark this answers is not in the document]
CarsonF: I think it depends on what the DX goal is.

If the goal is to just use the cli flag and expect that `future.ts` is statically overridden, then no we don't need `TypesystemOptions` or the declaration merging on `SetTypesystemOptions`.

If the goal is to allow this to be set dynamically. Then yes we still need it.
In that case, devs would set the futures to true.
```ts
future.polymorphismAsDiscriminatedUnions = true;
future.strictTypeNames = true;
```
And then denote to TS that they intend to do that with the declaration merge.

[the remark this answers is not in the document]
scotttrinh: I don't think it makes sense for this to be dynamic in any way at all. You're either opting into this future behavior and will refactor all of your code to match it, or you're not. I also care about making sure the runtime behavior and types are aligned, so allowing (or really I guess encouraging) both seems dangerous.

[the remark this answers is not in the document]
CarsonF: Then yeah we don't need any of this. Just base it all of the futures object like you suggested

packages/generate/src/cli.ts
```diff
@@ -238,6 +238,9 @@ const run = async () => {
       case "--no-update-ignore-file":
         options.updateIgnoreFile = false;
         break;
+      case "--future":
+        options.future = true;
```
CarsonF, line 242: I don't think this is sufficient. We need to be able to specify which futures you are opting into. I could be ok with future A and have updated my codebase to account for it. But then future B is added and I would be automatically opted into it as well which is a breaking change I wasn't ready for.

scotttrinh: Yeah, @diksipav and I talked a bit about this yesterday and here are my thoughts:

1. From the perspective of a general feature of the generator package, this `future` behavior is similar to TypeScript's own `strict` behavior, which if you opt into directly is breaking between versions. It stands for some set of more granular configuration values, so I think this _is_ useful in general.
2. From the perspective of this future behavior, I expect most (all?) folks to want to opt into both, but also would be fine exposing granular flags for each here, too. I don't want to add tests for each permutation since we plan on making this the _only_ behavior in a future breaking change, but that does mean there might be bugs or poorly understood side-effects of having one or the other enabled, but not both.

I don't think we need to block on this, though, and I'm fine with trying to land this as-is now. I'll leave it up to @diksipav to decide if she wants to address this in this PR specifically, or as a follow-up PR.

diksipav: hey lets merge this (once it's ready) and I'll create right away a follow up PR with individual future flags

diksipav: @scotttrinh I'll merge this and start working on the PR with individual flags, if u think we should "swap" tests will do it in the next PR

PR #1091, opened 2024-09-03
Add support for protocol v2

diksipav:


packages/driver/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "@repo/tsconfig/base.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "noUnusedLocals": false
```
jaclarke, line 6: Can we just remove the unused local (or add a tsignore) instead of setting this to false?

packages/driver/src/conUtils.ts
```diff
@@ -643,7 +643,7 @@ async function parseConnectDsnAndArgs(
           secretKey: `'EDGEDB_SECRET_KEY' environment variable`,
           tlsCA: `'EDGEDB_TLS_CA' environment variable`,
           tlsCAFile: `'EDGEDB_TLS_CA_FILE' environment variable`,
-          tlsServerName: `'EDGEDB_TLS_SERVER_NAME' environment variable`,
+          tlsServerName: `EDGEDB_TLS_SERVER_NAME environment variable`,
```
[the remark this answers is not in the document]
diksipav: yes, I don't recall what happened here

packages/driver/tsconfig.json
diksipav: hm there are quite a lot of unused vars in codecs/registry and I can't find a way to supres TS checks for the whole file just for this rule, but all these unused vars are also checked with eslint if they are not prefixed with _

scotttrinh: Yeah, it _is_ annoying, but I agree with @jaclarke that we should keep this usage. I'd rather us use `ts-expect-error`, fwiw. In the future, we'll actually use these fields and we can remove these rule exceptions.

diksipav: @scotttrinh ah okay, I think u said few months ago to try to remove all `ts-expect-error` lines, and use _ for all unused things (maybe it was for another project). I'll revert for this use case. 

Also, I may am a bit confused what we wanna have as a TS error and what as a lint error.

packages/driver/src/conUtils.ts
```diff
@@ -587,7 +587,7 @@ async function parseConnectDsnAndArgs(
       tlsCA: `'tlsCA' option`,
       tlsCAFile: `'tlsCAFile' option`,
       tlsSecurity: `'tlsSecurity' option`,
-      tlsServerName: `'tlsServerName' option`,
+      tlsServerName: `tlsServerName option`,
```
scotttrinh, line 590: another stray edit 😅

diksipav: it's weird, I kinda remember you already mentioned these fixes in some PR and I've done them...

PR #1110, opened 2024-09-30
Add individual future cli flags & behaviour

diksipav:


packages/generate/src/edgeql-js.ts
```diff
@@ -196,9 +208,26 @@ export async function generateQueryBuilder(params: {
...
+  ) {
+    content += futureFileContent;
+  } else {
+    if (options.strictTypeNames) {
+      content += strictTypeNamesFileContent;
+    }
+    if (options.polymorphismAsDiscriminatedUnions) {
+      content += polymorphismAsDiscriminatedUnionsFileContent;
+    }
+  }
```
CarsonF, line 225: Probably need to create this content dynamically from future object.
```ts

const future = {
  polymorphismAsDiscriminatedUnions: false,
  strictTypeNames: false,
  ...options.future,
};

content += `export const future = ${JSON.stringify(future, undefined, 2)} as const;
`;

File: packages/generate/src/commandutil.ts [hunk 2]
```diff
@@ -15,6 +15,8 @@ export interface CommandOptions {
   updateIgnoreFile?: boolean;
   useHttpClient?: boolean;
   future?: boolean;
+  strictTypeNames?: boolean;
+  polymorphismAsDiscriminatedUnions?: boolean;
```
CarsonF, line 19: Maybe something like
```suggestion
  future?: Partial<Record<keyof typeof future, boolean>>;
```
where we could pull the keys from a default future object

CarsonF (changes requested):

CarsonF:
Hope you don't mind me giving feedback 😄

packages/generate/src/cli.ts
```diff
@@ -241,6 +241,12 @@ const run = async () => {
       case "--future":
         options.future = true;
         break;
```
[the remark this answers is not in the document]
diksipav: @CarsonF is the idea here to provide every flag separately because I thought we also want one flag that includes all future features (I'd personally probably use just that)?

packages/generate/src/edgeql-js.ts
diksipav: yes, will do sth like this, was thinking to do it if we add more flags, but can do it now too

packages/generate/src/cli.ts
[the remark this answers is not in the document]
CarsonF: Gotcha. That's cool with me!

packages/generate/src/cli.ts
```diff
@@ -443,6 +458,10 @@ OPTIONS:
         Do not prompt to update gitignore with generated code
     --future
         Include future features
+    --strictTypeNames
+        Return the exact string literal for .__type__.name instead of a general string type
+    --polymorphismAsDiscriminatedUnions
+        Use a discriminated union as the return type for polymorphic queries, where each member includes __typename
```
CarsonF, line 464:
```suggestion
    --future-strict-type-names
        Return the exact string literal for .__type__.name instead of a general string type
    --future-polymorphism-as-discriminated-unions
        Use a discriminated union as the return type for polymorphic queries, where each member includes __typename
```

diksipav: tnx, my brain doesn't work great after 8pm :D

CarsonF (approved):
LGTM!

PR #1111, opened 2024-10-02
1109 polymorphic types union

### LINKED ISSUE #1109: Polymorphic field is not present
**Code**
The code causing the error.

https://github.com/iforge-uos/ignis/blob/feature/protein-filled-dinner/apps/anvil/src/training/training.service.ts#L39 

```typescript
  async getTraining(id: string, editing?: boolean): Promise<training.Training> {
    const x = e
      .assert_exists(
        e.select(e.training.Training, () => ({
          ...(editing
            ? {
                ...e.training.Training["*"],
                sections: TrainingSection,
              }
            : {...}),
          enabled: true,
          rep: { id: true, name: true, description: true },
          filter_single: { id },
        })),
      );
    console.log(x.toEdgeQL());
  }
```

refs: https://discord.com/channels/841451783728529451/1265373494561083583

**Schema**

https://github.com/iforge-uos/ignis/tree/feature/protein-filled-dinner/apps/anvil/dbschema/training.esdl

**Generated EdgeQL**

None, error generated

**Error or desired behavior**

```
anvil:dev:     'Error: Field "index" does not exist in training::Question | training::TrainingPage\n' +
anvil:dev:       '    at shapeToEdgeQL (/Users/James/Programming/ignis/apps/anvil/dbschema/edgeql-js/toEdgeQL.ts:1338:15)\n' +
anvil:dev:       '    at renderEdgeQL (/Users/James/Programming/ignis/apps/anvil/dbschema/edgeql-js/toEdgeQL.ts:968:11)\n' +
anvil:dev:       '    at shapeToEdgeQL (/Users/James/Programming/ignis/apps/anvil/dbschema/edgeql-js/toEdgeQL.ts:1375:26)\n' +
anvil:dev:       '    at Array.map (<anonymous>)\n' +
anvil:dev:       '    at renderEdgeQL (/Users/James/Programming/ignis/apps/anvil/dbschema/edgeql-js/toEdgeQL.ts:1148:32)\n' +
anvil:dev:       '    at Proxy.$toEdgeQL (/Users/James/Programming/ignis/apps/anvil/dbschema/edgeql-js/toEdgeQL.ts:319:16)\n' +
anvil:dev:       '    at TrainingService.getTraining (/Users/James/Programming/ignis/apps/anvil/src/training/training.service.ts:56:8)\n' +
anvil:dev:       '    at TrainingController.getTraining (/Users/James/Programming/ignis/apps/anvil/src/training/training.controller.ts:41:33)'
anvil:dev:   ]
```

- OS:
- EdgeDB version (e.g. `2.0`): 5.4+06b93a6
- EdgeDB CLI version (e.g. `2.0`): EdgeDB CLI 5.2.0-dev.1166+b5e9df3
- `edgedb-js` version (e.g. `0.20.10`;): 1.5.12
- `@edgedb/generate` version (e.g. `0.0.7`;): v0.5.3

diksipav:
fixes #1109 

We allow union of multiple types in schema computed fields. Some members of the union can extend some other common types, but not all members have to.


jaclarke (changes requested):
@diksipav I think this needs some tests

integration-tests/stable/dbschema/default.esdl
```diff
@@ -8,6 +8,14 @@ module default {
 
   type User {
     identity: ext::auth::Identity;
+
+    multi movies: Movie;
+    multi shows: Show;
+
+    multi watching_list := ( 
+      select .movies union .shows
```
jaclarke, line 16: Can we add a third item to the union, to test that we properly union more than two types. And if edgedb allows it, can it also not inherit from `Content` but have at least one of same fields to check we're handling common fields not just common ancestors?

integration-tests/stable/select.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const WatchingList = e.shape(e.User.watching_list, () => ({
+      index: true,
+      title: true,
+      ...e.is(e.Movie, {
+        plot: true,
+      }),
+      ...e.is(e.Show, {
+        seasons: true,
+      }),
+    }));
```
jaclarke, line 28: I think we should have an expect error test trying to select a field that only exists on one of the union items outside the `e.is(...)` to check we reject it.

integration-tests/stable/dbschema/default.esdl
diksipav: hmm how do u mean common fields? if they have the same field but that field is not inherited, it is not "common" imo

jaclarke: Yes, you probably shouldn't write an actual schema like that, and I'm not 100% sure off the top of my head if edgedb supports it (I think it does), but if it does the querybuilder should still handle that just in case.

diksipav:
In this test:
```
const q = e.select(e.User, () => ({
      all_media: (all_media) => ({
        title: true,
        ...e.is(e.Content, { year: true }),
        ...e.is(e.Show, {
          seasons: true,
        }),
        plot: e.op(
          all_media.is(e.Movie).plot,
          "??",
          all_media.is(e.Documentary).plot,
        ),
      }),
    }));

    type qT = $infer<typeof q>;

    tc.assert<
      tc.IsExact<
        qT,
        {
          all_media: (
            | {
                title: string;
                plot: string;
                __typename: "default::Documentary";
              }
            | {
                title: string;
                plot: string;
                year: number;
                __typename: "default::Movie";
              }
            | {
                title: string;
                plot: string;
                year: number;
                seasons: number;
                __typename: "default::Show";
              }
          )[];
        }[]
      >
    >(true);
```
I think that type of the `plot` inside the `Show` should be `null`. Show doesn't have plot. Or you think it should be `string | null` @jaclarke? @scotttrinh what do u think? 

I can fix this in a separate PR.

integration-tests/stable/dbschema/migrations/00007-m1n37rk.edgeql
```diff
@@ -0,0 +1,18 @@
+CREATE MIGRATION m1n37rkzwbdej2nsfr73ujxlwjaum7zhqsaovf7l2sdw2g4dnayfqq
```
jaclarke, line 1: Can these all be squashed into a single migration.

jaclarke:
> I think that type of the `plot` inside the `Show` should be `null`. Show doesn't have plot. Or you think it should be `string | null` @jaclarke? @scotttrinh what do u think?
> 
> I can fix this in a separate PR.

Yeah, it should be `plot: string | null` in every object of the union, because the cardinality of `plot: e.op(all_media.is(e.Movie).plot, "??", all_media.is(e.Documentary).plot)` should be `AtMostOne`. My reasoning is the cardinality of `plot` in both `Movie` and `Documentary` is exactly `One` because it's required, then `.is(...)` should make the cardinality of every field of the object it's called on optional, so `AtMostOne` for `plot`, then `e.op(..., '?', ...)` should return the cardinality of the first arg if it's not optional, else just the cardinality of the second arg.

diksipav:
> > I think that type of the `plot` inside the `Show` should be `null`. Show doesn't have plot. Or you think it should be `string | null` @jaclarke? @scotttrinh what do u think?
> > I can fix this in a separate PR.
> 
> Yeah, it should be `plot: string | null` in every object of the union, because the cardinality of `plot: e.op(all_media.is(e.Movie).plot, "??", all_media.is(e.Documentary).plot)` should be `AtMostOne`. My reasoning is the cardinality of `plot` in both `Movie` and `Documentary` is exactly `One` because it's required, then `.is(...)` should make the cardinality of every field of the object it's called on optional, so `AtMostOne` for `plot`, then `e.op(..., '?', ...)` should return the cardinality of the first arg if it's not optional, else just the cardinality of the second arg.

Hm yes i agree. Its anyway kind of weird how very diff the type is between different usages:
`seasons: media.is(e.Show).seasons`
and 
```
 ...e.is(e.Show, {
      seasons: true,
}),
```

jaclarke (approved):

PR #1113, opened 2024-10-04
Add support for PostGIS types

jaclarke:


.github/workflows/tests.yml
```diff
@@ -82,11 +82,17 @@ jobs:
         with:
           instance-name: test
           server-version: ${{ matrix.edgedb-version }}
+          cli-version: ${{ matrix.edgedb-version == 'nightly' && 'nightly' || 'stable' }}
```
scotttrinh, line 85: 👍

packages/driver/src/cli.mts
```diff
@@ -341,6 +341,7 @@ async function downloadFile(url: string | URL, path: string) {
     throw new Error(`  - Download failed: ${response.statusText}`);
   }
 
+  // @ts-ignore
```
scotttrinh, line 344: Can we use `@ts-expect-error Some note about why` here? We have a lint rule for this, but I wonder if the `.mts` extension is throwing off the matcher. I'll take a look at that.

packages/driver/src/primitives/buffer.ts
```diff
@@ -197,6 +197,14 @@ export class WriteBuffer {
     return this;
   }
 
+  writeDeferredSize(): () => void {
```
scotttrinh, line 200: ?

packages/driver/src/cli.mts
jaclarke: I have no idea why that was there. It seems to be fine without it now.

packages/driver/src/codecs/postgis.ts
```diff
@@ -0,0 +1,490 @@
...
+    const points = poly.rings[0].points;
+    return new Box2D([points[0].x, points[0].y], [points[2].x, points[2].y]);
+  }
+}
+
+export class PostgisBox3dCodec extends ScalarCodec implements ICodec {
+  encode(buf: WriteBuffer, object: any): void {
+    if (!(object instanceof Box3D)) {
+      throw new InvalidArgumentError(
+        `a Box2D object was expected, got "${object}"`,
```
dnwpark, line 98: Should `Box3D`?

PR #1115, opened 2024-10-07
Add warnings handler

jaclarke:
See https://github.com/edgedb/edgedb/issues/7822 for details.

### DIFF
--- a/packages/driver/src/baseConn.ts
+++ b/packages/driver/src/baseConn.ts
@@ -24,7 +24,7 @@ import type { CodecsRegistry } from "./codecs/registry";
 import { EmptyTupleCodec, EMPTY_TUPLE_CODEC, TupleCodec } from "./codecs/tuple";
 import { versionGreaterThanOrEqual } from "./utils";
 import * as errors from "./errors";
-import { resolveErrorCode } from "./errors/resolve";
+import { resolveErrorCode, errorFromJSON } from "./errors/resolve";
 import type {
   QueryOptions,
   ProtocolVersion,
@@ -102,6 +102,7 @@ export type ParseResult = [
   capabilities: number,
   inCodecBuffer: Uint8Array | null,
   outCodecBuffer: Uint8Array | null,
+  warnings: errors.EdgeDBError[],
 ];
 
 export type connConstructor = new (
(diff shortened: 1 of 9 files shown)


packages/driver/src/baseConn.ts
```diff
@@ -1230,7 +1272,7 @@ export class BaseRawConnection {
         (!inCodec && args !== null) ||
         (this.stateCodec === INVALID_CODEC && state !== Session.defaults())
       ) {
-        [card, inCodec, outCodec] = await this._parse(
+        [card, inCodec, outCodec, _, _, _, warnings] = await this._parse(
```
scotttrinh, line 1275: btw, I _think_ these `_` variable names are unnecessary? Maybe eslint would complain if you did this, but I think this is equivalent:

```suggestion
        [card, inCodec, outCodec,,,, warnings] = await this._parse(
```

Just a tiny nit, not even something I'm suggesting we do here, just putting it out there.

packages/driver/src/errors/base.ts
```diff
@@ -34,17 +35,15 @@ export class EdgeDBError extends Error {
     return this.constructor.name;
   }
 
-  hasTag(tag: symbol): boolean {
-    // Can't index by symbol, except when using <any>:
-    //   https://github.com/microsoft/TypeScript/issues/1863
-    const error_type = this.constructor as typeof EdgeDBError as any;
-    return Boolean(error_type.tags?.[tag]);
+  hasTag(tag: tags): boolean {
+    const error_type = this.constructor as typeof EdgeDBError;
```
scotttrinh, line 39: 🎉

packages/driver/test/client.test.ts
```diff
@@ -2080,6 +2083,43 @@ test("pretty error message", async () => {
...
+test("warnings handler", async () => {
+  if (getEdgeDBVersion().major < 6) return;
+
+  let client = getClient();
+
+  try {
+    let warnings: EdgeDBError[] | null = null;
+    client = client.withWarningHandler((_warnings) => (warnings = _warnings));
+
+    await expect(client.query("select _warn_on_call();")).resolves.toEqual([0]);
```
scotttrinh, line 2095: This is some kind of built-in?

scotttrinh (approved):
👑

packages/driver/src/baseConn.ts
jaclarke: Huh, I didn't know you could do that 🌈⭐. Though I think I prefer the underscores as it looks more intentional to me, and not that I've just typoed it.

packages/driver/test/client.test.ts
jaclarke: Yes, Sully 'documented' it here: https://github.com/edgedb/edgedb/pull/7823. I think it only exists in test mode/dev mode though.

PR #1116, opened 2024-10-09
Update cardinality of is expression method

diksipav:


diksipav:
Tnx @jaclarke for the help.

PR #1121, opened 2024-10-17
EdgeDB AI provider

diksipav:
**I think we  need a call to discuss what we want to support with the provider and extension (or maybe I'm just not aligned with the goals and provider should just be built to support current capabilities of the extension).**

Examples of how we can use this with the Vercel SDK:
[the completion route](https://github.com/edgedb/edgedb-examples/blob/add-nextjs-ai-example/nextjs-ai-new/app/api/completion/route.ts), 
[the chat route](https://github.com/edgedb/edgedb-examples/blob/add-nextjs-ai-example/nextjs-ai-new/app/api/chat/route.ts)

This PR has the basic support for completion and chat that we also support with edgedb-js ai lib. I made it to work with the AI ext we have and its current capabilities. And it can be used as this but definitely needs improvements in order to have fully compatible interfaces with Vercel `LanguageModelV1` provider.

**TODO:** I need to add README and update some identifiers and interface names throughout the provider since we are not really a language model.

**QUESTIONS:**

1. We don't support any of LLM settings
 maxTokens,
 temperature,
 topP,
 topK,
 frequencyPenalty,
 presencePenalty,
 stopSequences,
 responseFormat,
 seed etc,
 so I don't know do we wan't to add support for these or should I return an error from the provider if user tries to set any of these.

2. Do we want to support setting for `safe_prompt`?
3. Do we support structured outputs?
4. Do we have support for tools?
5. In the output our RAG returns it doesn't return:
```
usage: {
  promptTokens: ...,
  completionTokens: ...,
},
 ```
and some other things that are used/returned from the Vercel `LanguageModelV1` provider. So I return 0 for these 2 values for now from the provider.

6. Do we want to support different settings for different LLMs or we navigate more towards supporting basic works and settings all LLMs support. I mean some of them even support images and audio.

### CODE REVIEW

**Comment by diksipav** [comment]
Here are the examples of how we can use this with the Vercel SDK:
[the completion route](https://github.com/edgedb/edgedb-examples/blob/add-nextjs-ai-example/nextjs-ai-new/app/api/completion/route.ts), 
[the chat route](https://github.com/edgedb/edgedb-examples/blob/add-nextjs-ai-example/nextjs-ai-new/app/api/chat/route.ts)

**Comment by scotttrinh** [comment]
> so I don't know do we wan't to add support for these or should I return an error from the provider if user tries to set any of these.

https://github.com/vercel/ai/blob/65e108f94f40b80890b00ccc12eeb04c792a4b92/packages/mistral/src/mistral-chat-language-model.ts#L71-L78

Yeah it looks like there is some expectation that we return some warnings.

> Do we want to support setting for safe_prompt?
>
> Do we support structured outputs?
>
> Do we have support for tools?

We can skip this stuff for now.

> So I return 0 for these 2 values for now from the provider.

Looks like the Mistral provider uses `NaN` 😂 https://github.com/vercel/ai/blob/65e108f94f40b80890b00ccc12eeb04c792a4b92/packages/mistral/src/mistral-chat-language-model.ts#L242-L245

> Do we want to support different settings for different LLMs or we navigate more towards supporting basic works and settings all LLMs support. I mean some of them even support images and audio.

I think for now we support just the API that our AI extension exposes and we treat it as if we're a really restricted LLM that has built in RAG. Developers will be configuring the AI extension itself to take advantage of features in the underlying model, but we'll need to expose it directly from the extension in order for this provider to take advantage of it.

File: packages/driver/src/cli.mts [hunk 1]
```diff
@@ -95,46 +95,48 @@ async function main(args: string[]) {
 async function whichEdgeDbCli() {
   debug("Checking if CLI is in PATH...");
   const locations = await which("edgedb", { nothrow: true, all: true });
-  for (const location of locations) {
-    const actualLocation = await fs.realpath(location);
-    debug(
-      `  - CLI found in PATH at: ${location} (resolved to: ${actualLocation})`,
-    );
+  if (locations) {
```
scotttrinh, line 98: Small nit: maybe just treat this as an early return instead of wrapping this whole section in an `if`?

```
if (locations == null) {
  debug("  - No CLI found in PATH.");
  return null;
}
```

I'm annoyed that we aren't catching this in our tests 🫤

diksipav:
RAG provider should be fine. But I think we need to extend capabilities of the AI binding. It should support other parameters besides the query and also function calling.

diksipav:
I created a new cleaner PR for this: https://github.com/edgedb/edgedb-js/pull/1129.

PR #1124, opened 2024-10-29
Add codecs for `ext::pgvector::halfvec`+`ext::pgvector::sparsevec` and add `SparseVector` datatype

jaclarke:


scotttrinh (approved):
🏆

scotttrinh:
> Does this seem ok?

Yeah, I think this is great, and we can remove it once it ever hits baseline support, but that's likely years away at this point. FWIW, the official pgvector package itself just uses a number array for both regular and sparse vectors so this is already a huge improvement over that.

PR #1129, opened 2024-11-06
Add ai rag provider

diksipav:


packages/edgedb-rag-sdk/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@edgedb/rag-sdk",
```
scotttrinh, line 2: This name feels a bit too vague: What about `@edgedb/ai-vercel-provider`?

diksipav: hmm we can include vercel in the name, none of the [community providers](https://sdk.vercel.ai/providers/community-providers) do that tho, maybe just ai-provider is better and having in the package.json description, in the readme and in the docs the explanation that it is for vercel is enough... but also, we can include it in the name, I just don't think it's so crucial

scotttrinh: Portkey's provider is called `@portkey-ai/vercel-provider`, but yeah, most of them are called `something-ai-provider`, so `@edgedb/ai-provider` seems _fine_ but also a little weird considering we have `@edgedb/ai`. Doesn't seem clear that this is a provider for the vercel `ai` package, but maybe that's really just a problem with vercel calling the package `ai` instead of `@vercel/ai` 😅

diksipav: oh yes, I think their package should have included vercel, u go on npm and just install ai, it is weird... regarding edgedb, maybe we should ask more people for their thoughts

diksipav: hmm also I dont know will we write more providers in the future, so actually I believe is ok to call it  @edgedb/ai-vercel-provider or @edgedb/vercel-ai-provider, but I am not sure which one of these two is better

scotttrinh: > But now I have a problem with how do we call our interfaces and functions, atm we use: EdgeDBRagProvider, createEdgeDBRag, default instance is edgedbRag... Other LLMs call it mistralAI for example but they are AI , EdgeDB is not. How it should look on the user side?

Good question. We have both a language model (enhanced with RAG) and an embeddings model, right? I think mostly the pattern seems to be (looking at the OpenAI docs):

```ts
import { createEdgeDB } from "@edgedb/vercel-ai-provider";

const edgedb = createEdgeDB();

const languageModel = edgedb("some-language-model-name");

const embeddingsModel = edgedb.textEmbeddingModel("some-embeddings-model-name");
```

I don't think the distinction between this being a RAG-enhanced language model is super important at the naming level, but we should certainly spend plenty of time explaining it in the documentation and README. This isn't something I feel very strongly about so if anyone else feels like it's confusing that we don't mention RAG in the naming, I'm fine with it. `edgedDBRag.textEmbeddingModel` feels a little strange to me, but not a big deal.

packages/ai/src/core.ts
```diff
@@ -68,11 +63,17 @@ export class EdgeDBAI {
     });
   }
 
-  private async fetchRag(request: Omit<RAGRequest, "model" | "prompt">) {
+  private async fetchRag(request: any) {
```
scotttrinh, line 66: Is there something more useful than `any` we can use here? Like maybe `unknown`?

packages/vercel-ai-provider/src/edgedb-provider.ts
```diff
@@ -0,0 +1,90 @@
...
+  textEmbeddingModel: (
+    modelId: EdgeDBEmbeddingModelId,
+    settings?: EdgeDBEmbeddingSettings,
+  ) => EmbeddingModelV1<string>;
+}
+
+export async function createEdgeDB(client: Client): Promise<EdgeDBProvider> {
+  const connectConfig: ResolvedConnectConfig = (
+    await (client as any).pool._getNormalizedConnectConfig()
+  ).connectionParams;
```
scotttrinh, line 44: I think we have a new API for getting these on the Client object. Like `getConnectionParams` or something?

packages/ai/src/types.ts
```diff
@@ -25,53 +27,109 @@ export interface RAGRequest {
...
+export type StreamingMessage = z.infer<typeof _edgedbRagChunkSchema>;
 
-export type StreamingMessage =
-  | MessageStart
-  | ContentBlockStart
-  | ContentBlockDelta
-  | ContentBlockStop
-  | MessageDelta
-  | MessageStop;
+const _edgedbRagChunkSchema = z.discriminatedUnion("type", [
```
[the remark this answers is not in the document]
diksipav: I used this in the vercel provider so it was easier to copy paste it.

packages/ai/src/core.ts
diksipav: tnx, I updated this

packages/vercel-ai-provider/src/edgedb-provider.ts
diksipav: I updated it.

diksipav:
There is a new Vercel AI SDK version out - 4. :D Let's complete this and see how we proceed.

packages/ai/src/core.ts
```diff
@@ -68,37 +69,68 @@ export class EdgeDBAI {
...
     const headers = request.stream
       ? { Accept: "text/event-stream", "Content-Type": "application/json" }
       : { Accept: "application/json", "Content-Type": "application/json" };
 
+    const { messages: initialMessages, prompt } = request;
+
+    if (prompt && initialMessages)
+      throw new Error(
+        "You can provide either a prompt or a messages array, not both.",
+      );
```
scotttrinh, line 82: Can we express this in the types instead of this runtime check?

packages/ai/src/core.ts
```diff
@@ -68,37 +69,68 @@ export class EdgeDBAI {
...
+          // to not include default prompt msgs, but if user provides messages
+          // and doesn't provide prompt.custom, since we add messages to the
+          // prompt.custom we also have to include default prompt messages
+          ...(!this.options.prompt?.custom &&
+            !providedPrompt && {
+              name: "builtin::rag-default",
+            }),
+          custom: [...(this.options.prompt?.custom || []), ...messages!],
+        },
+        query: [...messages!].reverse().find((msg) => msg.role === "user")!
```
scotttrinh, line 113: Can we make `messages` an easier type to deal with earlier? Do we lose anything to add `?? []` at the end of the initial assignment to get `EdgeDBMessage[]`? Maybe there is a semantic difference between `undefined` and `[]` that we want to preserve?

packages/ai/src/core.ts
```diff
@@ -68,37 +69,68 @@ export class EdgeDBAI {
...
+          .content[0].text,
       }),
     });
 
     if (!response.ok) {
       const bodyText = await response.text();
-      throw new Error(bodyText);
+      throw new Error(JSON.parse(bodyText).message);
```
scotttrinh, line 120: I'd be careful about parsing the arbitrary text here. It'll be JSON when we return an error, but if we hit some weird case (like CDN returning a 503 or something) that returns HTML or plain text, this will throw a totally unhelpful error.

If you want to make it a nicer message when it's JSON, I'd say do something like: check the content-type of the response, and if it's JSON-like, use `await response.json()`, otherwise use just dump the full text into the error.

packages/ai/src/types.ts
```diff
@@ -17,61 +19,118 @@ export interface QueryContext {
...
-  type: "message_start";
-  message: {
-    role: "assistant" | "system" | "user";
-    id: string;
-    model: string;
-  };
-}
+export type RagRequest =
+  | (RagRequestBase & { messages: EdgeDBMessage[]; prompt?: undefined })
+  | (RagRequestBase & { prompt: string; messages?: undefined });
```
scotttrinh, line 29: I think we want:

```suggestion
  | (RagRequestBase & { messages: EdgeDBMessage[] })
  | (RagRequestBase & { prompt: string });
```

Right?

packages/ai/src/types.ts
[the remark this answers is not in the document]
scotttrinh: Could we export it then maybe? Not a blocker.

scotttrinh (commented):
A few more comments. Still not finished reviewing everything, but wanted to leave these thoughts.

packages/vercel-ai-provider/src/edgedb-chat-language-model.ts
```diff
@@ -0,0 +1,574 @@
...
+
+  async doGenerate(
+    options: Parameters<LanguageModelV1["doGenerate"]>[0],
+  ): Promise<Awaited<ReturnType<LanguageModelV1["doGenerate"]>>> {
+    const { args, warnings } = this.getArgs(options);
+    const { messages } = args;
+
+    const providedPromptId =
+      this.settings.prompt &&
+      ("name" in this.settings.prompt || "id" in this.settings.prompt);
```
scotttrinh, line 196: We seem to repeat this logic (and the `prompt` building logic below) a bunch. Can we export a utility that encapsulates this?

packages/vercel-ai-provider/src/edgedb-prepare-tools.ts
```diff
@@ -0,0 +1,152 @@
...
+) {
+  const isOpenAI = isOpenAIModel(model);
+  const isAnthropic = isAnthropicModel(model);
+
+  // when the tools array is empty, change it to undefined to prevent errors:
+  const tools = mode.tools?.length ? mode.tools : undefined;
+  const toolWarnings: LanguageModelV1CallWarning[] = [];
+
+  if (tools == null) {
+    return { tools: undefined, tool_choice: undefined, toolWarnings };
```
scotttrinh, line 41: Nit: do we need to set this explicitly to `undefined` or can we just return:

```suggestion
    return { toolWarnings };
```

packages/vercel-ai-provider/src/edgedb-prepare-tools.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  const toolChoice = mode.toolChoice;
+  const edgedbTools = isAnthropic
+    ? edgedbAnthropicTools
+    : edgedbOpenAILikeTools;
+
+  if (toolChoice == null) {
+    return {
+      tools: edgedbTools,
+      tool_choice: undefined,
```
scotttrinh, line 87: Same question here about an explicit `undefined` value.

packages/vercel-ai-provider/src/edgedb-provider.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export async function createEdgeDB(client: Client): Promise<EdgeDBProvider> {
+  const connectConfig = await client.resolveConnectionParams();
+
+  const fetch = await getAuthenticatedFetch(
+    connectConfig,
+    httpSCRAMAuth,
+    "ext/ai/",
+  );
```
scotttrinh, line 47: Mostly leaving this as a note for the future:

We can avoid having to construct an authenticated fetch in every package by having the `Client` instance return one. The client has already had to construct an authenticated fetch client, and in this particular case, if you are in a Node environment, the one the driver constructs uses the native node crypto instead of browser crypto like this.

Maybe an API like:

```ts
const fetch = await client.getAuthenticatedFetch({
  path: "ext/ai/",
});
```

scotttrinh (approved):
A few more comments. In general, I think we can merge this as-is and follow up on some of the ideas and comments I laid out to keep this PR from getting to stale.

packages/ai/src/types.ts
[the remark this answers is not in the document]
diksipav: we can export it, I just dont understand why would someone need it? 
we export:
`export type StreamingMessage = z.infer<typeof _edgedbRagChunkSchema>;`
which will user most probably use?

packages/ai/src/types.ts
diksipav: yes yes!

packages/vercel-ai-provider/src/edgedb-prepare-tools.ts
diksipav: I didnt really try, all other providers return with undefined

packages/ai/src/core.ts
diksipav: I believe u were thinking sth like this:
```
    if (!response.ok) {
      const contentType = response.headers.get("content-type");
      let errorMessage: string;

      if (contentType && contentType.includes("application/json")) {
        const json = await response.json();
        errorMessage = json.message || "An error occurred";
      } else {
        const bodyText = await response.text();
        errorMessage = bodyText || "An unknown error occurred";
      }
      throw new Error(errorMessage);
    }
```

packages/vercel-ai-provider/src/edgedb-chat-language-model.ts
diksipav: we repeat this twice in the provider repo, in doGenerate and doStream, I can make a utility

packages/ai/src/types.ts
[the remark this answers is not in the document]
scotttrinh: > I just dont understand/see when would someone need it?

Oh, I thought you mean that you are using a copy-pasted version of this schema in the Vercel provider, no? I'm saying export it here and import it in the Vercel provider instead of copy-pasting it. Or export it from the Vercel provider and import the type here. Whichever makes sense.

This is not a blocker for merging, it just feels a little strange here to have an essentially "private" Zod schema that isn't used for anything, but has to be kept in sync across two packages.

packages/ai/src/core.ts
scotttrinh: Yeah, something like that! Although, you might want to do something a bit more strict for JSON since JSON could be all kinds of shapes (including a `null` literal), like:

```ts
        if (typeof json === "object" && json != null && "message" in json) {
          errorMessage = json.message;
        } else {
          errorMessage = `An error occurred: ${json}`;
        }
```

Definitely a lot more work than returning the unparsed text, and probably can/should move into some general purpose error handling util function. I just want to avoid a runtime error that will make it hard to tell what happened.

packages/ai/src/types.ts
scotttrinh: Yeah, you'll have to do:

```ts
if ("messages" in request) {
```

etc.

I mostly think it's a code smell to explicitly assign something as `undefined`. Maybe a compromise that keeps the ergonomics here is to use `null` and make it non-optional so there is no ambiguity? You'll probably have to do some work to conform to that type though, right?

One other more principled solution is to make this a more explicit discriminant by adding a `type` or `tag` or `kind` property to make it easier to detect which logical branch we're in without having to do a bunch of defensive checks for undefined/null.

packages/vercel-ai-provider/src/edgedb-chat-language-model.ts
scotttrinh: > providedPromptId is actually a very simple check, maybe we dont need a utility

Yeah, the "rule of three" says we should wait for one more case, right? 😅 My perspective on this "it's simple, just copy-paste" is fine if the logic is only a line or two, but the full logic of what this boolean represents and how it's used later is actually kind of complicated.

Mostly I care about a reviewer or someone coming along wanting to understand or change this code and the big spreading logic we do to construct the `prompt` object, they have to understand how this boolean is related to the later spread. A nicely named function can make this easier to understand without needing to dive too deeply into what it's doing unless that logic needs to change. And _if_ that logic needs to change, making it a shared function means it won't accidentally diverge in the two places.

packages/vercel-ai-provider/src/edgedb-prepare-tools.ts
scotttrinh: We shouldn't block on this, it's not _wrong_, but as I mentioned earlier, I personally find explicit use of `undefined` to be some kind of smell.

packages/ai/src/types.ts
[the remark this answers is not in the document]
diksipav: hm okay I can do this, even tho it makes sense and is better to have one source of truth it feels a bit weird to me to install the whole lib just because one type, but maybe that's usually the way to go in this kind of situations?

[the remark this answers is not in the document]
scotttrinh: Oh, I hadn't noticed: Does the Vercel AI SDK Provider not use the `@edgedb/ai` package for anything already?

packages/vercel-ai-provider/src/edgedb-chat-language-model.ts
diksipav: I created buildPrompt and buildQuery, I agree it is better. Thank you! Previously I tried to encapsulate in the separate function the whole body or even the whole object we pass to the `postJsonToApi` which then complicates types handling a bit more (especially for args and zod schemas), but is ofc also possible.

packages/ai/src/types.ts
diksipav: fixed this, I actually also needed to use the function:
```
export function isPromptRequest(
  request: RagRequest,
): request is RagRequestPrompt {
  return "prompt" in request;
}
```
To explicitly say that the type is RagRequestPrompt if `"prompt" in request`

Using just the `in` check is not enough, I guess it is still not able to differentiate the type .

I don't really like using the additional property like `type` in this case since the `RagRequest` comes from the client, I can tho probably update it in the `queryRag/streamRag` and send the type too to the `fetchRag`.

scotttrinh: Yeah, using a custom type guard here makes sense to me, good thinking.

PR #1133, opened 2024-11-26
Add support for query tagging

jaclarke:
Depends on https://github.com/edgedb/edgedb/pull/8036

### DIFF
--- a/packages/driver/src/options.ts
+++ b/packages/driver/src/options.ts
@@ -1,4 +1,5 @@
 import * as errors from "./errors";
+import { utf8Encoder } from "./primitives/buffer";
 
 export type BackoffFunction = (n: number) => number;
 
@@ -118,6 +119,8 @@ export class TransactionOptions {
   }
 }
 
+const TAG_ANNOTATION_KEY = "tag";
+
 export interface SessionOptions {
   module?: string;
   moduleAliases?: Record<string, string>;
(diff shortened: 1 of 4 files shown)


packages/driver/src/options.ts
```diff
@@ -150,33 +160,56 @@ export class Session {
...
 
   withGlobals(globals: { [name: string]: any }): Session {
-    return new Session({
-      ...this,
+    return this._clone({
       globals: { ...this.globals, ...globals },
     });
   }
 
+  withTag(tag: string | null): Session {
```
jaclarke, line 193: I'm not sure exactly what api we want here? It looks like the rust client has a `withTag` api (also with reserved tag names) (https://github.com/edgedb/edgedb-rust/pull/364/files#diff-608add6cc9859de62fac8b605b75549b985354b28530c7c1b07180ccd86fd729R574), but the python implementation just exposes the underlying annotations? (https://github.com/edgedb/edgedb-python/pull/552/files#diff-ac82d0a568b6ac577eb791f28078350639f6ffd9b904254a17557bb7dff1fb62R416)

fantix: Ah maybe I should just stay with `with_tag()` in Python too (so that the same tag limitation can be applied). cc @1st1 ?

scotttrinh (approved):

packages/driver/src/options.ts
fantix: `withQueryTag()` it is then.

fantix (approved):

PR #1136, opened 2024-11-28
Add ability to specify OAuth endpoints

scotttrinh:
If your database is deployed to a private network and is not available at a public URL, this adds the ability to drive the whole flow from your application server, delegating to the auth server from your application server.

---

When testing this manually, I set up a new route handler at `/auth/oauth` that shadows the one built into our `createAuthRouteHandlers` behavior like this:

```ts
export async function GET(request: NextRequest): Promise<NextResponse> {
  const callbackUrl = new URL("auth/server/oauth/callback", authConfig.baseUrl);
  const authorizeUrl = new URL(
    "auth/server/oauth/authorize",
    authConfig.baseUrl,
  );
  request.nextUrl.searchParams.set("callback_url", callbackUrl.toString());
  request.nextUrl.searchParams.set("authorize_url", authorizeUrl.toString());

  return await auth.oAuth.handleOAuth(request);
}
```

Which set the authorize and callback URLs explicitly to routes in my application, and then those route handlers simply called through to these new methods like:

```ts
//  app/auth/server/oauth/authorize.ts
export async function GET(request: NextRequest): Promise<NextResponse> {
  return await auth.oAuth.handleAuthorize(request);
}

// app/auth/server/oauth/callback.ts
export async function GET(request: NextRequest): Promise<NextResponse> {
  return await auth.oAuth.handleCallback(request);
}
```

Of course, instead of just calling this methods, one could do any kind of check here on the request if your intention is to add some behavior here (analytics? logic? check email against an invite list?)

### DIFF
--- a/packages/auth-nextjs/src/shared.ts
+++ b/packages/auth-nextjs/src/shared.ts
@@ -1,7 +1,6 @@
 import { type Client } from "edgedb";
 import {
   Auth,
-  builtinOAuthProviderNames,
   type BuiltinOAuthProviderNames,
   type TokenData,
   ConfigurationError,
@@ -20,7 +19,7 @@ import {
 
 import { cookies } from "next/headers";
 import { redirect } from "next/navigation";
-import type { NextRequest, NextResponse } from "next/server";
+import { type NextRequest, NextResponse } from "next/server";
 
 export { type BuiltinProviderNames, NextAuthHelpers, type NextAuthOptions };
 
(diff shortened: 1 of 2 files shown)


packages/auth-nextjs/src/shared.ts
```diff
@@ -138,6 +137,107 @@ export abstract class NextAuth extends NextAuthHelpers {
...
+      const pkceSession = await (await this.core).createPKCESession();
+      this.setVerifierCookie(pkceSession.verifier);
+
+      const location = pkceSession.addOAuthParamsToUrl(authorizeUrl, {
+        providerName,
+        redirectTo: redirectTo.toString(),
+        redirectToOnSignup: redirectToOnSignup.toString(),
+        ...(callbackUrl ? { callbackUrl } : {}),
+      });
+      console.log(`Redirecting to ${location}`);
```
jaclarke, line 184:
```suggestion
```

packages/auth-nextjs/src/shared.ts
```diff
@@ -138,6 +137,107 @@ export abstract class NextAuth extends NextAuthHelpers {
...
+
+      const authBasePath = this._authRoute.endsWith("/")
+        ? this._authRoute
+        : `${this._authRoute}/`;
+      const redirectTo = new URL("oauth/callback", authBasePath);
+      const redirectToOnSignup = new URL(redirectTo);
+      redirectToOnSignup.searchParams.set("isSignUp", "true");
+
+      const authorizeUrl =
+        req.nextUrl.searchParams.get("authorize_url") ??
```
jaclarke, line 172: Should this (and `callback_url`) be args to this `handleOAuth` function? I'm not sure they are meant to be something the user should be able to control? and in the example usage thay're appended to the req right before it's passed to this function anyway.

scotttrinh: The API I'm going for here is that these "handler" functions should always take a `NextRequest` and return a `Promise<NextResponse>` in a middleware like fashion. An alternative way to do this is to be more explicit about what parameters we explicitly depend on within the function itself, but I'm worried that it adds boilerplate of getting the data we need from the request. I can sketch that out though and see how that feels in actual use though.

jaclarke: Ok, maybe strip out those two values from the search params here: https://github.com/edgedb/edgedb-js/pull/1136/files#diff-660934b707fc817f8bd5cef3341ffb7f24ffc16f9e200843c762e384e2820c41L173 then? I haven't fully thought out the whole flow and there probably isn't an attack you could do using these params, but just be safe maybe 😄.

scotttrinh: > Ok, maybe strip out those two values from the search params here: [#1136 (files)](https://github.com/edgedb/edgedb-js/pull/1136/files#diff-660934b707fc817f8bd5cef3341ffb7f24ffc16f9e200843c762e384e2820c41L173) then?

Maybe I'm not understanding the suggestion but these parameters _will_ be present in the URL that the browser uses as a link, just like `provider`, and it needs to be passed all the way to the auth extension. It's not something an _end user_ needs to worry about, but the application developer will absolutely need to set this either in the link button's URL, or manually in their own backend like I did in my example.

> I haven't fully thought out the whole flow and there probably isn't an attack you could do using these params, but just to be safe

The URLs are checked against the `allowed_redirect_urls` config, so if there is an attack vector there, the attack also exists in the existing `redirect_to` search parameters.

jaclarke (approved):

PR #1137, opened 2024-11-29
[auth-nextjs] Async Request APIs, compat with Next.js 15

MartinCura:
This PR adapts the `@edgedb/auth-nextjs` package for the [kinda*] breaking change of Next.js 15, where [Request APIs like `cookies` and `params` are now async](https://nextjs.org/blog/next-15#async-request-apis-breaking-change).

To solve this, i simply added `await` to all corresponding places. Awaiting something sync is not strictly a noop, but for these cases it shouldn't change behavior afaict.

**Tested** in my own project with both Next.js before their breaking change and after it.

**Breaking change:** As `NextAppAuth.getSession()` has to await `cookies()`, it is now an `async` function, which means those upgrading have to add `await` when calling it.

```diff
const auth = createAuth(client, {
  baseUrl: `http://localhost:${process.env.PORT || 3000}`,
  authCookieName: "edgedb-session",
  pkceVerifierCookieName: "edgedb-pkce-verifier",
})

export const getCurrentUser = async () => {
- const session = auth.getSession()
+ const session = await auth.getSession()
  const userQuery = e.select(e.User, () => ({
    ...e.User["*"],
    filter_single: { id: e.global.current_user.id },
  }))
  return await userQuery.run(session.client)
}
```

This necessitates that `getSession` is now async, which is a breaking change, because it accesses the cookies.

*_kinda_ because as a compatibility layer it is still possible to not await it, but it's not what you want to do.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "ci:test": "turbo run ci:test"
   },
   "overrides": {
-    "@types/react": "npm:types-react@rc",
-    "@types/react-dom": "npm:types-react-dom@rc"
+    "@types/react": "npm:types-react@rc"
   }
 }
--- a/package.json
+++ b/package.json
@@ -24,5 +24,9 @@
     "format:fix": "prettier --write 'packages/*/(src|test)/**/*.(mts|ts)' 'integration-tests/*/*.(mts|ts)'",
     "ci:typecheck": "turbo run ci:typecheck",
     "ci:test": "turbo run ci:test"
+  },
+  "overrides": {
+    "@types/react": "npm:types-react@rc",
+    "@types/react-dom": "npm:types-react-dom@rc"
   }
 }
--- a/packages/auth-nextjs/src/app/index.ts
+++ b/packages/auth-nextjs/src/app/index.ts
@@ -203,8 +203,6 @@ export class NextAppAuth extends NextAuth {
   }
 }
 
-console.log("aaaaa");
-
 export default function createNextAppAuth(
   client: Client,
   options: NextAuthOptions,
--- a/packages/auth-nextjs/src/app/index.ts
+++ b/packages/auth-nextjs/src/app/index.ts
@@ -28,17 +28,19 @@ export {
 
 export class NextAppAuth extends NextAuth {
   getSession = cache(
-    () =>
+    async () =>
       new NextAuthSession(
         this.client,
-        cookies().get(this.options.authCookieName)?.value.split(";")[0] ?? null,
+        (await cookies())
+          .get(this.options.authCookieName)
+          ?.value.split(";")[0] ?? null,
       ),
   );
 
   createServerActions() {
     return {
       signout: async () => {
-        cookies().delete(this.options.authCookieName);
+        (await cookies()).delete(this.options.authCookieName);
       },
       emailPasswordSignIn: async (
         data: FormData | { email: string; password: string },
@@ -51,7 +53,7 @@ export class NextAppAuth extends NextAuth {
         const tokenData = await (
           await this.core
         ).signinWithEmailPassword(email, password);
-        this.setAuthCookie(tokenData.auth_token);
+        await this.setAuthCookie(tokenData.auth_token);
         return tokenData;
       },
       emailPasswordSignUp: async (
@@ -69,9 +71,9 @@ export class NextAppAuth extends NextAuth {
           password,
           `${this._authRoute}/emailpassword/verify`,
         );
-        this.setVerifierCookie(result.verifier);
+        await this.setVerifierCookie(result.verifier);
         if (result.status === "complete") {
-          this.setAuthCookie(result.tokenData.auth_token);
+          await this.setAuthCookie(result.tokenData.auth_token);
           return result.tokenData;
         }
         return null;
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -24,5 +24,8 @@
     "format:fix": "prettier --write 'packages/*/(src|test)/**/*.(mts|ts)' 'integration-tests/*/*.(mts|ts)'",
     "ci:typecheck": "turbo run ci:typecheck",
     "ci:test": "turbo run ci:test"
+  },
+  "overrides": {
+    "@types/react": "npm:types-react@rc"
   }
```
MartinCura, line 30: this is recommended for the react types to be found correctly, at least while React 19 is not marked stable. for pnpm, this must be in the root of the workspace; i'm not actually sure if the same is true for yarn.

packages/auth-nextjs/src/app/index.ts
```diff
@@ -27,18 +27,19 @@ export {
...
 
 export class NextAppAuth extends NextAuth {
-  getSession = cache(
-    () =>
-      new NextAuthSession(
-        this.client,
-        cookies().get(this.options.authCookieName)?.value.split(";")[0] ?? null,
-      ),
-  );
+  getSession = cache(async () => {
```
MartinCura, line 30: the necessary breaking change, making this function now async

packages/auth-nextjs/src/shared.ts
```diff
@@ -93,6 +93,10 @@ export interface CreateAuthRouteHandlers {
   onSignout(req: NextRequest): Promise<never>;
 }
 
+type ReadonlyRequestCookies = Awaited<ReturnType<typeof nextCookies>>;
+const cookies: () => Promise<ReadonlyRequestCookies> | ReadonlyRequestCookies =
+  nextCookies;
```
MartinCura, line 98: unnecessary, so can remove it, but just wanted to show that the return type depends on the next.js version used by the user

1st1:
Thank you, we'll take a look soon!

package.json
MartinCura, line 1: do you prefer i bump the minor version in this PR or do you prefer to do this in another one?

scotttrinh: We'll worry about the version numbers in the release PR, so no need to bump it here.

packages/auth-nextjs/src/shared.ts
```diff
@@ -153,8 +159,9 @@ export abstract class NextAuth extends NextAuthHelpers {
     return {
       GET: async (
         req: NextRequest,
-        { params }: { params: { auth: string[] } },
+        ctx: { params: Promise<{ auth: string[] }> },
```
scotttrinh, line 162: Will this work in <15 or will those users need to do a type-cast?

packages/auth-nextjs/src/shared.ts
```diff
@@ -232,8 +239,9 @@ export abstract class NextAuth extends NextAuthHelpers {
                 req,
               );
             }
-            this.setAuthCookie(tokenData.auth_token);
-            cookies().delete(this.options.pkceVerifierCookieName);
+            await this.setAuthCookie(tokenData.auth_token);
+            const cookieStore = await cookies();
+            cookieStore.delete(this.options.pkceVerifierCookieName);
```
scotttrinh, line 244: Not a blocker: Probably a good enough reason to promote this (and the auth cookie delete) to it's own function. Seemed superfluous when it was a one liner before.

scotttrinh (commented):
Thanks so much for taking on this mechanical work! It's been on our "next up" list for a few weeks now.

packages/auth-nextjs/src/shared.ts
scotttrinh: Feel free to remove it in this PR, thanks for the demonstration!

packages/auth-nextjs/src/shared.ts
MartinCura: sgtm, i'll name it `deletePkceVerifierCookie()`

packages/auth-nextjs/src/shared.ts
scotttrinh: Yeah, I'm not worried too much about runtime, and even if it breaks backwards compatibility, I'm fine with shipping it, I just want to understand everything that might break here so we can update the docs.

packages/auth-nextjs/src/shared.ts
MartinCura: done

packages/auth-nextjs/src/shared.ts
MartinCura: Well the `GET` is exported in a `route.ts` to act as a Next.js Route Handler but it's not type checked against anything. I can run and build the project running Next.js before their breaking change with this version, no errors. And i don't know of any way that someone could be type-checking this and get a warning/error after this change.

At most we could add a small note to the docs referencing the change introduced by Next.js 15 and how this small library targets the new behaviour but supports both.

MartinCura:
> Thanks so much for taking on this mechanical work! It's been on our "next up" list for a few weeks now.

No worries! Glad to help and love what you guys are doing. I'm working on a sizable project by myself, so these kinds of helper libraries that help me ship faster are very welcome.

MartinCura:
👋🏼 Small update: React 19 stable finally dropped, and with it a stable release of `@types/react`, so the `overrides` i added in this PR should not be needed anymore.

PR #1139, opened 2024-12-03
Update provider API with provider settings

diksipav:


packages/vercel-ai-provider/src/edgedb-provider.ts
```diff
@@ -24,21 +22,45 @@ import type {
...
     modelId: EdgeDBEmbeddingModelId,
     settings?: EdgeDBEmbeddingSettings,
   ) => EmbeddingModelV1<string>;
 }
 
-export async function createEdgeDB(client: Client): Promise<EdgeDBProvider> {
+export interface EdgeDBProviderSettings {
+  /**
+Use a different URL prefix for API calls, e.g. to use proxy servers.
+The default prefix is `https://api.mistral.ai/v1`.
```
scotttrinh, line 44: I assume this is from a copy-paste?

```suggestion
```

packages/vercel-ai-provider/src/edgedb-chat-settings.ts
```diff
@@ -70,6 +70,8 @@ export interface QueryContext {
 export interface EdgeDBChatConfig {
   provider: string;
   fetch: FetchFunction;
+  baseURL: string;
```
scotttrinh, line 73: If we're going to have this be either an explicit URL or something falsy, I'd rather it be:

```suggestion
  baseURL: string | null;
```

Instead of relying on `Boolean("") === false` behavior.

packages/vercel-ai-provider/src/edgedb-provider.ts
```diff
@@ -24,21 +22,45 @@ import type {
...
+     */
+  headers?: Record<string, string>;
+}
+
+export async function createEdgeDB(
+  client: Client,
+  options: EdgeDBProviderSettings = {},
+): Promise<EdgeDBProvider> {
   const connectConfig = await client.resolveConnectionParams();
+  const baseURL = withoutTrailingSlash(options.baseURL) ?? "";
```
scotttrinh, line 59: And therefore:
```suggestion
  const baseURL = withoutTrailingSlash(options.baseURL) ?? null;
```

scotttrinh (changes requested):

scotttrinh (approved):

PR #1145, opened 2024-12-10
Missing PKCE is success in email verification flow

scotttrinh:
Since end users might verify their email on a different device than the user agent they initiated the sign up (or sign in) flow with, treat this as a success condition. The application will need to detect this case and show a message that confirms that the email is verified, but that the user will need to sign in to complete.

Note: this is a breaking change, so we need to bump the minor (given this is a `0.x` release)


jaclarke (approved):

scotttrinh:
@jaclarke Sorry about missing this: the built-in UI needed this update, too, so: [`d6c1053` (#1145)](https://github.com/edgedb/edgedb-js/pull/1145/commits/d6c1053766fb90d9a1d4187716bb00215dba7d32)

jaclarke (approved):

packages/auth-nextjs/src/shared.ts
```diff
@@ -560,9 +564,16 @@ export abstract class NextAuth extends NextAuthHelpers {
...
+              // sign-up. This is fine, but the application will need to detect
+              // this and inform the end user that they will need to initiate a
+              // new sign up attempt to complete the flow.
               return onBuiltinUICallback(
                 {
-                  error: new PKCEError("no pkce verifier cookie found"),
+                  error: null,
+                  tokenData: null,
+                  provider: null,
+                  isSignUp: false,
```
jaclarke, line 576: Actually, this probably should be `true`, as email verification is part of the sign up flow.

scotttrinh: Hmm, actually, I don't think this works as-is: it's impossible to tell the difference between a sign in attempt from an unverified email vs. a successful verification without the PKCE verifier since they have the same shape.

Let me do a little design work here. I think we might need to introduce a discriminant into this callback payload type to make it less ambiguous for the consumer.

scotttrinh:
Going to put this back in draft, it's not nearly ready:

- The core verify methods do _not_ verify emails without a pkce verifier. They should be updated.
- We don't really have a way to signal that a `sign-up` -> `verify` --missing pkce--> `sign in` is really a sign up with more steps. I plan on adding the `identity_id` to the sign up response from the server so you can at least create your user before email verification. This will impact this interface.

PR #1146, opened 2024-12-10
Add SQL query method that returns array of tuples

scotttrinh:
For now it does the most naive thing and uses `Object.values`, but we are working on first-class support for this from the server and will switch to using that updated style when that lands.

### DIFF
--- a/packages/driver/src/baseClient.ts
+++ b/packages/driver/src/baseClient.ts
@@ -250,7 +250,7 @@ export class ClientConnectionHolder {
     );
   }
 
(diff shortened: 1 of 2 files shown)


1st1 (changes requested):
-1 to add this as we'll just have a proper data type soon. But I'm OK to temporarily have a `_querySQLTuple` method if this is urgent.

scotttrinh:
> -1 to add this as we'll just have a proper data type soon.

Does that mean the type of `querySQL` is changing to be this new datatype or will there still be two APIs (or some kind of option on the method itself?), one for getting an array of objects, and one for getting an array of tuples?

PR #1151, opened 2024-12-11
auth-remix: rename edgedb to gel

diksipav:


packages/auth-remix/package.json
```diff
@@ -1,12 +1,12 @@
...
+  "name": "@gel/auth-remix",
+  "description": "Helper library to integrate the Gel Auth extension with Remix",
   "version": "0.2.2",
-  "author": "EdgeDB <info@edgedb.com>",
+  "author": "Gel <info@gel.com>",
   "type": "module",
   "repository": {
     "type": "git",
-    "url": "https://github.com/edgedb/edgedb-js.git",
+    "url": "https://github.com/gel/gel-js.git",
```
diksipav, line 9: may need an update

packages/auth-remix/src/client.ts
```diff
@@ -27,8 +27,8 @@ export class RemixClientAuth {
   /** @internal */
   constructor(options: RemixAuthOptions) {
     this.options = {
-      authCookieName: "edgedb-session",
-      pkceVerifierCookieName: "edgedb-pkce-verifier",
+      authCookieName: "gel-session",
+      pkceVerifierCookieName: "gel-pkce-verifier",
```
scotttrinh, line 31: Hmm.. I wonder if we can support a fallback here so that apps that update to use the latest packages don't accidentally log all of their users out.

Let's think of an approach here that will work easily for new users and upgrading users.

diksipav: oh yes, this is also a thing with all other auth libs

diksipav: hey, I pushed a commit (I can maybe make some code more neat), but the functionality is basically:
in constructor we define gel (can be overriden by user names),
when I parse I parse both (firstly gel and if not exists check for edgedb one),
when I delete cookies I delete both gel and edgedb,
when I create new cookies I create only gel...

packages/auth-remix/src/server.ts
```diff
@@ -1169,6 +1188,27 @@ function _extractParams(
...
+function parseCookies(
+  req: Request,
+  options: {
+    baseUrl: string;
+    authRoutesPath: string;
+    authCookieName: string;
+    pkceVerifierCookieName: string;
+    passwordResetPath?: string;
+    magicLinkFailurePath?: string;
+  },
```
scotttrinh, line 1200: You don't need to define all of the options here, just the ones you use.

```suggestion
  options: {
    authCookieName: string;
    pkceVerifierCookieName: string;
  },
```

---

Another option is to make this a method on the class so you can use `this.options` instead of passing it in, but maybe this would need to be defined on multiple classes?

packages/auth-remix/src/server.ts
```diff
@@ -313,6 +318,13 @@ export class RemixServerAuth extends RemixClientAuth {
                 path: "/",
               }),
             );
+            headers.append(
```
scotttrinh, line 321: Why are we deleting this verifier cookie here and in the other two places?

diksipav: u are right, this is not needed... we delete the pkce when/after verifying

diksipav: hmm let me check what happens if user started signing up flow but didnt finish, while in the meantime the lib version changed

scotttrinh (approved):

PR #1154, opened 2024-12-11
Fix some bugs in toEdgeQL when using `e.with`

jaclarke:
Reported on discord: https://discord.com/channels/841451783728529451/1313928437886619738

Fix 1:
While debugging realised the original query is incorrect: the expressions being added to the `with` block can't reference the select expression the with block is attached to, as that select can't be hoisted into a var in it's own with block. The expression tree walker was not checking for this case (an expr's parent also being one of its child exprs), which could cause an infinite loop in `toEgdeQL`.

Fix 2:
We check in `toEdgeQL` that all the places where a repeated expr is used exist within the scope of the with block it's being hoisted to. Here we were incorrectly not considering expressions already explicitly bound to that with block to be valid scopes.

### DIFF
--- a/integration-tests/lts/update.test.ts
+++ b/integration-tests/lts/update.test.ts
@@ -101,7 +101,7 @@ describe("update", () => {
         },
       }));
 
-      return e.with([updateChar, updateProfile], movie);
+      return e.with([updateChar, updateProfile], e.select(movie));
     }).toEdgeQL();
   });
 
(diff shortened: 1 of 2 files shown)


jaclarke:
> 👏
> 
> I'd like to also produce better error output to help troubleshoot these, but let's do that as a separate PR.

Opened an issue for this: https://github.com/edgedb/edgedb-js/issues/1155

PR #1156, opened 2024-12-14
Implement proper array-like SQL record/row compatible with 6b1

1st1:


packages/driver/src/codecs/registry.ts
```diff
@@ -261,6 +263,15 @@ export class CodecsRegistry {
...
 
+        case CTYPE_RECORD: {
+          const els = frb.readUInt16();
+          for (let i = 0; i < els; i++) {
+            const elm_length = frb.readUInt32();
+            frb.discard(elm_length + 2);
+          }
+          break;
+        }
+
```
jaclarke, line 274: I don't think this code will ever actually get executed, since for protocol v2+ the type descriptors are length prefixed, so we skip the whole typedesc in one go (line 113).

packages/driver/src/codecs/record.ts
```diff
@@ -0,0 +1,84 @@
...
+      let val = null;
+      if (elemLen !== -1) {
+        buf.sliceInto(elemBuf, elemLen);
+        val = subCodecs[i].decode(elemBuf);
+        elemBuf.finish();
+      }
+      result[i] = val;
+    }
+
+    return result;
```
jaclarke, line 70: I thought we were going to decode sql records into some custom datatype where you could access columns by name or index?

scotttrinh: I'm pesonally fine with this being an array of tuples by default. If we had a custom datatype for this here are some options I've just thought of:

- `asTuples` and `asObjects` methods to make it easy to serialize, but then we pay for the overhead of the custom datatype.
- `toJSON` method that serializes as an array of tuples
- wrap an array of tuples (and has the appropriate getter in the `valueOf` prototype method) and provide a `withNames` method that will wrap it in an object interface with the names

packages/driver/src/codecs/registry.ts
1st1: Hm, good point, I'll double check

packages/driver/src/codecs/record.ts
1st1: I thought so too and then I started digging why we stopped doing custom data types for named tuples and found this: https://github.com/edgedb/edgedb-js/pull/310

tl;dr: we can do custom types but some JS frameworks hardcode their JSON serialization exclusively for Array / Object types and fail with anything that extends them or does anything funky

scotttrinh: Yeah, I support just making this an array of tuples and worry about exposing a way of getting an array of objects as a separate effort.

jaclarke: Yeah, I had forgot about the serialisation issues. Maybe this can be solved with the custom codecs api? We could default to decoding to arrays as it is now, but also provide a builtin 'custom' codec that would decode to objects with names, so users could choose?

1st1: Yeah, I thought about this too. It might be a slightly different special cased version of `.withCodecs`, because "SQL Record" doesn't really have a type name. Can be symbol though...

PR #1161, opened 2024-12-18
Introduce client.withCodecs()

1st1:
**This is a draft**

ToDo:

- [x] Unify Session / Options objects
- [x] Implement withCodecs/decode
- [x] Implement withCodecs/encode
- [x] Add strict typing for withCode(spec)
- [ ] Add docs
- [x] Cleanup more legacy protocol code
- [x] Kill `setCustomCodecs`
- [ ] Enum is a bit weird -- add a test for enum types


packages/driver/src/codecs/codecs.ts
```diff
@@ -49,8 +50,94 @@ import {
...
+
+  export type AnyCodec = Codec<any>;
+
+  export type BoolCodec = Codec<boolean>;
+  export type Int16Codec = Codec<number>;
+  export type Int32Codec = Codec<number>;
+  export type Int64Codec = Codec<bigint>;
+  export type Float32Codec = Codec<number>;
+  export type Float64Codec = Codec<number>;
+  export type BigIntCodec = Codec<string>;
```
jaclarke, line 74:
```suggestion
  export type BigIntCodec = Codec<bigint>;
```

packages/driver/src/codecs/json.ts
```diff
@@ -74,7 +84,11 @@ export class PgTextJSONCodec extends JSONCodec {
 export class JSONStringCodec extends ScalarCodec implements ICodec {
```
jaclarke, line 84: With the old customCodecs removed we can also delete this.

packages/driver/t.mjs
```diff
@@ -0,0 +1,11 @@
...
+
+let c = createClient({
+  host: 'localhost',
+  tlsSecurity: 'insecure',
+});
+
+await c.ensureConnected();
+
+console.log('here')
+console.log(await c.query('select (<default::foo>42, "aa")'))
```
jaclarke, line 11: Debugging?

packages/driver/src/baseConn.ts
```diff
@@ -1121,6 +716,10 @@ export class BaseRawConnection {
     let parsing = true;
     let warnings: errors.EdgeDBError[] = [];
 
+    if (ctx == null) {
+      ctx = new CodecContext(state.codecs);
+    }
```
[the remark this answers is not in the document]
1st1: If we cache it in Options, how would you handle schema changes?

packages/driver/src/codecs/codecs.ts
1st1: Believe it or not -- this is intentional, as it's an extra step for us to create a bigint so string is more fundamental. So I was torn. That said it's highly unlikely someone brings their own bigint implementation

packages/driver/src/codecs/json.ts
1st1: good

packages/driver/src/codecs/codecs.ts
jaclarke: Ah, ok. That makes sense.

packages/driver/src/baseConn.ts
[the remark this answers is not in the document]
jaclarke: The type ids would change if the schema changes right? And these codecs are just for scalars which have stable ids/names anyway.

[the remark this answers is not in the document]
jaclarke: Oh, unless you have user defined scalars and they change the scalar type that's extended?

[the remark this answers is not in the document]
jaclarke: Maybe we could use the type id as the key in the codec context mapping instead of the name?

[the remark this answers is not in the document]
1st1: I think I know the way... let me prototype it

packages/driver/src/codecs/codecs.ts
scotttrinh: Should we also do that for Int64Codec?

packages/driver/src/codecs/codecs.ts
```diff
@@ -49,8 +50,94 @@ import {
...
+  export type MemoryCodec = Codec<bigint>;
+  export type PgVectorCodec = Codec<Float32Array>;
+  export type PGVectorSparseCodec = Codec<
+    [dimensions: number, indexes: Uint32Array, values: Float32Array]
+  >;
+  export type StrCodec = Codec<string>;
+  export type UUIDCodec = Codec<Uint8Array>;
+
+  // TODO: Figure out if we can drop the dep
+  // on an external package for Float16Array.
```
scotttrinh, line 98: Unfortunately not: only Deno, Firefox, and Safari support native Float16Array

scotttrinh: Sure, you mean basically vendor the existing library? We can also just use `Float32Array` until there is better support for `Float16Array`.

packages/driver/src/codecs/codecs.ts
1st1: Int64Codec gives you bigint already as that's the most efficient way (we can read int64 to bigint in node fast).

I think I'll go and update the bigint codec to give you bigint. The point of this specific API isn't to replace native bigint with some stuff, but rather to wrap it in `Money` or something, for which us returning a string makes no sense.

So I'll incorporate James' suggestion.

packages/driver/src/baseConn.ts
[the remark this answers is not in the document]
1st1: OK, done. I think my approach will do it, I'll try to remember chatting about it on Monday

packages/driver/src/codecs/codecs.ts
1st1: I've looked at that implementation, it's huge and complicated :/ Vendoring is an option

PR #1162, opened 2024-12-19
Gel prisma

vpetrovykh:


packages/generate/src/gel-prisma.ts
```diff
@@ -0,0 +1,714 @@
+import type { CommandOptions } from "./commandutil";
+import type { Client } from "edgedb";
+import * as fs from "node:fs";
+
+const WARNINGS: string[] = [];
```
scotttrinh, line 5: Should we log these?

packages/generate/src/gel-prisma.ts
```diff
@@ -0,0 +1,714 @@
...
+  if (!CLEAN_NAME.test(name)) {
+    warn(`Non-alphanumeric names are not supported: ${name}`);
+    return false;
+  }
+  return true;
+}
+
+async function getSchemaJSON(client: Client) {
+  const types = <JSONType[]>JSON.parse(await client.queryJSON(INTRO_QUERY));
+  const modules = <string[]>JSON.parse(await client.queryJSON(MODULE_QUERY));
```
scotttrinh, line 152: Why `queryJSON` -> `JSON.parse` instead of `query`?

packages/generate/src/gel-prisma.ts
vpetrovykh: Maybe instead of collecting them, just print them into stderr as they happen?

packages/generate/src/gel-prisma.ts
vpetrovykh: This was mostly a translation from the Python version of this tool. There I was collecting things into dicts rather than native objects so that I could access things uniformly via `[ ]` and not worry about fields. I realize that in JS version this is not necessary, so a query is probably right in this spot.

packages/generate/src/gel-prisma.ts
scotttrinh: I think you can print them to stdout with `console.log`, tbh. I don't think our debug output in our generators are very principled about that sort of thing.

Issue #674, opened 2023-07-03
If you want to make EdgeDB's base64 encode function faster & use less memory


Jarred-Sumner:
Change this code:
https://github.com/edgedb/edgedb-js/blob/b2d9ef57a9efddcc3737ecd9709533650f09ac36/packages/driver/src/primitives/buffer.ts#L35-L44

To something more like:

```ts
if (typeof Buffer === "function") {
  decodeB64 = (b64: string): Uint8Array => {
    // @ts-ignore
    return Buffer.from(b64, "base64");
  };
  encodeB64 = (data: Uint8Array): string => {
    // @ts-ignore
    //  - Avoid cloning the contents of Buffer
    //  - Avoid creating a new Buffer if it is already a Buffer instance
    const buffer = !Buffer.isBuffer(data) ? Buffer.from(data.buffer, data.byteOffset, data.byteLength) : data;
    return buffer.toString("base64");
  };
} 
```

scotttrinh:
Thanks for the tip @Jarred-Sumner! I'll take a look. Love these easy wins.

PR #1163, opened 2025-01-06
Remove deno

diksipav:
Remove Deno specific code and Deno tests from driver, generate, integration-tests.
Don't publish anymore specific driver Deno build to Deno Land.
Update generate pkg to use "npm:edgedb" imports in generated code for Deno.


packages/driver/src/primitives/buffer.ts
```diff
@@ -18,41 +18,25 @@
...
+const decodeB64 = (b64: string): Uint8Array => {
+  return Buffer.from(b64, "base64") as unknown as Uint8Array;
+};
+
+const encodeB64 = (data: Uint8Array): string => {
+  const buf = !Buffer.isBuffer(data)
+    ? Buffer.from(data.buffer, data.byteOffset, data.byteLength)
+    : data;
+  return buf.toString("base64");
+};
```
1st1, line 37: Hm, I'm not sure about this change -- this is for supporting decodeB64 & encodeB64 in both browser and node, I think... @jaclarke thoughts?

packages/driver/src/adapter.ts
```diff
@@ -1,20 +1,23 @@
-import * as crypto from "crypto";
-import { promises as fs } from "fs";
-import * as net from "net";
-import * as os from "os";
-import * as path from "path";
-import * as tls from "tls";
+import * as crypto from "node:crypto";
```
1st1, line 1: This file probably shouldn't exist anymore, as there's nothing it "adapts" -- there's just one thing here as far as I can see.

packages/driver/src/adapter.crypto.ts
```diff
@@ -0,0 +1,33 @@
+import type { CryptoUtils } from "./utils";
```
1st1, line 1: Not sure it's needed to have "adapter." prefix in all files that facade one API or another depending on Node vs Browser

packages/driver/src/adapter.ts
```diff
@@ -1,20 +1,23 @@
...
 
 export { path, net, fs, tls, process };
 
 export async function readFileUtf8(...pathParts: string[]): Promise<string> {
   return await fs.readFile(path.join(...pathParts), { encoding: "utf8" });
 }
 
+// export function hasFSReadPermission(): boolean {
+//   return Deno.permissions.querySync({ name: "read" }).state === "granted";
+// }
```
1st1, line 20: This should be removed

packages/driver/src/primitives/buffer.ts
jaclarke: Yeah, ideally we shouldn't change this. Is there a reason for importing buffer now?

If I remember we replaced Buffer everywhere with Uint8Array, because node/deno/browser/edge runtime/etc. all support Uint8Array, but Buffer is just a node api (though some bundlers I think automatically inject a polyfill?). Then added this code to use Buffer if it exists as a perf optimisation: https://github.com/edgedb/edgedb-js/issues/674. So I think explicitly importing Buffer again may break some things?

diksipav: oh yes, node import will not work in browser, checking it now

packages/driver/src/adapter.ts
diksipav: We use functions from this file: walk, exists, readFileUtf8, input, hasFSReadPermission, hashSHA1toHex. I renamed it to `systemUtils.ts`. If you think `fsUtils` or some other name is better I can update. I removed exports from this file that can be directly imported from node at other places.

packages/driver/src/adapter.ts
diksipav: I left this comment to ask do we need this, and I believe we need it for Deno users, so I added it in the hasFSReadPermission func

packages/driver/src/primitives/buffer.ts
1st1: Is there a way to CI browser-like environment?

jaclarke: There's this test: https://github.com/edgedb/edgedb-js/blob/master/packages/driver/test/browser.test.ts, but I guess it doesn't perfectly replicate a browser environment as it didn't catch this.

jaclarke: Hmm, it probably needs updating to disable `node:` prefixed imports as well.

integration-tests/lts/deno/test.ts
jaclarke, line 1: @scotttrinh I think we want to keep this deno integration test right?

packages/driver/src/primitives/buffer.ts
scotttrinh: Once we get a little distance from this specific task of removing deno, I think we should treat "browser" as another platform and test it the same way we plan on testing deno (and bun, and cloudflare workers, etc) with a _real_ browser via playwright or something similar.

packages/driver/src/utils.ts
```diff
@@ -20,6 +20,13 @@ import type { ResolvedConnectConfigReadonly } from "./conUtils";
 import type { HttpSCRAMAuth } from "./httpScram";
 import type { ProtocolVersion } from "./ifaces";
 
+export {
+  Float16Array,
+  getFloat16,
+  isFloat16Array,
+  setFloat16,
+} from "@petamoriken/float16";
```
jaclarke, line 28:
```suggestion
} from "npm:@petamoriken/float16";
```

packages/driver/src/conUtils.ts
```diff
@@ -137,6 +136,10 @@ export type ResolvedConnectConfigReadonly = Readonly<
   >
 >;
 
+function getEnv(envName: string, _required = false): string | undefined {
+  return process.env[envName];
```
jaclarke, line 140: Do we need to handle the Deno specific stuff here that was in adaptor.shared.deno.ts?

scotttrinh: nope! Deno supports the `process` global: https://docs.deno.com/runtime/fundamentals/node/#node.js-global-objects

packages/driver/src/utils.ts
jaclarke: Ignore this. It's a dependency in package.json, so it should work without the npm prefix.

packages/generate/README.md
```diff
@@ -35,7 +35,7 @@ $ npx @edgedb/generate <generator> [options]
 **Deno**
 
 ```bash
-$ deno run --allow-all --unstable https://deno.land/x/edgedb/generate.ts <generator> [options]
+$ deno run --allow-all --unstable npm:@edgedb/generate <generator> [options]
```
jaclarke, line 38: Do we still need the `--unstable` flag? I guess `--allow-all` is needed for fs access.

packages/driver/src/utils.ts
diksipav: yes, and I'm quite sure this don't work with `npm:` prefix in Node

packages/driver/src/primitives/buffer.ts
1st1: +1.

packages/generate/README.md
diksipav: u are right, we don't need it, I was testing it in Deno example project without` --unstable` flag and it worked

packages/driver/src/systemUtils.ts
```diff
@@ -1,35 +1,23 @@
...
-
 import process from "node:process";
 import * as readline from "node:readline";
 import { Writable } from "node:stream";
 
-export { path, net, fs, tls, process };
-
 export async function readFileUtf8(...pathParts: string[]): Promise<string> {
   return await fs.readFile(path.join(...pathParts), { encoding: "utf8" });
 }
```
1st1, line 10: `readFileUtf8` is so trivial -- why do we need it? Why not just inline the actual call into whatever uses `readFileUtf8`?

packages/driver/src/adapter.ts
1st1: good stuff

packages/driver/src/systemUtils.ts
```diff
@@ -1,35 +1,23 @@
...
 
-// export function hasFSReadPermission(): boolean {
-//   return Deno.permissions.querySync({ name: "read" }).state === "granted";
-// }
 export function hasFSReadPermission(): boolean {
+  // @ts-ignore
+  if (typeof Deno !== "undefined") {
+    // @ts-ignore
+    return Deno.permissions.querySync({ name: "read" }).state === "granted";
+  }
```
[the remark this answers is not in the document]
scotttrinh: It's used exactly once, here: https://github.com/edgedb/edgedb-js/blob/acc8c7791d9589969e3c34cf5dd295deb18150fe/packages/driver/src/conUtils.server.ts#L14-L17

I think we can probably kill this. We should _require_ fs permissions for Deno, in my opinion.

integration-tests/lts/deno/test.ts
jaclarke: Yeah, but for now I think we should keep this test, as without it I don't think there's any tests that any of the driver/generate/qb actually works in deno.

diksipav: I'm just working on getting it back, I removed lts/deno yesterday because the tasks from deno.json were using packages/deno folder that we dont generate anymore and I rushed to make all this pass, so I thought maybe we should come up with different approach. I tested both driver and generate in an example deno project. I'll get back test now and try to make it pass

scotttrinh: If we want to keep it for safety until we can add a real integration setup for Deno, that's fine with me. Same for the browser-related testing in the driver test suite. But I think the original spirit of this change was to remove all of the package-time and test-time related special-casing for Deno, and I think this part of the integration testing flow fits into that scope.

packages/driver/src/systemUtils.ts
diksipav: we can, I mostly tried to remove deno stuff and not change too much the organisation of code that is not related to that... we use `readFileUtf8` in multiple files and in some files multiple times so I guess it was/is easier to not repeat the `{ encoding: "utf8" } ` part every time

jaclarke: I guess we don't strictly need it, but most places where we read files we're joining some path onto a base path, then reading as utf8, so this util makes it look a bit cleaner.

jaclarke: > mostly tried to remove deno stuff and not change too much the organisation of code that is not related to that

Yeah, this PR is probably sprawing enough as it is 😄.

jaclarke (approved):

integration-tests/lts/deno/test.ts
jaclarke: Ok, I'm fine with making that a follow up if we want to merge this PR now to unblock things.

diksipav:
I have one question. @scotttrinh @1st1 I updated generate pkg to use` npm:edgedb` in imports but I realised tonight that it doesn't update imports to use `:npm` prefix everywhere, in generated files that are not part of the syntax it still uses just "edgedb". This can be handled in 2 ways:
- telling users to add npm:edgedb import in their deno.json files:
```
  "imports": {
    "edgedb": "npm:edgedb@^1.5.12"
  }
```
- or to update these imports in all generated files.

In my test project I updated `deno.json` with the right import so it worked fine. But I believe it is probably better if we **fix** all generated files. My brain just doesn't work now anymore so I can continue tomorrow.

packages/driver/src/systemUtils.ts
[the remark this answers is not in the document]
diksipav: @scotttrinh if u think we can do this in another PR, and also the thing I mentioned in the comment at the end of the PR, and reintroducing Deno tests we can merge this PR now?

1st1:
> I have one question. @scotttrinh @1st1 I updated generate pkg to use npm:edgedb in imports but I realised tonight that it doesn't update imports to use :npm prefix everywhere, in generated files that are not part of the syntax it still uses just "edgedb". This can be handled in 2 ways:

I'm honestly not sure, but I'd say telling users to update deno.json should be OK. Less work for us, more appropriate in the long run anyway.

1st1 (approved):

PR #1166, opened 2025-01-14
driver: rename 2gel

diksipav:
Once we finish with renaming should check if all links to `https://github.com/....` to code inside other Gel repos are correct.


packages/driver/package.json
```diff
@@ -43,13 +43,12 @@
   },
   "scripts": {
     "typecheck": "tsc --project tsconfig.json --noEmit",
-    "build": "echo 'Building edgedb driver...' && rm -rf dist && yarn build:cjs && yarn build:cli",
+    "build": "echo 'Building gel driver...' && rm -rf dist && yarn build:cjs && yarn build:cli",
```
diksipav, line 46: This can also be Gel instead of gel? @scotttrinh

packages/driver/src/cli.mts
```diff
@@ -92,9 +92,9 @@ async function main(args: string[]) {
   process.exit(0);
 }
 
-async function whichEdgeDbCli() {
+async function whichGelCli() {
   debug("Checking if CLI is in PATH...");
-  const locations = (await which("edgedb", { nothrow: true, all: true })) || [];
+  const locations = (await which("gel", { nothrow: true, all: true })) || [];
```
diksipav, line 97: should we check here for both egdedb and cli?

scotttrinh: Yeah, if you want to update this now, we should fallback like:

```suggestion
  const locations =
    (await which("gel", { nothrow: true, all: true })) ||
    (await which("edgedb", { nothrow: true, all: true })) ||
    [];
```

packages/driver/src/httpScram.ts
```diff
@@ -46,7 +46,7 @@ export function getHTTPSCRAMAuth(cryptoUtils: CryptoUtils): HttpSCRAMAuth {
 
     // The first request must have status 401 Unauthorized and provide a
     // WWW-Authenticate header with a SCRAM-SHA-256 challenge.
-    // See: https://github.com/edgedb/edgedb/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L153-L157
+    // See: https://github.com/gel/gel/blob/09782afd3b759440abbb1b26ee19b6589be04275/edb/server/protocol/auth/scram.py#L153-L157
```
diksipav, line 49: will this be updated?

packages/driver/test/datetime.test.ts
```diff
@@ -9,12 +9,12 @@ import {
...
 import { ReadBuffer, WriteBuffer } from "../src/primitives/buffer";
 
-// EdgeDB date/time types have microsecond precision.
+// Gel date/time types have microsecond precision.
 // JS Date has millisecond precision, Temporal types have nanosecond.
 // When converting from micro -> milli, or nano -> micro, round to nearest even
 // should be used.
 
-// Tests adapted from https://github.com/edgedb/edgedb-rust/blob/master/edgedb-protocol/tests/datetime_chrono.rs
+// Tests adapted from https://github.com/gel/gel-rust/blob/master/gel-protocol/tests/datetime_chrono.rs
```
diksipav, line 17: check this

packages/driver/test/testUtil.ts
```diff
@@ -85,14 +85,12 @@ export const getServerCommand = (
   strictSecurity = true,
 ): { args: string[]; availableFeatures: string[] } => {
   const availableFeatures: string[] = [];
-  let srvcmd = `edgedb-server`;
-  if (process.env.EDGEDB_SERVER_BIN) {
-    srvcmd = process.env.EDGEDB_SERVER_BIN;
-  }
+  const srvcmd =
+    process.env.GEL_SERVER_BIN || process.env.EDGEDB_SERVER_BIN || "gel-server";
```
diksipav, line 89: I believe this should be gel-server and gel should be used in commands below?

packages/driver/src/cli.mts
diksipav: hmm is my reasoning correct: do we have `gel` or `edgedb` depends on what CLI version is installed, or what server version is installed (my CLI usually depends on the local server version I have), so we can use the latest JS driver but still have older (edgedb) CLI installed

scotttrinh: > do we have gel or edgedb depends on what CLI version is installed, or what server version is installed

It's based on which CLI version you have installed. Newer CLI versions still work with older server versions.

>  we can use the latest JS driver but still have older (edgedb) CLI installed

Yes, that's right! And we should respect that for now: running `npx gel project init` should call `edgedb project init` if you have an older CLI installed.

packages/driver/src/httpScram.ts
scotttrinh: Yes, I believe it will be gel/gel

PR #1167, opened 2025-01-14
Auth Core & Auth Express rename to Gel

diksipav:


packages/auth-express/package.json
```diff
@@ -36,16 +36,17 @@
     "@repo/tsconfig": "*",
     "@types/express": "^4.17.21",
     "@types/node": "^20.12.13",
-    "edgedb": "*",
+    "gel": "*",
+    "@gel/auth-core": "*",
```
diksipav, line 40: if we don't add this here it will not install it

PR #1172, opened 2025-01-15
EdgeDB to Gel codemod

beerose:
Example output:

CleanShot 2025-01-15 at 16 28 07@2x


scotttrinh (commented):
Love this, awesome work!

packages/codemod/package.json
```diff
@@ -0,0 +1,48 @@
+{
```
scotttrinh, line 1: Can we make this `"type": "module"`? I'm trying to ensure that going forward new packages are ESM (eventually they will all be converted).

packages/codemod/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
scotttrinh, line 1: I think we can extend one of the existing tsconfigs, maybe [this one](https://github.com/edgedb/edgedb-js/blob/master/packages/tsconfig/base.json), like in https://github.com/edgedb/edgedb-js/blob/df2faf9e4bcdfade1837f2efab0e8a61cdaa83f9/packages/driver/tsconfig.cli.json#L2

packages/codemod/package.json
beerose: Yeah, I wanted that too, but jscodeshift is not compatible with ESM. Some discussion here: https://github.com/facebook/jscodeshift/issues/500.

I was thinking of ditching jscodeshift - it seems like we only use it for imports.

beerose:
Currently, the build is failing with:

```
$ tsc -p tsconfig.json
Error: ../../node_modules/ast-types/main.d.ts(1,19): error TS2865: Import 'Type' conflicts with local value, so must be declared with a type-only import when 'isolatedModules' is enabled.
Error: ../../node_modules/ast-types/main.d.ts(2,10): error TS2865: Import 'NodePath' conflicts with local value, so must be declared with a type-only import when 'isolatedModules' is enabled.
Error: ../../node_modules/ast-types/main.d.ts(4,10): error TS2865: Import 'builders' conflicts with local value, so must be declared with a type-only import when 'isolatedModules' is enabled.
error Command failed with exit code 2.
```

`ast-types` is jscodeshift's dependency. 

Do we need `isolatedModules` set to true in the global config?

PR #1177, opened 2025-01-17
Make CLI wrapper install quieter

scotttrinh:
1. Do not prompt for confirmation when installing the CLI, even for TTYs
2. More gracefully handle errors from the CLI. `stderr` is already being
   piped to the parent process, so no need to log errors so loudly.

### DIFF
--- a/packages/driver/src/cli.mts
+++ b/packages/driver/src/cli.mts
@@ -73,16 +73,16 @@ async function main(args: string[]) {
     null;
 
   if (cliLocation === null) {
-    throw Error("Failed to find or install EdgeDB CLI.");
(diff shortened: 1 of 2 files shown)


packages/driver/src/cli.mts
```diff
@@ -28,6 +32,18 @@ interface Package {
...
+  tag: "Ok";
+  stdout: string | Buffer;
+}
+
+interface Err {
+  tag: "Err";
+  error: Error & SpawnSyncReturns<string | Buffer>;
+}
+
+type Result = Ok | Err;
```
scotttrinh, line 45: I could've made this generic, but... Yeah, maybe let me make this generic 🙈

PR #1181, opened 2025-01-21
Use correct pkg runner to run pkgs

diksipav:


packages/create/src/utils.ts
```diff
@@ -153,3 +153,35 @@ stdout: ${stdout}`,
     });
   });
 }
+
+export class ProjectManager {
+  public packageManager: PackageManager;
+
+  constructor() {
+    this.packageManager = getPackageManager();
+  }
```
scotttrinh, line 162: There are two ways we can go here, and I'm fine with either, with a slight preference for the first:

### Return a `ProjectManager` instance in `_base/index.ts`

So, roughly:

```ts
const packageManager = new PackageManager();
```

### Only instantiate this class when needed

Remove all of the `packageManager` stuff from `BaseOptions` and the base recipe, and update any recipes that use the `packageManager` option to instantiate it themselves.

---

Either way we will want to stop exporting the _type_ `PackageManager` (let's rename it or inline it so it doesn't conflict with the _class_ `PackageManager`) and the function `getPackageManager`. We probably want to add a `toString` method to the class to maybe return the package manager string, too. Just to make it cleaner when logging.

```ts
  public toString(): string {
    return this.packageManager;
  }
```

packages/create/src/utils.ts
```diff
@@ -153,3 +153,35 @@ stdout: ${stdout}`,
...
+
+  private getPackageRunner(): string {
```
scotttrinh, line 164: Small nit: since this value isn't actually dynamic, we can move this to the constructor:

```ts
const PM_TO_RUNNER_MAP = {
  "yarn": "yarn dlx",
  // etc.
} as const;

// in constructor: 
this.runner = PM_TO_RUNNER_MAP[this.packageManager];
```

Or something like that. Then just `${this.runner}` in your string builder.

diksipav:
Too many conflicts after renaming packages/driver to packages/gel. It is easier to create a new [PR](https://github.com/edgedb/edgedb-js/pull/1186).

PR #1183, opened 2025-01-23
Add edgedb release yml

diksipav:


diksipav:
My browser was too slow, so accidentally created PR for master too.

PR #1185, opened 2025-01-27
Update pkg versions

diksipav:


packages/generate/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@gel/generate",
-  "version": "0.5.6",
+  "version": "1.0.0",
```
scotttrinh, line 3: I think we should consider 0.6.0 here: there are a number of other breaking changes we might want to make before really settling on 1.0 which we will try to keep around for a long time without breaking.

diksipav: yes ofc, I'm fine with 0.6.0

scotttrinh:
@diksipav I think we also needed to change the wrapper package to point at 0.6.0 of the main package.

PR #1186, opened 2025-01-27
edgedb-js/create refactor

diksipav:
Use PackageManager class.
Use user's package runner to run the Gel CLI instead of installing it.


scotttrinh (approved):
👏

PR #1187, opened 2025-01-27
Allow customizing how SQL rows are unpacked

1st1:


packages/driver/src/codecs/context.ts
```diff
@@ -62,6 +67,19 @@ export class CodecContext {
     return NOOP;
   }
 
+  getContainerOverload<T extends ContainerNames>(
+    kind: T,
+  ): ContainerOverload<T> {
```
scotttrinh, line 72: I think we want the nullability of this return type in the signature, right?

```suggestion
  ): ContainerOverload<T> | null {
```

Otherwise, you're not required to check downstream.

packages/driver/src/codecs/record.ts
```diff
@@ -61,6 +61,11 @@ export class RecordCodec extends Codec implements ICodec {
       result[i] = val;
     }
 
+    const overload = ctx.getContainerOverload("sql_row");
+    if (overload != null) {
+      return overload.fromDatabase(result, { names: this.names });
+    }
```
scotttrinh, line 67: My suggestion in the meeting was to intercept building the result and have the `fromDatabase` build the data structure directly. So pass it an iterable that returns an "entries-like" `[key, value]` where the `value` has already been decoded from the subcodec. That means the `fromDatabase` looks more like:

```ts
const objRecord = client.withCodec({
  sql_row: {
    fromDatabase: (elements: Iterable<[string, any]>): Record<string, any> => {
      return Object.fromEntries(elements);
    },
  },
});
```
I think what that looks like is that if you _don't_ specify a container overload, we use this same iterator ourselves to fill the `result: any[]` array here like:

```ts
const results: any[] = Array.from(elements, (e) => e[1]);
```

Mostly this is a memory/space thing, but I can imagine people requesting pretty large rowsets where you don't want a bunch of allocations. We can probably do a more memory optimized version of the implementation you already have below though. Without a bunch of actual benchmarking, I'm not sure I can make my case much better than "vibes" here. 😂

packages/driver/test/client.test.ts
```diff
@@ -2602,6 +2602,42 @@ if (getEdgeDBVersion().major >= 6) {
...
+            fromDatabase(data, desc) {
+              return Object.fromEntries(
+                desc.names.map((key, index) => [key, data[index]]),
+              );
+            },
+            toDatabase() {
+              // toDatabase is required to be specified (limitation
+              // of TypeScript type system). This isn't super elegant
+              // and most likely we'll need a higher-level nicer API
+              // in top of this.
```
scotttrinh, line 2624: I assume you already tried just leaving off the `toDatabase` in the type def, and ran into some errors?

packages/driver/test/client.test.ts
```diff
@@ -2602,6 +2602,42 @@ if (getEdgeDBVersion().major >= 6) {
...
+              //
+              // Maybe `client.withCodecs(gel.SQLRowAsObject)`?
```
scotttrinh, line 2626: Yes, absolutely, 100%. Could even short-cut this with `.withSQLObjectCodec()`. I almost think that _this_ should be the default and have the array of tuples be the opt-in option.

packages/driver/src/codecs/context.ts
scotttrinh: Oh, I just noticed that this mapped type is already `| undefined`, nevermind!

packages/driver/src/codecs/record.ts
scotttrinh: Another idea would be to return a data structure that has `.entries`, `.values`, `.keys` methods that returned different iterables to avoid the extra allocations for the name if they're not needed and make the Array case more like `Array.from(elements.values())` instead of needing to define a mapping function.

1st1: > My suggestion in the meeting was to intercept building the result and have the fromDatabase build the data structure directly. So pass it an iterable that returns an "entries-like" [key, value] where the value has already been decoded from the subcodec. 

and 

> Mostly this is a memory/space thing, but I can imagine people requesting pretty large rowsets where you don't want a bunch of allocations. 

Sadly this won't work. In order to build a classic **synchronous** iterator/generator to pass it to `fromDatabase()` we'd need to **first parse the entire incoming data into an array** and then return an iterator over it.

If we want to progressively **push** values into `fromDatabase()` as we parse them, then the only way is to use an async iterator. Sadly, the overhead of that will be so large that we'll just kill performance (compare `array[i] = el` on a pre-sized array vs. a chain of callbacks and function calls).

-----

If we *really* want to make it possible for users to have super fast decoding for rows then we have to have a special codec API for collections.

`fromDatabase(values, names)` needs to be replaced with a pair of `makeFromDatabaseContainer(size)` and `setFromDatabaseContainerIndex(index, name, value)`.

This way we can efficiently pre-allocate container (if need be) and progressively populate it with values. This is the fastest way possible.

But this is very complex, so let's ask ourselves do we actually need any of this?

-----

Let's step back and analyze what's going on here (doing this also for the sake of finally documenting the decision tree somewhere).

In an ideal world we'd just have ONE data type for `querySQL`. JavaScript is dynamic enough to let us build a hybrid type, where a field can be accessed either by an index or by its name. 

We could do that via extending the `array` type (creating types dynamically per data descriptor and caching them) or by creating `object`s with a custom prototype with nice name getters.

Either approach would work, and we even implemented this before for named tuples! Everything worked super fast and was quite ergonomic. But we ended up [removing all of that](https://github.com/edgedb/edgedb-js/pull/310), opting for using pure JS objects or arrays.

Why? Because next-js (and a few other frameworks) for some reasons enforce runtime validation on user data before allowing it to be serialized to JSON. I thought this could be a thing of the past (see the last comment [in this thread](https://github.com/vercel/next.js/issues/11993), but I decided to check the latest nextjs source and found [this](https://github.com/vercel/next.js/blob/90164a10017c7a7f42d26c7ac36c4cc1db258a6c/packages/next/src/lib/is-serializable-props.ts#L49-L79) and [that](https://github.com/vercel/next.js/blob/90164a10017c7a7f42d26c7ac36c4cc1db258a6c/packages/next/src/shared/lib/is-plain-object.ts#L21). So nextjs still enforces serialization checks somewhere, and given that nextjs isn't the only offender we probably should just not attempt doing anything smart ever again.

-----

Back to square one: what do users want? They want to access values in a row by index and/or name. With JavaScript we don't have a way to allow both at the same time. So it's one or another and the user has to choose ahead of time.

This is what pglite [is offering](https://pglite.dev/docs/api#query) - `rowMode: "object" | "array"`.

Can we do this? Yes, that's what the current PR *almost* is. I'll push an update to it now to streamline it for the user further and make it close to pglite.

I still think it's useful to keep the API this PR introduces. Why? Because with the current API you can just wrap the passed values array and implement getters/setters on top of it (no need to re-allocate or copy anything). But I do now this that this would be a super low level API that few people would ever use.

packages/driver/test/client.test.ts
1st1: Believe it or not, I spent a considerable amount of time debugging typing here, ending up quite frustrating that I can't solve this puzzle.

Out of complete desperation I asked chatgpt and here's the sacred knowledge it shared with me.
CleanShot 2025-01-27 at 21 11 44@2x

This stems from this:

CleanShot 2025-01-27 at 21 13 35@2x

See the highlighted part. `CodecSpec` is used to type `client.withCodecs`. I literally tried everything, so I don't think I can solve this. You're welcome to try.

packages/driver/test/client.test.ts
1st1: OK, let's do this. I'll push some updates in a few.

1st1:
@scotttrinh @jaclarke alright, I've pushed another commit to "hide" `withCodecs` for setting up "SQL row" type and added a new high-level API `withSQLRowMode(mode: 'array' | 'object'): Client`.

If either of you decide to merge, please don't "squash" but instead "rebase". I'd like both commits of this PR to survive in the history.

packages/driver/test/client.test.ts
1st1: Well, with the new commit just pushed we don't care about this (for now).

1st1:
Alright, turns out that both `node-postgres` and `pglite` by default return arrays of objects, not arrays of arrays. And so should we.

I've just pushed another commit to flip the default.

`client.querySQL() -> array<record<string, any>>`
`client.withSQLRowMode('array').querySQL() -> array<array<any>>`

packages/driver/src/codecs/record.ts
```diff
@@ -18,11 +18,38 @@
...
 import { ReadBuffer } from "../primitives/buffer";
 import { InvalidArgumentError, ProtocolError } from "../errors";
 import type { CodecContext } from "./context";
 
+const SQLRowArrayCodec: Codecs.SQLRowCodec = {
+  fromDatabase(values: any[], _desc: { names: string[] }): any[] {
+    return values;
+  },
+  toDatabase() {
+    throw "cannot encode SQL record as a query argument";
```
jaclarke, line 32:
```suggestion
    throw new InternalClientError("cannot encode SQL record as a query argument");
```
I know it shouldn't be be possible to reach this error, but just in case.

jaclarke (approved):

packages/driver/src/codecs/record.ts
scotttrinh: I'm going to do a little bit of experimenting and benchmarking to validate our conflicting vibes here to ensure we don't end up performing poorly versus other drivers when returning SQL results. This _may_ end up "breaking" the interface, but since this isn't yet released, and the actual public interface doesn't involve setting up the codec manually, I think that's fine.

packages/driver/src/options.ts
```diff
@@ -273,6 +278,14 @@ export class Options {
       clone.codecs = this.codecs;
     }
 
+    if (mergeOptions._dropSQLRowCodec && clone.codecs.has("sql_row")) {
+      // This is an optimization -- if "sql_row" is the only codec defined
+      // and it's set to "object mode", the we want the codec mapping to be
+      // empty instead. Why? Empty codec mapping short circuits a lot of
```
beerose, line 284:
```suggestion
      // and it's set to "object mode", we want the codec mapping to be
      // empty instead. Why? Empty codec mapping short circuits a lot of
```

beerose: Was 4 minutes too late 😅

scotttrinh: ha! PRs welcome 😂

packages/driver/src/codecs/record.ts
scotttrinh: Alright! Did some benchmarking and I was not able to give a nicer interface for consumers that didn't measurably impact the performance of the default case, so I'm going to withdraw my suggestion here. The closest I got was giving the consumer an entries-like array (`[key, value][]`) which I feel like is a more natural API than `(values, { names })`, but to get the same performance involves just as much duplication of logic, and if the downstream consumer wants to ignore the names, they suffer a small penalty for doing so.

1st1: Yeah, and I also realized I was wrong suggesting that a generator-based API can't be implemented. It sure can be. But it's still going to be much slower than what we have now, so let's not bother.

PR #1191, opened 2025-02-07
Fix querybuilder `toEdgeQL` when doing 'insert many-to-one, select one' queries using `e.with`

jaclarke:
Fixes a bug where we were not also checking the child exprs of the additional valid scopes added in this fix: https://github.com/edgedb/edgedb-js/pull/1154.

Also fixes bug where scope validity was not being checked in the correct order when some expressions in the `e.with` refs array are children of other exprs.


scotttrinh (approved):
🫶

PR #1192, opened 2025-02-07
Fix querybuilder generated query when using unpathed scoped expr

### LINKED ISSUE #808: Query builder builds broken query
**Code**
The code causing the error.

```typescript
import {createClient} from 'edgedb';
import e from './dbschema/edgeql-js';

const client = createClient({});

async function run() {
  const query = e.select(
    e.select(e.Movie, (movie) => ({
      filter: e.op(movie.id, '=', e.uuid('1f406dd0-f1c7-11ed-ab80-d32f12dd164e'))
    })),
    (movie) => ({
      comp: e.op('exists', movie.is(e.Movie))
    }),
  );

  console.log(query.toEdgeQL());
  const result = await query.run(client);
}

run();
```

**Schema**

From edgedb/edgedb-movies

**Generated EdgeQL**

```
WITH
  __withVar_2 := (__scope_1_defaultMovie.id = <std::uuid>"1f406dd0-f1c7-11ed-ab80-d32f12dd164e"),
  __scope_0_defaultMovie := (
    WITH
      __scope_1_defaultMovie := DETACHED default::Movie
    SELECT __scope_1_defaultMovie {
      id
    }
    FILTER __withVar_2
  )
SELECT __scope_0_defaultMovie {
  single comp := (exists (__scope_0_defaultMovie {id})[IS default::Movie])
}
```

**Error or desired behavior**

```
InvalidReferenceError: object type or alias 'default::__scope_1_defaultMovie' does not exist
   |
 2 |   __withVar_2 := (__scope_1_defaultMovie.id = <std::uuid>"1f406dd0-f1c7-11ed-ab80-d32f12dd164e"),
   |                   ^^^^^^^^^^^^^^^^^^^^^^

```

- `edgedb-js` version (e.g. `1.4.1`;):
- `@edgedb/generate` version (e.g. `0.4.1`;):

jaclarke:
Fixes #808

When walking the expr tree the scoped copy of a select/update expr passed to the shape should have been treated as an opaque reference and not walked, since it will get walked again by the enclosing select/update.

Currently we are only checking this in path expressions (eg. `movie.title`) and not when the scoped expr appears anywhere else (eg. `movie.is(e.Movie)`). And since the expr is getting walked multiple times, it's messing up the ref counting, and exprs are getting hoisted when they shouldn't, causing the broken queries in #808.

### DIFF
--- a/packages/generate/src/syntax/toEdgeQL.ts
+++ b/packages/generate/src/syntax/toEdgeQL.ts
@@ -397,6 +397,13 @@ function walkExprTree(
 
   const expr = _expr as SomeExpression;
 
+  if ((expr as any).__scopedFrom__ != null) {
+    // If expr is marked as being a scoped copy of another expr, treat it as
+    // an opaque reference and don't walk it. The enclosing select/update that
+    // owns the scope will walk the actual unscoped expr.
+    return [expr];
+  }
+
   function walkShape(shape: object) {
     for (let param of Object.values(shape)) {
       if (param.__kind__ === ExpressionKind.PolyShapeElement) {
@@ -473,16 +480,9 @@ function walkExprTree(
     case ExpressionKind.PathLeaf:
     case ExpressionKind.PathNode:
       if (expr.__parent__) {
-        if ((expr.__parent__.type as any).__scopedFrom__) {
-          // if parent is scoped expr then don't walk expr
-          // since it will already be walked by enclosing select/update
-
-          childExprs.push(expr.__parent__.type as any);
-        } else {
-          childExprs.push(
-            ...walkExprTree(expr.__parent__.type, parentScope, ctx),
-          );
-        }
+        childExprs.push(
+          ...walkExprTree(expr.__parent__.type, parentScope, ctx),
+        );
 
         if (
           // is link prop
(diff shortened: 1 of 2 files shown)


packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -473,16 +480,9 @@ function walkExprTree(
     case ExpressionKind.PathLeaf:
     case ExpressionKind.PathNode:
       if (expr.__parent__) {
-        if ((expr.__parent__.type as any).__scopedFrom__) {
```
scotttrinh, line 476: Are we losing this special check or is the `expr.__scopedFrom__` going to catch the same things that this `expr.__parent__.type.__scopedFrom__` check is looking for?

jaclarke: Yes, the scoped expr check for `expr.__parent__.type` is now just inside the `walkExprTree` call, so it also catches the scoped exprs wherever they're used, and not just in paths.

scotttrinh: ahh, that makes sense 👍

scotttrinh (approved):
Glorious.

PR #1193, opened 2025-02-13
Implement Transaction.withSQLRowMode() method

1st1:
Uncover two bugs in the process:

- SQL codecs were accessed in different places by two different names "_private_sql_row" and "sql_row", so really SQL return configuration only worked once :)

- There was a pathway when options object leaked codecs from one version to another (because we want to make things fast!)

Both bugs are fixed as part of this commit.


scotttrinh (approved):
💯

PR #1195, opened 2025-02-13
Optimize codecs context cacheing

1st1:
We should really copy the cached codex context for cloned Options instances (unless .withCodecs() was called).

### DIFF
--- a/packages/driver/src/options.ts
+++ b/packages/driver/src/options.ts
@@ -178,8 +178,10 @@ export class Options {
   /** @internal */
   readonly annotations: ReadonlyMap<string, string> = new Map<string, string>();
 
-  private cachedCodecContext: CodecContext | null = null;
-  private cachedCodecContextVer = -1;
+  /** @internal */
+  cachedCodecContext: CodecContext | null = null;
+  /** @internal */
+  cachedCodecContextVer = -1;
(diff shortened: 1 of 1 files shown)


scotttrinh (approved):
Nice improvement 👍

packages/driver/src/options.ts
```diff
@@ -269,30 +271,36 @@ export class Options {
...
       ]) as ReadonlyCodecMap;
     } else {
       clone.codecs = this.codecs;
-      originalCodecsMap = true;
+      // It  makes sense to just inherit the cached codec context
+      // as it's expensive to build. The only case when it can't be
+      // cached is when .withCodecs() was called, but caching will
+      // work great for all other .withSomething() methods we have.
+      clone.cachedCodecContext = this.cachedCodecContext;
+      clone.cachedCodecContextVer = this.cachedCodecContextVer;
```
jaclarke, line 286:
```
clone.cachedCodecContext = this.makeCodecContext();
```
If we want maximum caching™ we should make sure the `cachedCodecContext` is actually initalised before we copy it.

1st1: There's a check when the cache is used, I think that check is enough.

PR #1197, opened 2025-02-14
Gel prerelease script

diksipav:


scotttrinh (approved):
cool! let's give it a shot.

packages/ai/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@gel/ai",
   "description": "Helper library for the Gel AI extension",
-  "version": "0.1.0",
+  "version": "0.1.0-rc.1",
```
scotttrinh, line 4: oh, actually, the test failures are right here: I think because these are prereleases, they're not matching. I'll fix this up.

scotttrinh (changes requested):

PR #1199, opened 2025-02-18
Update CLI wrapper for `gel` command

scotttrinh:
Now that the package is called `gel`, we need to switch to checking `gel` in argv. Along with this change are a few more actual logical changes:

1. Prefer `PATH` to find the binary instead of the cached location.
2. _Only_ check for the command we're wrapping when using `which` to find the path to the CLI binary.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -139,11 +139,6 @@ jobs:
         with:
           node-version: 20
 
-      - name: Set up Deno
-        uses: denoland/setup-deno@v2
-        with:
-          deno-version: v2.x
-
       - name: Install dev deps
         run: |
           yarn --frozen-lockfile
@@ -162,6 +157,12 @@ jobs:
           npm exec gel -- project init --non-interactive
           npm exec gel -- --version
           npm exec gel -- query 'select sys::get_version_as_str()'
+          if gel --version; then
+            echo "Error: gel command should not be available outside of npm exec"
+            exit 1
+          else
+            echo "gel command is not available outside of npm exec, as expected"
+          fi
 
       - name: Test CLI wrapper with yarn
         run: |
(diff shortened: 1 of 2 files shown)


1st1:
@elprans has to review

diksipav (approved):

diksipav (commented):
I believe we still don't do everything completely right for cashed version:
```
const IS_TTY = process.stdout.isTTY;
const SCRIPT_LOCATION = await fs.realpath(fileURLToPath(import.meta.url));
const EDGEDB_PKG_ROOT = "https://packages.edgedb.com";
const CACHE_DIR = envPaths("edgedb").cache;
const TEMPORARY_CLI_PATH = path.join(CACHE_DIR, "/edgedb-cli");
const CLI_LOCATION_CACHE_FILE_PATH = path.join(CACHE_DIR, "/cli-location");
```
because at my mac `cli-location` is inside `/Users/dijanapavlovic/Library/Caches/edgedb-nodejs` and other relevant things are inside Caches/edgedb. Why don't we store location inside edgedb folder and then if there's need for node-cli an deno-cli we can diferentiate inside edgedb folder.

scotttrinh:
> because at my mac cli-location is inside /Users/dijanapavlovic/Library/Caches/edgedb-nodejs and other relevant things are inside .../Caches/edgedb. Why don't we store location inside edgedb folder and then if there's need for node-cli an deno-cli we can diferentiate inside edgedb folder.

Mostly I just stuck with the defaults, and there is [even a warning in the `env-paths` package to be careful using the prefix option to avoid conflicts with other native apps.](https://www.npmjs.com/package/env-paths#suffix). Luckily we kind of "own" this un-prefixed path ourselves. Also, this is unrelated to node vs deno: it's meant to differentiate between _this_ CLI wrapper script and anything else that might use the name `edgedb`.

Just to be clear, the purpose of this particular cache directory is to:

1. Cache the temporary CLI which we use to find the probable path to an existing CLI binary, and if it doesn't exist, to run the self-installer. Once this is done, we can just remove this temporary CLI.
2. Cache the location that we find of the CLI binary. This covers the case where the user has an unusual shell and we cannot automatically add the binary location to the PATH using the typical means of updating the shell profile. We also use this cached value to avoid needing to restart the shell to pick up the updated PATH.

As long as we're consistent here, I don't personally care what this directory is, so I'm fine with dropping the `-nodejs` suffix if we want to store everything together, we just need to be a bit more defensive about ensuring we don't accidentally overwrite something that some other tool (like uvx, or the CLI itself) uses.

> Are we going to update these locations to gel eventually?

Yeah, I'm fine with switching this now. I think the new uvx tool uses `gel` already.

---

FWIW, I don't think these issues are a blocker for merging what we have here: these are non-functional issues.

PR #1214, opened 2025-02-25
Use correct package ranges in create recipes

scotttrinh:
We should consider using a more robust way of managing these ranges to avoid needing to always touch these directly, while also avoiding making them roughly "*".

This only "broke" because we don't have these old versions published and `^` for 0.x versions won't automatically install the latest minor version since that's considered a breaking change.


diksipav:
I also wanted to ask when is going to be the right time to publish 1.0.0 versions of auth packages? They seem stable enough. Are we going to add more features soon?

PR #1219, opened 2025-02-28
Updated pkce cookie with new default name

dotlouis:
fixed the missing occurrences of the PKCE session cookie name.


dotlouis:
Ok, I just figured out that keeping the old name is for backward compatibility.
My bad, closing the PR.

PR #1220, opened 2025-02-28
Require explicit index checks or assertions

scotttrinh:
Many places in the code were using unchecked indexed access, which conflicts with many people's TypeScript configurations. We copy the hydrate.ts module directly into the generated code, so we will now check the rest of the generate package with the same strict configuration.

An alternative approach would be to separate out the code we copy into the query builder into a more strict subset and keep the looser config for our own code, but I think stricter is better anyway.

Most places just got the assertion treatment, but in a few places it was straightforward enough to actually do something more safe.

### DIFF
--- a/packages/generate/src/gel-prisma.ts
+++ b/packages/generate/src/gel-prisma.ts
@@ -128,7 +128,7 @@ function getMod(fullname: string): string {
   return <string>fullname.split("::").slice(0, -1).join("::");
 }
 
-function getModAndName(fullname: string): string[] {
+function getModAndName(fullname: string): [string, string] {
   const mod = fullname.split("::");
   const name = <string>mod.pop();
   return [mod.join("::"), name];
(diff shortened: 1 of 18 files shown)


packages/generate/src/gel-prisma.ts
```diff
@@ -272,7 +272,7 @@ function processLinks(types: JSONType[], modules: string[]): ProcessedSpec {
 
           link_objects.push(lobj);
           link["has_link_object"] = true;
-          objtype["backlinks"][-1]["has_link_object"] = true;
+          objtype["backlinks"].at(-1)!["has_link_object"] = true;
```
scotttrinh, line 275: @vpetrovykh This was the one major change I made across this whole refactor: was this the intention of this code path? I think you _can_ use a negative index on an array:

```
> const a = [];
undefined
> a[-1] = "beep";
'beep'
> a
[ '-1': 'beep' ]
```

But I doubt that's what you were _trying_ to do here?

jaclarke (approved):

packages/generate/src/gel-prisma.ts
vpetrovykh: Yeah, your fix is correct, I meant the last element of the array, not the actual index '-1'.

PR #1230, opened 2025-03-07
Bump to latest setup-gel action

scotttrinh:


.github/workflows/tests.yml
```diff
@@ -63,20 +63,20 @@ jobs:
           yarn format
 
       - name: Install Gel
-        uses: geldata/setup-gel@6763b6de72782d9c2e5ecc1095986a1c707da68f
+        uses: geldata/setup-gel@358410cc5618fa956d16f1b7157a8c86f9b517cf # v1.4
```
elprans, line 66: This is our trusted action, so just using `v1` is fine.

elprans (approved):

.github/workflows/tests.yml
scotttrinh: Well, I mostly mean this to be stable across runs rather than a lack of trust of the upstream. But I'm not sure what we actually get from that, so fair enough!

PR #1238, opened 2025-03-19
Tweak WarningHandler to use subclass in signature & export

CarsonF:


packages/gel/src/options.ts
```diff
@@ -43,22 +43,22 @@ export interface SimpleRetryOptions {
   backoff?: BackoffFunction;
 }
 
-export type WarningHandler = (warnings: errors.GelError[]) => void;
+export type WarningHandler = (warnings: errors.WarningMessage[]) => void;
```
scotttrinh, line 46: This is a breaking change, right? Could we keep the old signature even though it's "wider" than we'd like? I guess the whole point of this change is to not require manual narrowing in your subtype, right?

CarsonF: It's not a breaking change IMO. The input being handed to users still fulfills `GelError` shape. It would be breaking to go the other way.

Aside: the only type incompatibility is that the base error does not have a `code: number` prop. Maybe this is something to be considered?
I could also see it the other way, where `code` could be worthless since we have `instanceof` checks.

scotttrinh: > The input being handed to users still fulfills GelError shape. It would be breaking to go the other way.

Ah, it's covariant in this case because the parameters are contravariant, that makes sense I think. I always have a hard time intuiting this 🤯 , but I just double checked in the test suite. Do you mind adding a wee test case to just ensure this stays backwards compatible with the old signature?

> Aside: the only type incompatibility is that the base error does not have a code: number prop. Maybe this is something to be considered?
> I could also see it the other way, where code could be worthless since we have instanceof checks.

I think the explicit `code` helps with serializing and deserializing errors, so I think it makes sense to keep them. But, I wonder if we should do either or both of the following things:

1. Add a `code?: number` prop to the `GelError` base type
2. Make the return type of `get code()` be a number literal type instead of `number`

scotttrinh:
Looks like the warning system can actually warn on any kind of error, so I think what we currently have is correct.

CarsonF:
Ah I see. https://github.com/geldata/gel/pull/564
The `WarningMessage` has nothing to do with the new warning system.

PR #1239, opened 2025-03-19
Export WarningHandler type / Share options exports with browser index

CarsonF:
Idk if actually moving the options exports to shared helps the browser index at all. I suspect not, since it's gone this long without. They've probably never been needed at runtime.
Regardless, this feels more correct, and it is saner to reason about too internally IMO.


scotttrinh (approved):
I agree we want these in the browser export 👍

PR #1240, opened 2025-03-19
Adjust default warning handlers for DX

CarsonF:
It is more idiomatic to merge a list of errors with `AggregateError.errors` rather than `Error.cause`.

For logging, giving the original error to the wrapped one in `cause`, causes the error & stacktrace to be duplicated/nested on the console.

### DIFF
--- a/packages/gel/src/options.ts
+++ b/packages/gel/src/options.ts
@@ -46,17 +46,16 @@ export interface SimpleRetryOptions {
 export type WarningHandler = (warnings: errors.GelError[]) => void;
 
 export const throwWarnings: WarningHandler = (warnings) => {
-  throw new AggregateError(
-    warnings,
-    `warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
-  );
+  throw new AggregateError(warnings, formatWarnings(warnings));
 };
 
 export const logWarnings: WarningHandler = (warnings) => {
-  for (const warning of warnings) {
-    console.warn(new Error(`Gel warning: ${warning.message}`));
-  }
+  const merged = new Error(formatWarnings(warnings));
+  console.warn(Object.assign(merged, { name: "" }));
 };
+
+const formatWarnings = (warnings: errors.GelError[]) =>
+  `warnings occurred while running query:\n${warnings.map((warn) => warn.message).join("\n")}`;
 
 export class RetryOptions {
   // This type is immutable.
--- a/packages/gel/src/options.ts
+++ b/packages/gel/src/options.ts
@@ -48,7 +48,7 @@ export type WarningHandler = (warnings: errors.GelError[]) => void;
 export const throwWarnings: WarningHandler = (warnings) => {
   throw new AggregateError(
     warnings,
-    `Warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
+    `warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
   );
 };
 
--- a/packages/gel/src/options.ts
+++ b/packages/gel/src/options.ts
@@ -46,17 +46,15 @@ export interface SimpleRetryOptions {
 export type WarningHandler = (warnings: errors.GelError[]) => void;
 
 export const throwWarnings: WarningHandler = (warnings) => {
-  throw new Error(
-    `warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
-    { cause: warnings },
+  throw new AggregateError(
+    warnings,
+    `Warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
   );
 };
 
 export const logWarnings: WarningHandler = (warnings) => {
   for (const warning of warnings) {
-    console.warn(
-      new Error(`Gel warning: ${warning.message}`, { cause: warning }),
-    );
+    console.warn(new Error(`Gel warning: ${warning.message}`));
   }
 };
 
--- a/packages/gel/test/client.test.ts
+++ b/packages/gel/test/client.test.ts
@@ -2216,9 +2216,11 @@ test("warnings handler", async () => {
 
     client = client.withWarningHandler(throwWarnings);
 
-    await expect(client.query("select _warn_on_call();")).rejects.toThrow(
-      /warnings occurred while running query: Test warning please ignore/,
+    const queryRes = client.query("select _warn_on_call();");
+    await expect(queryRes).rejects.toThrow(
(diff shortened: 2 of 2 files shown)


packages/gel/src/options.ts
```diff
@@ -46,17 +46,15 @@ export interface SimpleRetryOptions {
...
+    `Warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
   );
 };
 
 export const logWarnings: WarningHandler = (warnings) => {
   for (const warning of warnings) {
-    console.warn(
-      new Error(`Gel warning: ${warning.message}`, { cause: warning }),
-    );
+    console.warn(new Error(`Gel warning: ${warning.message}`));
```
scotttrinh, line 57: How about just:

```suggestion
    console.warn(warning);
```

?

I guess maybe we want to indicate that this output is just some "warn"-level logging?

Maybe

```ts
export const logWarnings: WarningHandler = (warnings) => {
  if (warnings.length) {
    console.warn("Warning from Gel:");
  }
```

packages/gel/src/options.ts
```diff
@@ -46,17 +46,15 @@ export interface SimpleRetryOptions {
 export type WarningHandler = (warnings: errors.GelError[]) => void;
 
 export const throwWarnings: WarningHandler = (warnings) => {
-  throw new Error(
-    `warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
-    { cause: warnings },
+  throw new AggregateError(
+    warnings,
+    `Warnings occurred while running query: ${warnings.map((warn) => warn.message)}`,
```
scotttrinh, line 51: Let's keep the original casing here for better backwards compat. (see the failing tests, but also some folks could've stumbled into this already?)

scotttrinh (changes requested):

packages/gel/src/options.ts
CarsonF: I'm on the fence as well about whether we need to be explicit about this being a warning. We are calling `console.warn`, but that level isn't really represented in any default way. The message also starts with "QueryError: ..." (the type), so that is confusing.

I can't say I'm a fan of multiple subsequent console calls to facilitate line breaks. But I guess that could just be translated to
```ts
console.warn("Warning from Gel:\n" + warning.stack);
```
I'd want to confirm the `stack` has the correct `name`/`message` strings.

scotttrinh: Actually, I think this is my preferred solution: our default handlers collect all warnings into an `AggregateError` and either throw or `console.warn` that error with no further processing.

CarsonF: Let me try and see what that looks like

CarsonF: It doesn't look that good.
```ts
    const agg = new AggregateError(
      warnings,
      `warnings occurred while running query:\n${warnings
        .map((warn) => warn.message)
        .join('\n')}`,
    );
    console.warn(agg);
```
Screenshot 2025-03-19 at 2 42 23 PM

scotttrinh: This seems fine to me, honestly. I personally don't much care for the duplication of the warnings' messages in the `AggegrateError.message`, but that ship has somewhat sailed.

CarsonF: The these warnings all have the same effective stack trace for users. Why duplicate them & clutter the effective info?
```ts
const agg = new Error(
  `warnings occurred while running query:\n${warnings
    .map((warn) => warn.message)
    .join('\n')}`,
);
agg.name = '';
console.warn(agg);
```
Screenshot 2025-03-19 at 2 50 38 PM

CarsonF: Honestly the duplicate stack trace with `cause` is how I get here in the first place.

scotttrinh: I mean, yeah, I agree with you: this is noisier than I would like, but it seems like the least opinionated way of doing it while also being consistent with the throwing version. 

I'm pretty on the fence here: the default `console`-based logging here is just pretty bad, and I think _that's_ really what feels frustrating here.

Alright, I guess I'm also fine with removing the stack traces like you're doing here and just concatenating all of the messages. Once we have time to figure out proper telemetry hopefully this will be easier to have a sane default.

CarsonF: I agree that the console interface is poor. Actually I think what yall have with a configurable handler is best, and I'm already customizing for what fits me (that's different than what is here).
I was just trying to upstream my impression of a better DX default.

scotttrinh: yep, and I appreciate you taking the time to interate with me here! Let's take your approach of making a single new `Error` object with the warning messages concatenated into the message, and just `console.warn`ing that.

CarsonF: Should I keep the line break and update tests or remove it?

packages/gel/src/options.ts
```diff
@@ -46,18 +46,17 @@ export interface SimpleRetryOptions {
...
 export const logWarnings: WarningHandler = (warnings) => {
-  for (const warning of warnings) {
-    console.warn(new Error(`Gel warning: ${warning.message}`));
-  }
+  const merged = new Error(formatWarnings(warnings));
+  console.warn(Object.assign(merged, { name: "" }));
 };
 
+const formatWarnings = (warnings: errors.GelError[]) =>
+  `warnings occurred while running query:\n${warnings.map((warn) => warn.message).join("\n")}`;
```
scotttrinh, line 58: Just bump: https://github.com/geldata/gel-js/blob/34bca0b58f9676e16014644b25fbb6355e90a815/packages/gel/test/client.test.ts#L2220 to do two separate searches maybe?

```ts
/warnings occurred while running query/
/Test warning please ignore/
```

?

CarsonF: Actually it appears this is just testing the `throwWarnings` function. maybe we should just call it directly?

CarsonF: That's tested above

scotttrinh: I think I'd still rather keep this test as a full integration example rather than break out a separate unit test. There is still some value here to catch if we accidentally swallow errors in some change to the `warningHandler` handling in the future. I'd be fine to _add_ a unit test though, too, but that seems pointless when we have an integration test covering the unit.

scotttrinh (approved):

Jazzyjgirl1976 (approved):

PR #1241, opened 2025-03-20
Return query warnings from the `rawParse/Execute` methods used by UI

jaclarke:
Currently the UI can't show query warnings as `rawParse`/`rawExecute` don't return them. Also just return the whole return values array from the internal `_parse` method in `rawParse` instead of re-ordering it for no real reason.

### DIFF
--- a/packages/gel/src/fetchConn.ts
+++ b/packages/gel/src/fetchConn.ts
@@ -19,7 +19,7 @@
 import {
   BaseRawConnection,
   Capabilities,
-  ParseResult,
+  type ParseResult,
   PROTO_VER,
   RESTRICTED_CAPABILITIES,
 } from "./baseConn";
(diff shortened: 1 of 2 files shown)


scotttrinh:
I realize these are pretty obscure methods that do not typically get used, but they are public methods, right? Is there any way to maybe expose new methods instead? Or at least, we need to be considerate of this kind of change. Thoughts?

jaclarke:
> I realize these are pretty obscure methods that do not typically get used, but they are public methods, right? Is there any way to maybe expose new methods instead? Or at least, we need to be considerate of this kind of change. Thoughts?

They're methods on `AdminUIFetchConnection` which is intended as private/undocumented api that's only meant for the UI (though it's technically exported/public so the UI can import it). So I don't think we need to keep backwards compatibility.

PR #1244, opened 2025-03-25
Export `SimpleConfig` via `index.shared.ts`

jackfischer:
Best guess at a fix for https://github.com/geldata/gel-js/issues/1243


CarsonF:
None of the `Simple*` types are exported. Perhaps there's a more descriptive name than this, if we are going to export.

scotttrinh:
Closing for #1262

PR #1245, opened 2025-03-25
Support windows in npx gel

mmastrac:
... at least in theory. I don't have a great way to test this but it looks correct.

Partial fix for https://github.com/geldata/gel-editor-plugin/issues/43


scotttrinh:
Fixed in https://github.com/geldata/gel-js/pull/1260

PR #1257, opened 2025-04-06
Fix parsing bug in parseRagResponse

Gobot1234:


scotttrinh (approved):
ay, thanks for finding that. Slipped through in my refactor. Will get another patch release out fixing it!

PR #1260, opened 2025-04-07
Match CLI package logic to Python wrapper

scotttrinh:
- Drop the index lookup and assemble the CLI package URL directly.
- Add support for `win32` platform.
- Drop `-gnu` libc: this was never used and we don't build for it.


mmastrac (approved):
LGTM, uncertain about the build failure however

scotttrinh:
> uncertain about the build failure however

Just good ol dependency management stuff. I'll get that sorted.

scotttrinh:
Going to revert removing `semver` from the dependency list for now so I can work on landing that separately from this fix.

PR #1266, opened 2025-04-16
Fix handling of assert_single wrapped select/update exprs when used as e.with expr

jaclarke:
When using `filter_single` with `e.op` in select or update, we wrap the output select/update expr in an `assert_single()` call to ensure type correctness, but lie to the typesystem that the returned expression is still a select/update expr.

So when one of these wrapped select/update expressions is passed to `e.with` as the `expr` arg, it gets tripped up by the expr actually being an assert_single function expression instead of a 'withable' expr, and fails to attach the refs to the with block of the wrapped select/update expr, generating broken queries.

This PR updates `e.with` to handle `assert_single` wrapped select/update exprs by moving the `assert_single` to wrap the whole `with` expr, so refs are correctly attached to the inner select/update expr by `toEdgeQL`.

### DIFF
--- a/integration-tests/lts/select.test.ts
+++ b/integration-tests/lts/select.test.ts
@@ -1739,4 +1739,26 @@ SELECT __scope_0_defaultPerson {
 
     await query2.run(client);
   });
+
+  test("assert_single wrapped select as expr in e.with", async () => {
+    const movies = e.select(e.Movie, () => ({
+      title: true,
+    }));
+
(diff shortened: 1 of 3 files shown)


1st1 (approved):
I'm barely qualified to review, but the comments / idea make me think I understand this PR and it looks good. :) Feel free to release if you don't think we need to wait until @scotttrinh is back.

PR #1268, opened 2025-04-21
[QB] Optimize scalar cast maps

### LINKED ISSUE #1242: QB getSharedParentScalar excessive type depth
Image

https://github.com/SeedCompany/cord-api-v3/blob/gel-bug/v6-qb-shared-parent-scalar/src/core/gel/generated-client/set.ts#L150

https://github.com/SeedCompany/cord-api-v3/blob/gel-bug/v6-qb-shared-parent-scalar/src/core/gel/generated-client/castMaps.ts#L241

CarsonF:
Fixes #1242.

## `scalarCastableFrom`/`scalarAssignableBy`

This changes the `scalarCastableFrom`/`scalarAssignableBy` types to just pick the type based on the given type's FQN.
This is much easier for TS, esp with generated enums, as their members are statically written as well and TS was duck checking all of these.
```diff
- export type scalarAssignableBy<T extends $.ScalarType> =
- T extends _std.number ? _std.number :
- T extends _std.int64 ? _std.int64 :
- ...
+ export interface ScalarAssignableByMap {
+   "std::number": _std.$number;
+   "std::int64": _std.$int64;
+   ...
+ }
```

## `getSharedParentScalar`

This also changes `getSharedParentScalar` to compare against the types' FQN.
This should be much more performant and still safe since these are scalar types with unique names.
We can also skip enumerating every case if `A` & `B` are the same.
This saves having to declare mapping for 80% of the scalars that can't be cast into another.

This removes 79 branches with my generated client, leaving only 6 specific scalar branches.
Before:
```ts
type getSharedParentScalar<A, B> =
  A extends _cal.$local_date ?
    B extends _cal.$local_datetime ?
      B
    :
    B extends _cal.$local_date ?
      B
    :
    never
  :
  A extends _std.$number ?
    B extends _std.$number ?
      B
    :
    never
  :
  A extends _sys.$VersionStage ?
    B extends _sys.$VersionStage ?
      B
    :
    never
  :
  ...
```
After:
```ts
type getSharedParentScalar<A, B> =
  AType extends $.ScalarType<infer AName> ?
  BType extends $.ScalarType<infer BName> ?
    AName extends "cal::local_date" ?
      BName extends "cal::local_datetime" ?
        B
      :
      BName extends "cal::local_date" ?
        B
      :
      never
    :
    ...
    AName extends BName ? A : never
```

### DIFF
--- a/packages/generate/src/edgeql-js/generateCastMaps.ts
+++ b/packages/generate/src/edgeql-js/generateCastMaps.ts
@@ -51,13 +51,13 @@ export const generateCastMaps = (params: GeneratorParams) => {
   assignableMap.writeln([
     t`export `,
     dts`declare `,
-    t`type scalarAssignableBy<T extends $.ScalarType> =`,
+    t`interface ScalarAssignableByMap {`,
   ]);
   const castableMap = new CodeBuffer();
   castableMap.writeln([
     t`export `,
     dts`declare `,
-    t`type scalarCastableFrom<T extends $.ScalarType> =`,
+    t`interface ScalarCastableFromMap {`,
   ]);
 
   const staticMap = new CodeBuffer();
@@ -83,22 +83,22 @@ export const generateCastMaps = (params: GeneratorParams) => {
 
   for (const outer of materialScalars) {
     assignableMap.writeln([
-      t`  T extends ${getRef(outer.name)} ? ${
+      t`  "${outer.name}": ${
         getStringRepresentation(types.get(outer.id), {
           types,
           casts: casts.assignableByMap,
           castSuffix: "λIAssignableBy",
         }).staticType
-      } : `,
+      };`,
     ]);
     castableMap.writeln([
-      t`  T extends ${getRef(outer.name)} ? ${
+      t`  "${outer.name}": ${
         getStringRepresentation(types.get(outer.id), {
           types,
           casts: casts.implicitCastFromMap,
           castSuffix: "λICastableTo",
         }).staticType
-      } : `,
+      };`,
     ]);
 
     const outerCastableTo = casting(outer.id);
@@ -168,8 +168,8 @@ export const generateCastMaps = (params: GeneratorParams) => {
     ]);
     runtimeMap.writeln([r`    }`]);
   }
-  assignableMap.writeln([t`  never\n`]);
-  castableMap.writeln([t`  never\n`]);
+  assignableMap.writeln([t`}\n`]);
+  castableMap.writeln([t`}\n`]);
   staticMap.writeln([t`never\n`]);
   runtimeMap.writeln([
     r`  throw new Error(\`Types are not castable: \${a.__name__}, \${b.__name__}\`);`,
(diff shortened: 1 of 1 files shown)


packages/generate/src/edgeql-js/generateCastMaps.ts
```diff
@@ -62,10 +62,25 @@ export const generateCastMaps = (params: GeneratorParams) => {
 
   const staticMap = new CodeBuffer();
   staticMap.writeln([dts`declare `, t`type getSharedParentScalar<A, B> =`]);
+  staticMap.writeln([t`  IsEqual<A, B> extends true ? A :`]);
+
   const runtimeMap = new CodeBuffer();
 
   const returnTypes = new Set<string>();
 
+  const uncastables = new Set(
```
CarsonF, line 71: Like uncrustables but saltier.
Open to different names.

packages/generate/src/edgeql-js/generateCastMaps.ts
```diff
@@ -87,7 +102,9 @@ export const generateCastMaps = (params: GeneratorParams) => {
     ]);
 
     const outerCastableTo = casting(outer.id);
-    staticMap.writeln([t`  A extends ${getRef(outer.name)} ?`]);
+    if (!uncastables.has(outer)) {
```
CarsonF, line 105: All of these ifs feel dirty, but I understand wanting to keep the runtime logic next to compile time. But I'm not sure now if we should deviate since the compile time is changing its logic somewhat.

packages/generate/src/edgeql-js/generateCastMaps.ts
```diff
@@ -400,4 +424,8 @@ export const generateCastMaps = (params: GeneratorParams) => {
   ]);
   f.writeln([r`}`]);
   f.addExport("literalToTypeSet");
+
+  f.writeln([t`type IsEqual<A, B> =`]);
+  f.writeln([t`  (<G>() => G extends A & G | G ? 1 : 2) extends`]);
+  f.writeln([t`  (<G>() => G extends B & G | G ? 1 : 2) ? true : false`]);
```
CarsonF, line 430: Perhaps useful/generic enough to promote to somewhere else.

CarsonF:
This is to prevent an excessive depth error, and it appears that the current TS benchmarks do not measure these depth calculations.

CarsonF:
Just tried v6 on my codebase again with this. This does fix the depth problem with this type alias, but `assignableBy` & `castableFrom` are still a problem.

packages/generate/src/edgeql-js/generateCastMaps.ts
scotttrinh: Nope, this is the perfect name 👨‍🍳

packages/generate/src/edgeql-js/generateCastMaps.ts
scotttrinh: Oh, I'm perfectly happy to make this division more clear _here_ vs in the generated code: we should prioritize the maintenance of this code over the understandability of the generated code. No change needed, but I'm 100% with you on wanting to branch harder here instead of littering `if` statements everywhere.

scotttrinh (approved):
👑

CarsonF:
Thanks for the review @scotttrinh.
I think the remaining issues for me with `assignableBy` & `castableFrom` come from the need look at each member of all my enums.
```ts
export type scalarCastableFrom<T extends $.ScalarType> =
  T extends _Project.$Status ? _Project.$Status :
  ...;

export type $Status = {
  "InDevelopment": $.$expr_Literal<$Status>;
  "Active": $.$expr_Literal<$Status>;
  "Terminated": $.$expr_Literal<$Status>;
  "Completed": $.$expr_Literal<$Status>;
  "DidNotDevelop": $.$expr_Literal<$Status>;
} & $.EnumType<"Project::Status", ["InDevelopment", "Active", "Terminated", "Completed", "DidNotDevelop"]>;
```

What do you think about _only_ checking the FQNs instead?
```ts
export type scalarCastableFrom<T extends $.ScalarType> =
  T extends $.ScalarType<infer N> ?
    N extends "Project::Status" ? _Project.$Status :
    ...
  : never;
```
Actually following this train of thought we could just use a mapping instead.
Thoughts?
Is there a use case I'm not thinking of where only checking the name is insufficient?

CarsonF:
Ok I'm done here. These changes do fix all related problems in my codebase, blocking my v6 upgrade.
Perhaps a future PR could untangle what's here more.

scotttrinh (approved):

PR #1270, opened 2025-04-22
Set default transaction options from `withTransactionOptions`

scotttrinh:
If an explicit isolation level and/or access mode is set in the transaction options, set the corresponding default transaction options that controls implicit transactions to this explicit value.

### DIFF
--- a/packages/gel/src/baseConn.ts
+++ b/packages/gel/src/baseConn.ts
@@ -482,6 +482,19 @@ export class BaseRawConnection {
     throw new errors.ProtocolError("invalid input codec");
   }
 
+  private isInTransaction(): boolean {
+    return (
+      this.serverXactStatus === TransactionStatus.TRANS_INTRANS ||
+      this.serverXactStatus === TransactionStatus.TRANS_ACTIVE
+    );
+  }
+
+  private _setStateCodec(state: Options): Uint8Array {
+    const buf = new WriteBuffer();
+    this.stateCodec.encode(buf, state._serialise(), NOOP_CODEC_CONTEXT);
+    return buf.unwrap();
+  }
+
   private _encodeParseParams(
     wb: WriteMessageBuffer,
     query: string,
@@ -491,6 +504,7 @@ export class BaseRawConnection {
     capabilitiesFlags: number,
     options: QueryOptions | undefined,
     language: Language,
+    isExecute: boolean,
   ) {
     if (versionGreaterThanOrEqual(this.protocolVersion, [3, 0])) {
       if (state.annotations.size >= 1 << 16) {
@@ -537,12 +551,30 @@ export class BaseRawConnection {
       if (this.stateCodec === INVALID_CODEC || this.stateCodec === NULL_CODEC) {
         wb.writeInt32(0);
       } else {
-        if (!this.stateCache.has(state)) {
-          const buf = new WriteBuffer();
-          this.stateCodec.encode(buf, state._serialise(), NOOP_CODEC_CONTEXT);
-          this.stateCache.set(state, buf.unwrap());
+        let encodedState: Uint8Array | null = null;
+
+        // Handle explicit transaction options isolation level, which should be
+        // set as the default isolation level for this query if it is not already
+        // in a transaction. If they match, do nothing.
+        if (
+          state.transactionOptions.isolation !==
+          state.config.get("default_transaction_isolation")
+        ) {
+          if (isExecute && !this.isInTransaction()) {
+            state = state.withConfig({
+              default_transaction_isolation: state.transactionOptions.isolation,
+            });
+          }
+          encodedState = this._setStateCodec(state);
+        } else {
+          encodedState = this.stateCache.get(state) ?? null;
+          if (encodedState === null) {
+            encodedState = this._setStateCodec(state);
+            this.stateCache.set(state, encodedState);
+          }
         }
-        wb.writeBuffer(this.stateCache.get(state)!);
+
+        wb.writeBuffer(encodedState);
       }
     }
   }
@@ -568,6 +600,7 @@ export class BaseRawConnection {
       capabilitiesFlags,
       options,
       language,
+      false,
     );
 
     wb.endMessage();
@@ -700,6 +733,7 @@ export class BaseRawConnection {
       capabilitiesFlags,
       options,
       language,
+      true,
     );
 
     wb.writeBuffer(inCodec.tidBuffer);
--- a/packages/gel/src/baseConn.ts
+++ b/packages/gel/src/baseConn.ts
@@ -482,7 +482,7 @@ export class BaseRawConnection {
     throw new errors.ProtocolError("invalid input codec");
   }
 
-  private isInTransaction(): boolean {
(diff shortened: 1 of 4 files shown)


packages/gel/src/baseConn.ts
```diff
@@ -537,12 +551,31 @@ export class BaseRawConnection {
...
-          this.stateCodec.encode(buf, state._serialise(), NOOP_CODEC_CONTEXT);
-          this.stateCache.set(state, buf.unwrap());
+        let encodedState: Uint8Array | null = null;
+
+        // Handle explicit transaction options isolation level, which should be
+        // set as the default isolation level for this query if it is not already
+        // in a transaction. If they match, do nothing.
+        if (
+          state.transactionOptions.isolation !==
+          state.config.get("default_transaction_isolation")
```
scotttrinh, line 561: I think this config value with typically be _unset_, so this path probably gets hit more often than it should considering that if you do nothing, the isolation will be `Serializable` which is also the default. But I didn't want to hard code the knowledge of the default into this driver code, so it seemed safer to just do an explicit comparison here and set the state more than needed.

packages/gel/src/options.ts
```diff
@@ -427,7 +432,9 @@ export class Options {
       this.config.size === 0 &&
       this.globals.size === 0 &&
       this.moduleAliases.size === 0 &&
-      this.module === "default"
+      this.module === "default" &&
+      this.transactionOptions === TransactionOptions.defaults() &&
+      this.retryOptions === RetryOptions.defaults()
```
scotttrinh, line 437: @1st1 I had to add the `transactionOptions` check here to get it to hit the state serialization path since you changed all of this code in your codec work. Do you mind just giving this method another close look to make sure we're not missing other places where state isn't being updated?

packages/gel/test/transaction.test.ts
```diff
@@ -42,16 +42,31 @@ async function run(test: (con: Client) => Promise<void>): Promise<void> {
...
         CREATE REQUIRED PROPERTY name -> std::str;
       };
+
+      CREATE TYPE ${typename}Conflict {
+          CREATE REQUIRED PROPERTY tmp -> std::str {
+              CREATE CONSTRAINT exclusive;
+          }
+      };
+
+      CREATE TYPE ${typename}ConflictChild extending ${typename}Conflict;
```
scotttrinh, line 57: Preemptively copying the test setup from the python work on `PreferRepeatableRead` knowing that we'll be adding this stuff later.

packages/gel/src/baseConn.ts
```diff
@@ -482,6 +482,19 @@ export class BaseRawConnection {
     throw new errors.ProtocolError("invalid input codec");
   }
 
+  private isInTransaction(): boolean {
```
1st1, line 485: nit: just for the consistency with the surrounding code I'd name this `_isInTransaction`

packages/gel/src/baseConn.ts
```diff
@@ -537,12 +551,31 @@ export class BaseRawConnection {
...
+
+        // Handle explicit transaction options isolation level, which should be
+        // set as the default isolation level for this query if it is not already
+        // in a transaction. If they match, do nothing.
+        if (
+          state.transactionOptions.isolation !==
+          state.config.get("default_transaction_isolation")
+        ) {
+          let newState = state;
+          if (isExecute && !this.isInTransaction()) {
```
1st1, line 564: Why is this `if (isExecute` important? Just curious, but this deserves a comment.

packages/gel/src/options.ts
1st1: Hm... I can try.

Separately speaking, why are we checking `retryOptions`? I don't think it affects server state in any way, it should be a pure client-side thing.

1st1 (approved):
All in all looks good

packages/gel/src/options.ts
scotttrinh: I mostly just wanted to make sure we were checking _all_ options to know if we should go into this faster path. I noticed that we don't have this fast path in Python, so I thought it would be safer to stick to a stricter sense of "default" here, but it's definitely true that this doesn't affect the state, so I'm happy to drop it.

packages/gel/src/baseConn.ts
scotttrinh: I matched the logic from the Python code that only modifies the state if called via execute. I don't have a complete picture of the full logic in my head though, so maybe this should only apply to the auto-select case?

scotttrinh:
Need to add the state descriptor check to avoid setting `default_transaction_isolation` on older versions of Gel server that do not support that config value. Will align with Python impl on this.

packages/gel/src/baseConn.ts
scotttrinh: [Latest update to make transaction options fully optional](https://github.com/geldata/gel-js/pull/1270/commits/f0e62738231ec020f3c15f0b9320e2a152a023fd) makes this a bit more explicit: the isolation will be `undefined` (which will match the config value) if you do not set either, which will avoid this extra state copy/serialization.

packages/gel/src/options.ts
```diff
@@ -427,7 +432,8 @@ export class Options {
       this.config.size === 0 &&
       this.globals.size === 0 &&
       this.moduleAliases.size === 0 &&
-      this.module === "default"
+      this.module === "default" &&
+      this.transactionOptions === TransactionOptions.defaults()
```
1st1, line 436: I don't like this comparison -- it's kinda weird because there are multiple ways how you can end up manually configuring `transactionOptions` to be in the "default" state and not match here with `===`.

Please add `TransactionOptions.isDefault()` method and call it here, checking against concrete defaults inside.

1st1: I think the `.defaults()` singleton must go, there is no use for it and allowing it to even exist in our code base was a mistake.

1st1 (approved):

packages/gel/src/options.ts
scotttrinh: The new "defaults" is now an empty object, but good call that it won't always be reference equal here. I'll fix that up.

1st1 (changes requested):
Oh wait, while we're here, we also need to enforce `default_transaction_access_mode` for implicit statements when transactions are configured to be readonly

packages/gel/src/baseConn.ts
```diff
@@ -537,12 +551,32 @@ export class BaseRawConnection {
...
-          this.stateCache.set(state, buf.unwrap());
+        let encodedState: Uint8Array | null = null;
+
+        // Handle explicit transaction options isolation level, which should be
+        // set as the default isolation level for this query if it is not already
+        // in a transaction. If they match, do nothing.
+        if (
+          versionGreaterThanOrEqual(this.protocolVersion, [3, 0]) &&
+          state.transactionOptions.isolation !==
+            state.config.get("default_transaction_isolation")
```
1st1, line 562: Need to also do the same dance for `default_transaction_access_mode`

1st1 (approved):

PR #1281, opened 2025-05-13
Use the correct runner for yarn and pnpm

scotttrinh:
`pnpm dlx` was trying to run a _package_ called `generate` rather than the locally installed binary. It turns out that yarn and pnpm can be called with no additional subcommand to run locally installed binaries.


scotttrinh:
Thanks Copilot, we should indeed add some tests to the `@gel/create` flow similar to our CLI, but pnpm refuses to `pnpm create` from a local tarball. It should still be possible, but we'll circle back to this soon.

PR #1291, opened 2025-06-03
Fix handling of inserts with both unless conflict and a with block

### LINKED ISSUE #679: Certain nested unlessConflict queries fail at runtime with: `Expr or its aliases used outside of declared 'WITH' block scope`
Minimal reproduction repo: https://github.com/freeatnet/repro-unless-conflict-issue/blob/main/src/pages/api/testcase.ts

There seems to be something causing a scope mismatch with the insert expression vs the insert unless conflict expression. I'm able to reproduce this in our main test suite as well.

### LINKED ISSUE #788: Error: `Expr or its aliases used outside of declared 'WITH' block scope` when `.unlessConflict` appended to `insert`
**Code**

Please clone https://github.com/varmiral/edgedb-with-issue.git
The code is based on https://www.edgedb.com/docs/clients/js/for#handling-conflicts-in-bulk-inserts with the addition of `.unlessConflict` (44:11) to handle existing characters.
Equivalent edgeql query works as expected.

Related issue: https://github.com/edgedb/edgedb-js/issues/679

**Error or desired behavior**

Error: Expr or its aliases used outside of declared 'WITH' block scope

- OS: `Ubuntu 23.10`
- EdgeDB version: `4.2`
- EdgeDB CLI version: `4.0.2+be1da73`
- `edgedb-js` version: `1.4.1`
- `@edgedb/generate` version: `0.4.1`
- Typescript version: `5.2.2`
- Node version: `20.9.0`

jaclarke:
Fixes #788, and should also fix #679

### DIFF
--- a/packages/generate/src/syntax/toEdgeQL.ts
+++ b/packages/generate/src/syntax/toEdgeQL.ts
@@ -606,7 +606,10 @@ function walkExprTree(
         );
       }
 
-      walkExprTree(expr.__expr__, parentScope, ctx);
+      childExprs.push(
+        ...walkExprTree(expr.__expr__, parentScope, ctx),
+        ...insertChildExprs,
+      );
       ctx.seen
         .get(expr.__expr__ as $expr_Insert)!
         .childExprs.push(...insertChildExprs);
--- a/integration-tests/lts/dbschema/default.esdl
+++ b/integration-tests/lts/dbschema/default.esdl
@@ -50,6 +50,12 @@ module default {
     multi link villains := .<nemesis[IS Villain];
   }
 
+  type Director {
+    required property name -> str {
+      constraint exclusive;
+    };
+  }
+
   scalar type year extending int16 {
     constraint min_value(1878);
   }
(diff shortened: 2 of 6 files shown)


packages/generate/src/syntax/toEdgeQL.ts
```diff
@@ -606,7 +606,10 @@ function walkExprTree(
         );
       }
 
-      walkExprTree(expr.__expr__, parentScope, ctx);
+      childExprs.push(
+        ...walkExprTree(expr.__expr__, parentScope, ctx),
+        ...insertChildExprs,
+      );
```
scotttrinh, line 612: I’m typing this out to ensure I understand the fix here:

In a previous with extraction fix, we extracted child expressions from the `unlessConflict` expression and stored them in the `insertChildExprs` array. However, we didn’t add these child expressions to the parent (`insert`) expression’s child expression list. Consequently, when the expression -> query machinery was validating the with access, it was not able to see that these expressions were allowed.

scotttrinh (approved):
🏆

PR #1292, opened 2025-06-03
feat(generate): allow `param` function to be composable

vancegillies:
## Overview

Issue: https://github.com/geldata/gel-js/issues/1290

This PR makes the return from the `param` function callable to allow composition while keeping the existing API stable.

Broken down into 3 parts
- Types
    - Add a `CallableWithParamsExpr` to `expr_WithParams` which adds the function interface on the type level
    - helper `paramsToInputArgs` to map the param's into an object of `$expr_Param` or a TS type so you can call the same function from within a `param` callback or with normal ts types
        - Note here: you can't use base exprs you have to use primitive TS types or "params" that are passed to the `param` callback, so you can not for example partially apply params, but I don't think there is a valid use case for that
     - Helper `paramOrTsType` for making a union for the base TS type of a param expr
 - The `param` function
     - pretty simple, just assign the original return  of the function to a `callableExpr` function that fulfills the above contract 
 - The `toEdgeQL` function
     - Adds a check in the `expr.__kind__ === ExpressionKind.WithParams` branch of `renderEdgeQL`function, which is checking if the expr has any args and building a subquery for them.
 

### Testing
I have done some little smoke tests, not everything works but I have a be able to get a basic composed query working here
```ts
import e from "./edgeql-js";

const testing1 = e.params(
  {
    arg1: e.uuid,
  },
  (params) => e.cast(e.str, params.arg1),
);

const testing2 = e.params(
  {
    arg2: e.uuid,
    arg3: e.optional(e.int16),
  },
  (params) => {
    const arg1 = testing1({
      arg1: params.arg2,
    });

    return e.set(e.select(arg1));
  },
);

const WITHOUT_ARGS = testing1.toEdgeQL();
const WITH_ARGS = testing1({
  arg1: "a216511e-78f4-4ccb-8492-3e5239b2d9db",
}).toEdgeQL();
const WITH_COMPOSE = testing2.toEdgeQL();
const COMPOSE_WITH_ARGS = testing2({
  arg2: "a216511e-78f4-4ccb-8492-3e5239b2d9db",
  arg3: 123,
}).toEdgeQL();

console.log(WITHOUT_ARGS);
console.log("");
console.log(WITH_ARGS);
console.log("");
console.log(WITH_COMPOSE);
console.log("");
console.log(COMPOSE_WITH_ARGS);
```
Which outputs
```bash
WITH
  __param__arg1 := <std::uuid>$arg1
SELECT (SELECT (<std::str>__param__arg1))

WITH
  __param__arg1 := <std::uuid>"a216511e-78f4-4ccb-8492-3e5239b2d9db"
SELECT (SELECT (<std::str>__param__arg1))

WITH
  __param__arg2 := <std::uuid>$arg2,
  __param__arg3 := <OPTIONAL std::int16>$arg3
SELECT (SELECT { (SELECT (WITH
  __param__arg1 := __param__arg2
SELECT (SELECT (<std::str>__param__arg1)))) })

WITH
  __param__arg2 := <std::uuid>"a216511e-78f4-4ccb-8492-3e5239b2d9db",
  __param__arg3 := <std::int16>123
SELECT (SELECT { (SELECT (WITH
  __param__arg1 := __param__arg2
SELECT (SELECT (<std::str>__param__arg1)))) })
```

Running these against a gel instance I get
Screenshot 2025-06-03 at 10 58 44 pm

#### TODO
- Testing with more complex types and queries
- Testing with client
- ...

### DIFF
--- a/packages/generate/src/syntax/params.ts
+++ b/packages/generate/src/syntax/params.ts
@@ -45,6 +45,34 @@ export type QueryableWithParamsExpression<
   runJSON(cxn: Executor, args: paramsToParamArgs<Params>): Promise<string>;
 };
 
+type paramOrTsType<Type extends ParamType, Optional extends boolean> =
+  | $expr_Param<string, Type, Optional>
+  | Readonly<BaseTypeToTsType<Type, true>>;
+
+type paramsToInputArgs<Params extends ParamsRecord> = {
+  [key in keyof Params as Params[key] extends ParamType
+    ? key
+    : never]: Params[key] extends ParamType
+    ? paramOrTsType<Params[key], false>
+    : never;
+} & {
+  [key in keyof Params as Params[key] extends $expr_OptionalParam
+    ? key
+    : never]?: Params[key] extends $expr_OptionalParam
+    ? paramOrTsType<Params[key]["__type__"], true>
+    : never;
+};
+
+type CallableWithParamsExpr<
+  Params extends ParamsRecord = Record<string, never>,
+  Expr extends TypeSet = TypeSet,
+> = {
+  (args: paramsToInputArgs<Params>): Expression<{
+    __element__: Expr["__element__"];
+    __cardinality__: Expr["__cardinality__"];
+  }>;
+};
+
 export type $expr_WithParams<
   Params extends ParamsRecord = Record<string, never>,
   Expr extends TypeSet = TypeSet,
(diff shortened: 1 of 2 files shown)


vancegillies:
@scotttrinh Would love some eyes on this to make sure I am heading in the right direction.

PR #1313, opened 2025-09-30
feat: allow users to customize the TypeScript type of a custom scalar

freeatnet:
## Why?

This PR adds the ability for users of the queries generator to override the TypeScript types of custom scalar variables via `OverrideCodecType`. The default scalar types are non-overrideable.

Example:

1. Set up a schema and a query:
`default.gel`:

```
module default {
  scalar type EvmAddress extending str {
    constraint expression on ( re_test(r'^0x[a-f0-9]{40}$', __subject__) );
  }

  scalar type HumanAge extending int16 {
    constraint max_value(120);
  }

  type User {
    required property primaryAddress -> EvmAddress {
      constraint exclusive;
    };
    required property firstName -> str;
    property username -> str {
      constraint exclusive;
    };
    property age -> HumanAge;
  }
}
```

`findUser.edgeql`:

```
select User {
  id,
  primaryAddress,
  username,
  age
} filter .id = <uuid>$id
```

2. Run `generate queries --target ts --use-resolved-codec-type`

3. Verify `findUser.query.ts` includes the overridable types:
```
export type FindUserReturns = {
  "primaryAddress": ResolvedCodecType<"EvmAddress", string>;
  "age": ResolvedCodecType<"HumanAge", number>;
  "id": string;
  "username": string | null;
} | null;

// Checking types -- should be types for base scalars
let result: NonNullable<FindUserReturns>;
result.username
    // ^? (property) "username": string | null

result.primaryAddress
    // ^? (property) "primaryAddress": string

result.age
    // ^? (property) "age": number
```

4. Add an override:
```typescript
declare module "gel" {
  export interface OverrideCodecType {
    "EvmAddress": `0x${string}`;
  }
}
```

5. Check types with overrides:

```
let result: NonNullable<FindUserReturns>;
result.username
    // ^? (property) "username": string | null

result.primaryAddress
    // ^? (property) "primaryAddress": `0x${string}`

result.age
    // ^? (property) "age": number
```

## How?

This PR attempts to implement this functionality in a backwards-compatible way, since `analyzeQuery` is technically a public API.[^1]

[^1]: Although it's only used in the queries generator within the repo, it's [mentioned in the README](https://github.com/geldata/gel-js/blob/6623d7ae97547fa11402e3a27c2d7afdf1212978/packages/generate/README.md?plain=1#L47-L63) as the primary API to build a custom generator, and it looks like [other projects](https://github.com/SeedCompany/cord-api-v3/blob/2aa9b360e0be5219f3563ced824c74825950b9aa/src/core/gel/generator/query-files.ts#L51) are using it.

### DIFF
--- a/packages/gel/src/codecTypeRegistry.ts
+++ b/packages/gel/src/codecTypeRegistry.ts
@@ -1,0 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface OverrideCodecType {}
+
+export type ResolvedCodecType<TDbTypeName, TDefaultTsType> =
+  TDbTypeName extends keyof OverrideCodecType
+    ? OverrideCodecType[TDbTypeName]
+    : TDefaultTsType;
--- a/packages/gel/src/index.shared.ts
+++ b/packages/gel/src/index.shared.ts
@@ -53,6 +53,8 @@ export * from "./errors";
 
 export type { Codecs } from "./codecs/codecs";
 
+export type { OverrideCodecType, ResolvedCodecType } from "./codecTypeRegistry";
+
 /* Private APIs */
 import type * as codecs from "./codecs/ifaces";
 import * as reg from "./codecs/registry";
--- a/packages/gel/src/index.shared.ts
+++ b/packages/gel/src/index.shared.ts
@@ -57,8 +57,10 @@ export type { OverrideCodecType, ResolvedCodecType } from "./codecTypeRegistry";
 
 /* Private APIs */
 import type * as codecs from "./codecs/ifaces";
+import { ScalarCodec } from "./codecs/ifaces";
 import * as reg from "./codecs/registry";
 import * as buf from "./primitives/buffer";
 export const _CodecsRegistry = reg.CodecsRegistry;
 export const _ReadBuffer = buf.ReadBuffer;
 export type _ICodec = codecs.ICodec;
+export const _ScalarCodec = ScalarCodec;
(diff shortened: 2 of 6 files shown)


scotttrinh (approved):
👏 Thanks for bearing with me on the back and forth!

freeatnet:
@scotttrinh no worries at all — I appreciate your responsiveness!

MartinCura:
Neat!

i'm trying this and because of TS magic i'm not familiar with claude had to add
```ts
import type {} from "gel"
```
to the `overrides.d.ts` so the `gel` type import in my `autogenerated.ts` didn't fail

scotttrinh:
Oh, thanks for the report! What is `autogenerated.ts`?

MartinCura:
> Oh, thanks for the report! What is `autogenerated.ts`?

Should've been explicit, i use
```bash
pnpm exec generate queries --target ts --use-resolved-codec-type --file ./queries/autogenerated && prettier --experimental-cli -w ./queries
```
`autogenerated.ts` is the one file where all my generated queries go, now starts with

```ts
// GENERATED by @gel/generate v0.6.4
// This file is automatically generated from .edgeql query files.
// To make changes, edit the source .edgeql file and regenerate.

import type {Executor, ResolvedCodecType} from "gel";

// ...
```

Also
```ts
// overrides.d.ts
// This import makes this file a module, enabling proper module augmentation
import type {} from "gel"

declare module "gel" {
  export interface OverrideCodecType {
    "default::ShopHandle": `@${string}`
  }
}
```

